PR #5, opened 2024-12-12
Support for GitHub issue/prs to markdown

gagb:
Enables the following

```bash
markitdown https://github.com/andrewyng/aisuite/issues/61
```

The output would be:

```markdown
# Request for Python Asyncio Support

I would like to request support for Python’s asyncio in this library. This feature would be particularly beneficial for Python services, which often rely on asynchronous programming for efficient and scalable operations.

Some providers, such as OpenAI, already offer native async support (e.g., `from openai import AsyncOpenAI`), making it straightforward to wrap these APIs. Others, like AWS, have community-supported async wrappers, such as `aioboto3`. For providers without async support, an interim solution using a synchronous wrapper could be implemented while awaiting a proper asyncio implementation.

Asyncio support would greatly enhance the usability of this library. Thank you for considering this enhancement.

**State:** open
**Created at:** 2024-11-26 02:16:05+00:00
**Updated at:** 2024-11-30 01:28:19+00:00
**Comments:**
- sarthakforwet (2024-11-26 03:47:34+00:00): Can you please assign this issue to me?
- soulcarus (2024-11-26 03:59:20+00:00): I refactored the code to use a thread pool instead of asyncio.

Initially, I attempted an asyncio-based solution. However, implementing a feature that solely uses asyncio would have required modifying several lines of code, which would have been time-consuming and inefficient for this specific task.

With just over 30 additional lines of code, I implemented a method that handles the heavy lifting by assigning each model inference to a separate thread. This change results in a performance improvement, reducing execution time by 
```

### DIFF
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -50,7 +50,7 @@ try:
 
     IS_GITHUB_ISSUE_CAPABLE = True
 except ModuleNotFoundError:
-    pass
+    IS_GITHUB_ISSUE_CAPABLE = False
 
 
 class _CustomMarkdownify(markdownify.MarkdownConverter):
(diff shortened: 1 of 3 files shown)


afourney:
Looks good, but I'll test it when I get back in to town today.

gagb:
I think my approach won't work with the CLI tho, I need to fix that.

gagb:
I fixed the cli issue as well. still not sure if `convert_stream` would work.

src/markitdown/_markitdown.py
```diff
@@ -904,6 +1033,28 @@ def convert(
...
+            x in parsed_url.path for x in ["/issues/", "/pull/"]
+        ):
+            github_token = kwargs.get("github_token", os.getenv("GITHUB_TOKEN"))
+            if not github_token:
+                raise ValueError(
+                    "GitHub token is required for GitHub issue or pull request conversion."
+                )
+            return GitHubIssueConverter().convert(
+                github_url=url, github_token=github_token
+            )
```
gagb, line 1051: @afourney this was the conditional that I added since accessing GH requires API call.

afourney:
PR has drifted too far from main. Merging is tough.

Also, I'm not 100% sold on calling an API here (though I suppose we do this for YouTube transcripts)

PR #10, opened 2024-12-13
Remove invalid classifiers

simonw:
requires-python says 3.10 and higher only


afourney:
Yup. Good catch. Thanks! Will merge once I sort out why CodeQL didn't run.

Issue #15, opened 2024-12-13
Add model client example in docs/readme


gagb:
> @gagb Would be great to have this as an example in the README! Thanks. 

 _Originally posted by @sdhutchins in [#12](https://github.com/microsoft/markitdown/issues/12#issuecomment-2542522406)_

```python
from markitdown import MarkItDown
from openai import OpenAI

client = OpenAI()
md = MarkItDown(mlm_client=client, mlm_model="gpt-4o")
image_file = 'https://gagb.github.io/imgs/bansal-chi21.png'
result = md.convert(source=image_file)
print(result.text_content)
```

ericsondea-collab:
xz  z cmz mx  ,z.x ,clmcsfnç

afourney:
The README now includes an OpenAI client example for image descriptions, using llm_client and llm_model. Closing as completed. Thanks!

Issue #12, opened 2024-12-13
LLM Integration


liamca:
As a feature suggestion, I would love if this could allow you to plug in an LLM (such as GPT-4o) where images that are included in the content could be sent to the LLM for further understanding. For example, if it was say a bar chart or graph, it would be sent to convert that to a Markdown table and be integrated in the final Markdown output.

dolbex:
I could see something coming back giving us markers of where that content _should_ appear along with the source content in a way that we could handle it as we please. Ex: Send to LLM, upload to a server and link, etc.

janzheng:
maybe as an MCP server? That wouldn't be too hard

gagb:
Relevant: https://github.com/microsoft/markitdown/blob/main/src/markitdown/_markitdown.py#L755 
Ive tested an older version of it, but not the new version.

sdhutchins:
@gagb Would be great to have this as an example in the README! Thanks.

liamca:
> [@gagb](https://github.com/gagb) Would be great to have this as an example in the README! Thanks.

Agreed. IMO, a PDF based example would be best where it is able to do the text in the normal way, but any images in the PDF are sent to the LLM.  I think this is a more compelling example to organizations then just PNG's.

alejlatorre:
> > As a feature suggestion, I would love if this could allow you to plug in an LLM (such as GPT-4o) where images that are included in the content could be sent to the LLM for further understanding. For example, if it was say a bar chart or graph, it would be sent to convert that to a Markdown table and be integrated in the final Markdown output.
> 
> It's already supported, here you go. You need an openai key.
> 
> from markitdown import MarkItDown
> from openai import OpenAI
> 
> client = OpenAI()
> md = MarkItDown(mlm_client=client, mlm_model="gpt-4o")
> image_file = 'https://gagb.github.io/imgs/bansal-chi21.png'
> result = md.convert(source=image_file)
> print(result.text_content)
> Output
> 
> ImageSize: 388x368
> 
> # Description:
> This image illustrates a comparison of decision accuracy with and without explanations. On the left, there's a computer icon connected to a user symbol, suggesting human-computer interaction. The vertical axis represents the "Accuracy of decisions," ranging from 0.0 to 1.0.
> 
[...]

Issue #9, opened 2024-12-13
PDF and JPG do not convert.


josangmi:
PDF and JPG  in the future?

gagb:
Pdfs are supported! How would you convert a JPG to markdown?

adrianliechti:
Maybe an OCR pre-process or a llm model doing something similar

Inspiration could be found at similar projects:
https://github.com/DS4SD/docling
https://github.com/Unstructured-IO/unstructured

both seem to accept images

gagb:
Good point. This is related to #12.

PR #16, opened 2024-12-13
Add example of using MarkItDown with OpenAI to README

gagb:
Add an example of using MarkItDown with OpenAI integration to the `README.md`.

* Add a new section "Example with OpenAI Integration".
* Include code to import `MarkItDown` and `OpenAI`.
* Show initializing the `OpenAI` client.
* Demonstrate converting an image file using `MarkItDown` with the `OpenAI` client.

Addresses #15 #12 #9

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,19 @@ result = markitdown.convert("test.xlsx")
 print(result.text_content)
 ```
 
+## Example with OpenAI Integration
+
+```python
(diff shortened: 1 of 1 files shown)


kylehowells:
I didn't realise the model could do this.

gagb:
Fixed by the other PR.

PR #19, opened 2024-12-14
Fix character decoding issues with text-like files

### LINKED ISSUE #18: use charset_normalizer
Currently, all the text-like files are loaded as utf-8. But this causes issues, for example, with csv files created on Windows in Japanese. Those might be encoded as CP932. https://github.com/jawah/charset_normalizer can help with this issue. Currently it throws this:

```sh
markitdown._markitdown.FileConversionException: Could not convert 'test.csv' to Markdown. File type was recognized as ['.csv']. While converting the file, the following error was encountered:

  File "/Users/brc-dd/foo/.venv/lib/python3.12/site-packages/markitdown/_markitdown.py", line 1041, in _convert
    res = converter.convert(local_path, **_kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/brc-dd/foo/.venv/lib/python3.12/site-packages/markitdown/_markitdown.py", line 166, in convert
    text_content = fh.read()
                   ^^^^^^^^^
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x91 in position 1: invalid start byte
```

`chardet` doesn't work with such files. It detects them as Shift JIS while Windows uses an extended version of Shift JIS (CP932).

brc-dd:
fixes #18

Non utf-8 html pages are still not supported. It will need detecting language from content-type header or meta tags like `<meta charset="utf-8" />` or `<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />`. And at last fall back to auto detecting.

Bing, WikiPedia and Youtube converters don't need this as they always have utf-8 encoding.


brc-dd:

gagb:
@brc-dd can you please expand the tests? Or at least share an example csv with this error?

brc-dd:
Updated the tests.

gagb:
Can you please run the pre-commit checks? I think it should be `pre-commit run --all-files`.

brc-dd:
Formatted.

gagb (approved):

tests/test_markitdown.py
```diff
@@ -87,6 +87,13 @@
     "data:image/svg+xml,%3Csvg%20width%3D",
 ]
 
+CSV_CP932_TEST_STRINGS = [
+    "名前,年齢,住所",
+    "佐藤太郎,30,東京",
+    "三木英子,25,大阪",
+    "髙橋淳,35,名古屋",
+]
+
```
brc-dd, line 96: This started appearing weird for some reason on GitHub diff:

image

But if you open the file it shows it correctly:

image

image

PR #24, opened 2024-12-14
Add installation instructions

haesleinhuepf:
Hi all,

that's a fantastic library! I'm just proposing to add installation instructions.

Best,
Robert


gagb (approved):
Would be useful to add instructions for installing from source as well (concisely).

gagb (approved):

PR #26, opened 2024-12-15
Update README.md

pawarbi:
I have updated the readme with three changes:
- Created sections for Installation and Usage to help users on the readme page
- Added installation instruction on how to install using pip
- Added additional example of using LLM. This will be the primary use case and will help users.


pawarbi:

gagb (approved):

Issue #13, opened 2024-12-13
asynchronous support?


fuad00:
Any plans on this?

gagb:
There is no current plan yet, but PRs welcome! @afourney

0xRaduan:
Would be great, I can try to take a stab on this tomorrow.

0xRaduan:
Just an update from my end: I checked the current implementation, and while there are some benefits of async support, many libraries that are used under the hood, are not supporting async:
- markdownify
- mammoth
- pandas
- pdfminer
- python-pptx

At this point, if all underlying calls will be sync, I am not sure, how much benefit it would give to provide async APIs on `markitdown`, outside of a thin wrapper. Otherwise, we need a full rewrite + using a full new set of libraries.

If people want to use MarkItDown in async manner, I think they should just create a thin wrapper for markitdown.

I did a tiny PR on how I would do that: https://github.com/microsoft/markitdown/pull/32;

clemlesne:
Would be awesome yes!

clemlesne:
Performance is regrettable when triggering parallel analysis or including in a web app. Adding async behavior would greatly improve this. Required implementation time should be low.

mohammedabdulrehman:
Perfect! waiting for async support

sglebs:
Converting an image to text using an LLM makes a network call to an external LLM provider. This I/O would greatly benefit from having an async implementation. With the current implementation, if you use FastAPI and use Markitdown, you block your entire Python VM.

fuad00:
i can see some MCP server stuff implemented, but basic examples in README still showing us sync code...

sglebs:
> Converting an image to text using an LLM makes a network call to an external LLM provider. This I/O would greatly benefit from having an async implementation. With the current implementation, if you use FastAPI and use Markitdown, you block your entire Python VM.

This code in _image_converter.py can block the Python VM for about 7 seconds (the time it takes to call the LLM remotely):

```
# Call the OpenAI API
        response = client.chat.completions.create(model=model, messages=messages)
        return response.choices[0].message.content

```

It's 7 seconds that your FastAPI app will be frozen :-)

PR #32, opened 2024-12-15
Add AsyncMarkItDown as a wrapper

0xRaduan:
My take on how this code should support async, given all underlying libraries are not supporting async.

For more details/context, see: https://github.com/microsoft/markitdown/issues/13#issuecomment-2543834157


pyproject.toml
```diff
@@ -40,6 +40,12 @@ dependencies = [
   "pathvalidate",
 ]
 
+[tool.uv]
+dev-dependencies = [
+    "pytest>=7.0",
+    "pytest-asyncio>=0.23.0",
+]
+
```
0xRaduan, line 48: out of curiosity, why normal dependencies actually don't specify a version they should be higher/lower?

tests/test_async_markitdown.py
0xRaduan, line 1: This is just port of normal `test_markitdown.py` to async

0xRaduan:

gagb:
@0xRaduan -- thanks for the PR.

@afourney @jackgerrits @0xRaduan,
I love this PR, but I am worried about the code duplication because of asynchronous tests.
What if we stuck with sync implementation, but added an FAQ showing how a user could create an asynchronous wrapper similar to your approach.

This was users know how to create an asynchronous wrapper, but we avoid the code/test duplication? Please lmk your thoughts.

0xRaduan:
Hey @gagb - good feedback, I agree.

In general, this async code is just a wrapper over sync code[1], so technically you don't need to have a full suite of async tests, as all of the functionality should be tested in normal tests.

I would suggest keeping just one async test to verify that the wrapper works, and remove everything else. Initially I've migrated all tests to just verify that it works identically for my own's sake.

[1]: https://github.com/microsoft/markitdown/pull/32/files#diff-7405f619a8ecc0b22b53b22a3cfcfd1fc92c62aea6e08de97674c8bbdd9bb949R42-R44

pyproject.toml
gagb: I think you could but in this case we don't have strong reason to?

gagb:
> Hey @gagb - good feedback, I agree.
> 
> In general, this async code is just a wrapper over sync code [[1](https://github.com/microsoft/markitdown/pull/32/files#diff-7405f619a8ecc0b22b53b22a3cfcfd1fc92c62aea6e08de97674c8bbdd9bb949R42-R44)], so technically you don't need to have a full suite of async tests, as all of the functionality should be tested in normal tests.
> 
> I would suggest keeping just one async test to verify that the wrapper works, and remove everything else. Initially I've migrated all tests to just verify that it works identically for my own's sake.

I like your framing. Would love @jackgerrits feedback on the strategy.

gagb:
@0xRaduan can you fix the precommit and test errors

0xRaduan:
@gagb - fyi, updated the code, fixed linter, removed all tests and just kept one async test to make sure wrapper works in async environment.

gagb (commented):
Please resolve feedback by Jack!

markthepixel:
Can we keep this moving (sorry to sound like a PM hahahaha)

0xRaduan:
@markthepixel - let me take a look tomorrow, sorry for the wait!

markthepixel:
I gatta be honest I thought this was my issue but it turns out when you feed a string in instead of a path/link it just gives a vague error 😆

Issue #40, opened 2024-12-15
'res' variable not declared inside _convert before usage


dsimanis:
`res` variable inside `MarkItDown` class `_convert` method is not declared before usage:

```
_markitdown.py", line 1045, in _convert
    if res is not None:
       ^^^
UnboundLocalError: cannot access local variable 'res' where it is not associated with a value
```

gagb:
Can you please share more context for repro?

sambartik:
> Can you please share more context for repro?

I got this error when I tried to convert a non-existing pdf file:

```python
from markitdown import MarkItDown

markitdown = MarkItDown()
result = markitdown.convert("file-that-doesnt-exist.pdf")
print(result.text_content)
```

tarrasque73:
I have the same issue just by executing the sample code on github (using a PDF) and the file definitely exists

```
from markitdown import MarkItDown

markitdown = MarkItDown()
result = markitdown.convert("test.pdf")
print(result.text_content)
```

sidesteps:
Getting same error for pdf file that exists in the current dir.

sambartik:
> I have the same issue just by executing the sample code on github (using a PDF) and the file definitely exists
> 
> ```
> from markitdown import MarkItDown
> 
> markitdown = MarkItDown()
> result = markitdown.convert("test.pdf")
> print(result.text_content)
> ```

I can't reproduce your results. This exact code works for me with repo's example test.pdf in CWD.

Environment:
```
Python 3.13.1
macOS 15.1.1 x86_64
markitdown             0.0.1a2
```

sidesteps:
With exact same code and test.pdf in cwd I get:

```
Traceback (most recent call last):
  File "/home/sstep/tomd/tomd.py", line 5, in <module>
    result = md.convert("test.pdf")
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sstep/tomd/.venv/lib/python3.12/site-packages/markitdown/_markitdown.py", line 902, in convert
    return self.convert_local(source, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sstep/tomd/.venv/lib/python3.12/site-packages/markitdown/_markitdown.py", line 922, in convert_local
    return self._convert(path, extensions, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #42, opened 2024-12-15
Proper handling when trying to read a non-existing file

axegon:
In relation to https://github.com/microsoft/markitdown/issues/40 , instead of UnboundLocalError, indicate that you are trying to access a non-existing file.

### DIFF
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -845,6 +845,10 @@ class UnsupportedFormatException(BaseException):
     pass
 
 
+class FileDoesNotExistException(BaseException):
+    pass
+
+
 class MarkItDown:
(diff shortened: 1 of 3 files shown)


axegon:

src/markitdown/_markitdown.py
```diff
@@ -911,6 +915,9 @@ def convert_local(
         ext = kwargs.get("file_extension")
         extensions = [ext] if ext is not None else []
 
+        if not os.path.exists(path):
+            raise FileDoesNotExistException(f"File {path} does not exist")
```
gagb, line 919: Can you place raise the Python built in exception instead?

https://docs.python.org/3/library/exceptions.html#FileNotFoundError

Issue #43, opened 2024-12-16
Basic CLI help text


Munchotaur:
As it stands, markitdown installs and works just fine via pipx.

However running it as a command line app with `-h` or `--help` throws an `UnsupportedFormatException`. Not that much help text is needed but it would be better than an error.

kevinclb:
I made a pull request to address this issue as my first contribution to this project.

https://github.com/microsoft/markitdown/pull/46

Let me know if you have any suggestions or comments.

@gagb

ArunSunhare:
Issue Title: Improve Basic CLI Help Text for markitdown
Description:
The markitdown tool installs and runs successfully when installed via pipx. However, when executing the command with --help or without any arguments, the output is unclear or does not provide sufficient guidance to new users.

Steps to Reproduce:
Install markitdown via pipx:
bash
Copy code
pipx install markitdown
Run the command without arguments:
bash
Copy code
markitdown
Alternatively, run the help command:
bash
Copy code
markitdown --help
Observed Behavior:
The output is either unclear, incomplete, or missing critical information about the tool's usage.
Expected Behavior:
The help text should clearly describe:
What the tool does.
How to use it (e.g., syntax and examples).
List of available commands and options.
Proposed Solution:
Update the CLI help text to provide:
A brief description of the tool.
A usage guide, including the command structure.
Examples of common use cases.
Example Help Text:

bash
Copy code
markitdown: A CLI tool for converting markdown files to HTML.

Usage:
  markitdown [OPTIONS] <input_file.md> [output_file.html]

Options:
  --help       Show this help message and exit.
  --version    Display the current version of markitdown.

Examples:
  markitdown input.md output.html
  markitdown input.md

Additional Notes:
Contributors may need to explore the CLI parsing library used (e.g., argparse, click, etc.) to make the necessary improvements.
Ensure the updated help text aligns with any additional features or options provided by markitdown.

PR #46, opened 2024-12-16
feature: add argument parsing and setup.py file for cli tool capability

kevinclb:
To address a feature request in [this issue](https://github.com/microsoft/markitdown/issues/43), I added argument parsing via the Python argparse library, along with a setup.py file for CLI setup. This tool now works on the CLI, printing md files into the terminal.

image


kevinclb:

setup.py
```diff
@@ -0,0 +1,31 @@
...
+        'console_scripts': [
+            'markitdown=markitdown.__main__:main',
+        ],
+    },
+    author='Adam Fourney',
+    author_email='adamfo@microsoft.com',
+    description='Convert various file formats to markdown',
+    license='MIT',
+    python_requires='>=3.6',
+)
```
l-lumin, line 31: Since markitdown are using Hatch, the setup.py file is unnecessary. The CLI settings can be found here.

https://github.com/microsoft/markitdown/blob/81e3f24acd0049a59cd2dcb2d01d0a98cc57c734/pyproject.toml#L51-L52

kevinclb: Got it. That might be something useful to add to the documentation. 

Do you have any opinions on using argparse to parse the input flags?

setup.py
gagb, line 1: Is this necessary given the `pyproject.toml`. @afourney and I agree the argparse in `__main__.py` is helpful.

kevinclb: Thanks for the input. I removed `setup.py`.

gagb:
Can you please run pre-commit? `pre-commit run --all-files`

PR #48, opened 2024-12-16
Fix: pass the kwargs to _convert method when converting an url file

Soulter:
I found my customized `mlm_prompt` cannot be applied properly, and I found out the reason is the `kwargs` hasn't been passed to `_convert` method. And I fixed it.

Before:

image

After:

image


Soulter:

PR #49, opened 2024-12-16
Small type hint fixing + added makefile with couple commands.

Abdujabbar:
Hello!

I've found some places where type hinting was missed, so I decided to fix these places and also added Makefile to simplify the command run.

### DIFF
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -882,7 +882,7 @@ class WavConverter(MediaConverter):
     """
 
     def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
-        # Bail if not a XLSX
+        # Bail if not a WAV
         extension = kwargs.get("file_extension", "")
         if extension.lower() != ".wav":
             return None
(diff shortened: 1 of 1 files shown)


Abdujabbar:

gagb:
@jackgerrits do you think it makes sense to mix make files and pyproject.toml ?

gagb:
Related #46

Makefile
gagb, line 1: Is there a way of doing without including a Makefile? E.g., by modifying the pyproject.toml?

jackgerrits:
> @jackgerrits do you think it makes sense to mix make files and pyproject.toml ?

Looks like this repo has no existing scripts or a system for defining scripts to run. A make file is totally reasonable and lightweight :) Can always be replaced if something like poe is used in future

jackgerrits (approved):
LGTM!

gagb:
@afourney are you okay with including a make file?

afourney:
> @afourney are you okay with including a make file?

Sure. If @jackgerrits says it's fine, I'll take that as gospel :)

gagb:
Sorry @Abdujabbar, we appreciate your contributions though.

PR #60, opened 2024-12-16
Added Dockerfile 

madduci:
Hello

Thanks for this library/tool, it is pretty interesting. I've added a `Dockerfile`, so the tool can be used without having the Python build environment locally.

Best regards


madduci:

Issue #25, opened 2024-12-14
Add binary and homebrew support for CLI


gagb:
small fr for the cli: 

- static binary (no idea if this is possible in python, but its nearly 2025)
- homebrew support

_Originally posted by @loeffel-io in https://github.com/microsoft/markitdown/issues/11#issuecomment-2543195986_

movahhedi:
+1
even without a binary, at least an intuitive way to use it in the CLI

daeho-ro:
In order to add `markitdown` to homebrew, one of the dependency `onnxruntime` should have sdist but it doesn't.
- https://github.com/microsoft/onnxruntime/issues/13650

PR #61, opened 2024-12-16
Add binary and Homebrew support for CLI #25

PratibhaK11:
## This PR adds support for building a static binary and installing via Homebrew for the MarkItDown CLI.

# Changes:
- Build: Added a pyinstaller --onefile build process to generate a static binary of the MarkItDown CLI.
- Install: Created a Homebrew formula (markitdown.rb) to allow easy installation of MarkItDown via Homebrew.

# Releases - https://github.com/PratibhaK11/markitdown/releases
# Test:
 - Build - pyinstaller --onefile /Users/pratibhakumari/Desktop/markitdown/src/markitdown/__main__.py
 - Install - brew install ./markitdown.rb


markitdown.rb
```diff
@@ -0,0 +1,16 @@
+class Markitdown < Formula
+  desc "Markitdown: A tool for converting markdown documents"
+  homepage "https://github.com/PratibhaK11/markitdown"
```
gagb, line 3: I have;t built home-brew installs before. But we can't have hardcoded paths in the repo.

jackgerrits: Agreed, especially if it is a different repo.

@PratibhaK11 if you'd like a homebrew package, I would suggest submitting a PR with the formula to https://github.com/Homebrew/homebrew-core

stepbystepcode:
You can easily install it using `uv tool install "markitdown[all]"` instead of Homebrew.

PR #73, opened 2024-12-16
Fix LLM terminology in code

CharlesCNorton:
Replaced all occurrences of mlm_client and mlm_model with llm_client and llm_model for consistent terminology when referencing Large Language Models (LLMs).


CharlesCNorton:
@microsoft-github-policy-service agree [company="{AI}"]

afourney:
I agree with this, but changing this now will break things for people already experimenting with the library. Can we go in 2 steps. First let's enable the llm_ keywords as an alias to mlm_, and mark mlm_ as deprecated. Then we can remove mlm in a future release (e.g., 1 or two releases from now)

gagb:
@CharlesCNorton can you please resolve the conflicts. Also double check any other mentions of mlm e.g., in the README.

gagb (approved):

PR #97, opened 2024-12-17
feat: Add RSSConverter 

Soulter:
Many websites provide RSS / Atom feeds, and most of them have very high information density and value, such as blogs([Microsoft's Blog](https://opensource.microsoft.com/blog/feed/), etc), news and other websites, so I want to add support for RSS and Atom formats for this project.

- Implemented RSSConverter class for converting `.rss`, `.atom`, `.xml`(with format of rss/atom) into Markdown format.
- Added a test for testing the converter.
- `xml` is a python standard library, the changes did not import third-party libraries


afourney:
This is very nice! And someone was asking about this in another context (Discord), so I'll be happy to share this with them. Let me do a little testing, and I think we're in good shape to include it.

PR #98, opened 2024-12-17
fix incorrect comments for "bail if not ..." for WAV and image cases.

waterimp:
correct the code comments that were misleading.


afourney (approved):
Oooops. Thanks!

gagb (approved):

PR #100, opened 2024-12-17
Added llm tests to the local test set.

afourney:


gagb (approved):

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
   "SpeechRecognition",
   "pathvalidate",
   "charset-normalizer",
+  "openai",
```
casperdcl, line 42: Unsure about this - perhaps should be an optional dep?

PR #112, opened 2024-12-18
chore: configure Dependabot for GitHub Actions updates

l-lumin:
Set up Dependabot to check and update GitHub Actions. Some current GitHub Actions are deprecated.
I noticed other Microsoft projects also use Dependabot, so I think it’s a good for us to use it too.

https://github.com/microsoft/PowerToys/blob/main/.github/dependabot.yml


gagb:
Merci @l-lumin

PR #115, opened 2024-12-18
Small type hint fixing + added makefile with couple commands. (Need to recreate)

Abdujabbar:
Hey, this is a duplicate MR. There were merge conflicts, so I decided to resolve them locally. The elder MR was closed due to the sync process on Github(Sorry for that): https://github.com/microsoft/markitdown/pull/49. 

That's why I opened another MR, here are the same changes, but without conflicts. 

```Hello!

I've found some places where type hinting was missed, so I decided to fix these places and also added Makefile to simplify the command run.```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+install:
+	pip install hatch
+
+test:
+	hatch shell & hatch test
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,2 @@
-install:
-	pip install hatch
-
 test:
 	hatch shell & hatch test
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Abdujabbar** [comment]

File: Makefile [hunk 1]
```diff
@@ -0,0 +1,5 @@
+install:
+	pip install hatch
```
vb-long, line 2: installing Hatch with pip might not be the best idea

> This method modifies the Python environment in which you choose to install. Consider instead using [pipx](https://hatch.pypa.io/latest/install/#pipx) to avoid dependency conflicts.
https://hatch.pypa.io/latest/install/#pip

Abdujabbar:
Closed, as team members rejected it.

PR #129, opened 2024-12-18
Safeguard against path traversal for ZipConverter

finchy:
fix: prevent path traversal vulnerabilities in ZipConverter

Added a secure check for path traversal vulnerabilities in the ZipConverter class. Now validates extracted file paths using `os.path.commonprefix` to ensure all files remain within the intended extraction directory. Raises a `ValueError` if a path traversal attempt is detected.

- Normalized file paths using `os.path.normpath`.
- Added specific exception handling for `zipfile.BadZipFile` and traversal errors.
- Ensured cleanup of extracted files after processing when `cleanup_extracted` is enabled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,3 +160,5 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+src/.DS_Store
+.DS_Store
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -1144,9 +1144,14 @@ class ZipConverter(DocumentConverter):
                 # Safeguard against path traversal
                 for member in zipObj.namelist():
                     member_path = os.path.normpath(os.path.join(extraction_dir, member))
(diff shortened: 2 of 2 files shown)


finchy:
@microsoft-github-policy-service agree [company="{ThreatSTOP}"]

finchy:

finchy:
@microsoft-github-policy-service agree company="ThreatSTOP"

gagb:
@Josh-XT FYI

Josh-XT:
> @Josh-XT FYI, can you please take a look?

Looks like a good improvement to me!

afourney:
Zip this issue is what keeps me up at night re: zip files. (The problem is worse with tar files, which we do not support). 

PR, looks reasonable to me, but I want to get another few opinions. We might even consider not excluding zip support by default -- making it optional.

Josh-XT (approved):

afourney:
Thinking on this more, I believe it is possible to open a byte stream for a file in the Zip, without requiring that we decompress the whole archive. We could in theory enumerate supported files, and one by one write them to a tmp file on disk (with a safe name and permissions), and convert the, and clean them up.

In any case, I'll merge this PR now (as it is strictly better), and we can revisit in a future PR.

afourney (commented):
Approved and merged. See discussion.

PR #136, opened 2024-12-19
feat: enable Git support in devcontainer

numekudi:
Added a process to install Git only when the devcontainer starts.
This should be helpful when the editor has source control features.
Avoided changes to the image and the addition of a separate Dockerfile.


numekudi:

gagb (approved):

gagb:
Thanks! We will likely move to a separate Dockerfiles for dev and running. In the meantime this looks good to us.

Issue #131, opened 2024-12-18
Create advanced PDF convertor


gagb:
> > [@gagb](https://github.com/gagb) Would be great to have this as an example in the README! Thanks.
> 
> Agreed. IMO, a PDF based example would be best where it is able to do the text in the normal way, but any images in the PDF are sent to the LLM. I think this is a more compelling example to organizations then just PNG's.
> 
> Exactly this. We could use something like PyMuPDF4LLM to extract images and classify them with VLM later, but if this could be done in one step during pdf->md conversion, that would be brilliant! 

 _Originally posted by @vaclcer in [#12](https://github.com/microsoft/markitdown/issues/12#issuecomment-2551568601)_

Either extend the existing one or create a new one. This is a frequent use case.

LockedThread:
> > > [@gagb](https://github.com/gagb) Would be great to have this as an example in the README! Thanks.
> > 
> > Agreed. IMO, a PDF based example would be best where it is able to do the text in the normal way, but any images in the PDF are sent to the LLM. I think this is a more compelling example to organizations then just PNG's.
> > 
> > Exactly this. We could use something like PyMuPDF4LLM to extract images and classify them with VLM later, but if this could be done in one step during pdf->md conversion, that would be brilliant! 
> 
>  _Originally posted by @vaclcer in [#12](https://github.com/microsoft/markitdown/issues/12#issuecomment-2551568601)_
> 
> Either extend the existing one or create a new one. This is a frequent use case.

I have an open source project that I implemented PDF vector drawing reconstruction for the purpose of ML-based analysis. I would be willing to contribute it. 

It uses a DFS to combine nearby vector drawing bounding boxes to create the maximally connected bbox. This bounding box can then be used to create a pymupdf pixmap and extract what is essentially a screenshot of the vector art.

Tables in PDFs are usually a collection of distinct vector drawing lines and rectangles, not actual images.

See: https://github.com/FEPrep/parser/blob/main/parser/drawing/drawing.py

HawkClaws:
I think PyMuPDF4LLM is a good option, but I believe there might be licensing issues.
Currently, markitdown is under the MIT license, so I think it’s better to avoid integrating libraries with more restrictive licenses than MIT.

Similarly, libraries that require high machine specifications might also face limitations, as there could be situations where they cannot be used effectively.
For features like image and audio extraction, I think it would be better to use APIs provided by LLMs such as OpenAI's.

For example, as follows:
[...]

PR #139, opened 2024-12-19
update: change pdf text parser to pymupdf4llm

tungsten106:
Using `pymupdf4llm` instead of `pdfminer` to parse pdf contents into markdown formats, as suggested by #131.

Pros and Cons:
- `pdfminer` extract texts only, generated files have no heading, titles, etc. `pymupdf4llm`, however, could perform a nice markdown featrues including different levels of heading, code blocks and images (could be saved to specific path, but not included in this commit)
- However,  `pymupdf4llm` may easily create lines of digits which belongs to plots, and create non-existing tables. This is a common problem to most PDF parsers, except those using ocr models (such as [markers](https://github.com/VikParuchuri/marker), [MinerU](https://github.com/opendatalab/MinerU)).

### DIFF
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -23,6 +23,7 @@ import markdownify
 import pandas as pd
 import pdfminer
 import pdfminer.high_level
+import pymupdf4llm
 import pptx
 
 # File-format detection
@@ -684,10 +685,12 @@ class PdfConverter(DocumentConverter):
         if extension.lower() != ".pdf":
             return None
 
-        return DocumentConverterResult(
-            title=None,
-            text_content=pdfminer.high_level.extract_text(local_path),
-        )
+        # return DocumentConverterResult(
+        #     title=None,
+        #     text_content=pdfminer.high_level.extract_text(local_path),
+        # )
+        text_content = pymupdf4llm.to_markdown(local_path, show_progress=False)
+        return DocumentConverterResult(title=None, text_content=text_content)
 
 
 class DocxConverter(HtmlConverter):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
   "pandas",
   "openpyxl",
   "pdfminer.six",
+  "pymupdf4llm",
   "puremagic",
   "pydub",
   "youtube-transcript-api",


src/markitdown/_markitdown.py
```diff
@@ -23,6 +23,7 @@
 import pandas as pd
 import pdfminer
```
hoangquochung1110, line 24: Let's remove unused imports

src/markitdown/_markitdown.py
```diff
@@ -684,10 +685,12 @@ def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
...
             return None
 
-        return DocumentConverterResult(
-            title=None,
-            text_content=pdfminer.high_level.extract_text(local_path),
-        )
+        # return DocumentConverterResult(
+        #     title=None,
+        #     text_content=pdfminer.high_level.extract_text(local_path),
+        # )
```
hoangquochung1110, line 691: Why don't we just remove them ?

alphaleadership:
maybe this can patch #142

l-lumin (commented):
I think it's better to let the user choose the engine rather than replacing it

afourney:
> I think it's better to let the user choose the engine rather than replacing it

I agree. There are pros and cons to each. The main thing is to allow a common interface.

Can you propose an interface for this? One option is to just call register_page_converter externally, and the precedence logic would give precedence to whichever converter is registered later.

Another option would be to see which dependencies are installed (though this is more opaque)

tungsten106:
> > I think it's better to let the user choose the engine rather than replacing it
> 
> I agree. There are pros and cons to each. The main thing is to allow a common interface.
> 
> Can you propose an interface for this? One option is to just call register_page_converter externally, and the precedence logic would give precedence to whichever converter is registered later. (see here for example: https://github.com/microsoft/markitdown/blob/925c4499f72757abcf6cb521ee10e4844967af3d/src/markitdown/_markitdown.py#L1269C1-L1287C1)
> 
> Another option would be to see which dependencies are installed (though this is more opaque)

I have added a parameter `pdf_engine` to let the user choose engine. For example, 

```python
source = "https://arxiv.org/pdf/2308.08155v2.pdf"
markitdown.convert(source, pdf_engine="pymupdf4llm")  # use pymupdf4llm
markitdown.convert(source, pdf_engine="pdfminer")  # use pdfminer
```

src/markitdown/_markitdown.py
```diff
@@ -676,19 +677,27 @@ def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
...
-            title=None,
-            text_content=pdfminer.high_level.extract_text(local_path),
-        )
+        pdf_engine = kwargs.get("pdf_engine", "pdfminer")
+        if pdf_engine == "pdfminer":
+            text_content = pdfminer.high_level.extract_text(local_path)
+        elif pdf_engine == "pymupdf4llm":
+            text_content = pymupdf4llm.to_markdown(local_path, show_progress=False)
+        else:
+            return None     # unknown method
```
l-lumin, line 699: This part doesn't return anything. Could you update it? Maybe add a warning message?

src/markitdown/_markitdown.py
```diff
@@ -676,19 +677,27 @@ def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
 
 class PdfConverter(DocumentConverter):
     """
-    Converts PDFs to Markdown. Most style information is ignored, so the results are essentially plain-text.
+    Converts PDFs to Markdown. Most style information is ignored, so the results are essentially plain-text.    
     """
 
     def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
+        """
```
l-lumin, line 684: I think `pdf_engine` should be a named parameter in the function instead of being accessed via `kwargs`. This provides more clarity and ensures better handling of default values

l-lumin (changes requested):
could you add test for `pymupdf4llm`?

src/markitdown/_markitdown.py
tungsten106: Thanks for you suggestion. I have added it to a named parameters ( I was meant to align with other converters' parameter definition), and added the exception case when `pdf_engine` is not valid. The new test cases could be seen on my new commit.

tests/test_markitdown.py
```diff
@@ -7,7 +7,8 @@
 import requests
 
 from warnings import catch_warnings, resetwarnings
-
+import sys
+sys.path.insert(0, "/home/yxl/Projects/markitdown/src")
```
l-lumin, line 11: I think don't need this. Let me know if you need help setting up the test!

src/markitdown/_markitdown.py
```diff
@@ -676,19 +677,27 @@ def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
...
-        return DocumentConverterResult(
-            title=None,
-            text_content=pdfminer.high_level.extract_text(local_path),
-        )
+        if pdf_engine == "pdfminer":
+            text_content = pdfminer.high_level.extract_text(local_path)
+        elif pdf_engine == "pymupdf4llm":
+            text_content = pymupdf4llm.to_markdown(local_path, show_progress=False)
+        else:
+            # return None     # unknown method
```
l-lumin, line 698: remove this

src/markitdown/_markitdown.py
l-lumin: This is optional, but adding one more named parameter could make this more customizable, like:

```py
def convert(self, local_path, engine: Literal['pdfminer', 'pymupdf4llm']='pdfminer', engine_kwargs=None, **kwargs) -> Union[None, DocumentConverterResult]:
```

src/markitdown/_markitdown.py
```diff
@@ -676,19 +677,27 @@ def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
...
+            raise FileConversionException("'pdf_engine' not valid. Please choose between ['pdfminer', 'pymupdf4llm'].")
```
l-lumin, line 699: I'd suggest checking the engine first. You could use `_engines` to define the allowed engines

```py
_engines: Mapping[str, Any] = {
        "pdfminer": pdfminer,
        "pymupdf4llm": pymupdf4llm,
    }
    
###
if engine is not None and engine not in self._engines:
    raise
```

tests/test_markitdown.py
tungsten106: I will remove this, it won't be used

src/markitdown/_markitdown.py
tungsten106: It is a good idea for the ease of adding further engines. I have updated those changes in the latest commit. An example of using `engine_kwargs` for pdf imeges extraction could also be found in the test file.

l-lumin:
@tungsten106 
Thank you for your contribution! It looks great so far. 
Just one more thing—when running the tests, files is generated. Could you add the following to the `.gitignore` file inside `tests/` and modify test to export inside `out/` folder? 
Also, do you think there’s any way to make the tests run faster?

```
out/
```

l-lumin:
https://github.com/microsoft/markitdown/pull/139#issuecomment-2553113646
answer this if you have time

tungsten106:
> @tungsten106 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

tungsten106:
> @tungsten106 Thank you for your contribution! It looks great so far. Just one more thing—when running the tests, files is generated. Could you add the following to the `.gitignore` file inside `tests/` and modify test to export inside `out/` folder? Also, do you think there’s any way to make the tests run faster?
> 
> ```
> out/
> ```

I have updated that. 
For test speed, have you tried to use `pytest-xdist` to run `test_markitdown.py` in parallel?

```bash
pip install pytest-xdist

# let it decide
pytest -n auto tests/test_markitdown.py
# or using specific cpu numbers, like 8
pytest -n 8 tests/test_markitdown.py
```

l-lumin:
> I have updated that.
For test speed, have you tried to use pytest-xdist to run test_markitdown.py in parallel?

Thank you. You can run tests in parallel without using `pytest-xdist`; simply run `hatch test -p`. I want to discuss ways to improve the speed of `test_markitdown_pdf`


```sh
(hatch-test.py3.13) root@e2c718eb6604:/workspaces/markitdown# hatch test -p
========================================================================================================== test session starts ==========================================================================================================
platform linux -- Python 3.13.1, pytest-8.3.4, pluggy-1.5.0
rootdir: /workspaces/markitdown
configfile: pyproject.toml
plugins: rerunfailures-14.0, mock-3.14.0, anyio-4.7.0, xdist-3.6.1
8 workers [5 items]     
ss...               
```

tungsten106:
> > I have updated that.
> > For test speed, have you tried to use pytest-xdist to run test_markitdown.py in parallel?
> 
> Thank you. You can run tests in parallel without using `pytest-xdist`; simply run `hatch test -p`. I want to discuss ways to improve the speed of `test_markitdown_pdf`
> 
> ```shell
> (hatch-test.py3.13) root@e2c718eb6604:/workspaces/markitdown# hatch test -p
> ========================================================================================================== test session starts ==========================================================================================================
> platform linux -- Python 3.13.1, pytest-8.3.4, pluggy-1.5.0
> rootdir: /workspaces/markitdown
> configfile: pyproject.toml
> plugins: rerunfailures-14.0, mock-3.14.0, anyio-4.7.0, xdist-3.6.1
> 8 workers [5 items]     
> ss...               
> ```

The speed of `pymupdf4llm.to_markdown` might be slow due to package inner processes.
We could use a smaller test pdf since the original article have 43 pages. Adding `pages=[i for i in range(10)]` parameters for `pymupdf4llm` or `page_numbers=[i for i in range(10)]` for `pdfminer` could be one solution.

l-lumin:
Work great. I just test `pages=range(10)` is enough, same with `page_numbers`. for run one test you can run below command so don't need comment 

```py
    # test_markitdown_remote()
    # test_markitdown_local()
    # test_markitdown_exiftool()
    # test_markitdown_deprecation()
    # test_markitdown_llm()
```

```sh
hatch test tests/test_markitdown.py::test_markitdown_pdf
```

alphaleadership:
maybe add clli option

tungsten106:
> maybe add clli option

I have added an cli option `--engine` to choose different converters' engine and you could test it with the following command:

```bash
python -m markitdown tests/test_files/2308.08155v2.pdf --engine "pymupdf4llm" -o document.md
```
For other `engine_kwargs` it might not be suitable for cli since it involves unknown data types.

@afourney Really appreciate it if you could review this when you have time? Also thanks for @l-lumin and @alphaleadership for giving very valuable advice.

markthepixel:
Yes, this is perfect. Let's ship 🚢

afourney:
It has been brought to my attention that `pymupdf4llm`, which makes it incompatible with our MIT license. So I'll have to close this PR unfortunately. 

If you would like to use pymupdf4llm for PDFs, I recommend implementing it as a 3rd party plugin. See: https://github.com/microsoft/markitdown/tree/main/packages/markitdown-sample-plugin

Issue #145, opened 2024-12-19
Raising ModuleNotFoundError if openai package not present


Viddesh1:
Hello,

Twice skip_llm variable is initialized with boolean True.

https://github.com/microsoft/markitdown/blob/925c4499f72757abcf6cb521ee10e4844967af3d/tests/test_markitdown.py#L18-L23

Raising the appropriate error if the openai python package is not present.

```
# Don't run the llm tests without a key and the client library
skip_llm = False if os.environ.get("OPENAI_API_KEY") else True
try:
    import openai
except ModuleNotFoundError:
    raise ModuleNotFoundError
```
OR

Directly importing it. If the package is not present then import error will displayed.

```
import openai
```

Regards!
Viddesh

afourney:
OpenAI is optional. The tests (and library should be able to run without it). What is the specific issue you are facing with the current setup?

Viddesh1:
Hello,

Okay understood, my bad will test code more. 

If the openai key is not present then skip_llm becomes boolean True and if the openai import does not work then skip_llm becomes True again. That was my concern that the skip_llm is getting initialized 2 times with True. 

We are checking if the key is present or not and also for the openai can be imported or not based on which skip_llm can be true or false.

Closing this issue.

Regards!
Viddesh

PR #146, opened 2024-12-19
ModuleNotFoundError raised if openai package not present.

Viddesh1:
#145


afourney:
https://github.com/microsoft/markitdown/issues/145#issuecomment-2555373097

PR #157, opened 2024-12-19
Feat/save document converter result

absadiki:
Hello, 
Thanks for the awesome package :)

This is a simple PR that adds a new `save` method to the `DocumentConverterResult` class, making it easier to save the converted document result to a file.

## Changes:

* Added a save method to the `DocumentConverterResult` class, which takes a `file_path` and optional encoding parameter. The method writes the `text_content` to the specified file using the specified encoding.
* Updated the sample script in the README file to showcase the new `save` method.

### DIFF
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -157,6 +157,7 @@ class DocumentConverterResult:
         with open(file_path, "w", encoding=encoding) as f:
             f.write(self.text_content)
 
+
 class DocumentConverter:
     """Abstract superclass of all DocumentConverters."""
 
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -146,6 +146,16 @@ class DocumentConverterResult:
         self.title: Union[str, None] = title
         self.text_content: str = text_content
 
+    def save(self, file_path: str, encoding: str = "utf-8") -> None:
+        """
+        Save the converted document result `text_content` to a file.
+
+        params:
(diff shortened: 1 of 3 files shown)


absadiki:

src/markitdown/_markitdown.py
```diff
@@ -146,6 +146,17 @@ def __init__(self, title: Union[str, None] = None, text_content: str = ""):
...
+        """
+        Save the converted document result `text_content` to a file.
+
+        params:
+            file_path: The path to save the document result to.
+            encoding: The encoding to use when writing the document.
+        """
+        with open(file_path, "w", encoding=encoding) as f:
+            f.write(self.text_content)
+
```
l-lumin, line 159: it serves the same purpose as #116. could you refactor the code to use a single function for better reusability? 
- #116

absadiki: sure @l-lumin,

I think it makes more sense to have the function to save the result into a file inside the `DocumentConverterResult`.

* I refactored the CLI to use this function.
* I also added an `encoding` parameter to the arguments.

l-lumin: Thank you. I think so too. Look good

afourney:

`DocumentConverterResult` still has no `save()` helper, so the idea itself is still open. The class has moved to `packages/markitdown/src/markitdown/_base_converter.py` in the monorepo restructure, though, and this branch targets the old top-level `src/markitdown/` layout.

PR #165, opened 2024-12-20
Added default issue template.

Viddesh1:


Viddesh1:

l-lumin:
I think you should spend some time to modifying the issue templates to suit this repository rather than using the default

l-lumin:
I think for the issue templates, we should start with two types:
- for feature requests, such as adding support for a new file format (and maybe including a sample file)
- for bug reports, which should include details like the problem description, steps to reproduce, additional information like the version, and the file

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Smartphone (please complete the following information):**
```
gagb, line 31: remove smartphone. combine with desktop

Viddesh1: Thanks for comment. I removed the smartphone information and removed extra space in between.

gagb:
Is there a reason you used md instead of YAML? https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

We should keep the issue template really concise.

Viddesh1:
Hello @gagb ,

The reason is simplicity. 
Markdown provides simplicity to us to update issue template and issue creator using md template to update the issue or feature request. Also, if they want they can update the template specific to creating the issues add or subtract specific information.

YAML form based issue templates is also good.  Since it is form based issue creator can add the specific information that they want in specific form based fields. It would also make sure that all the fields are filled with some information and no fields are empty. If we want concise information then we can go with YAML.

Which one do you prefer?

Thanks!
Viddesh

gagb (approved):

afourney:

`.github/` still has no `ISSUE_TEMPLATE/`, so this gap is real and worth filling — I've noted it separately so it doesn't disappear along with the PR.

PR #168, opened 2024-12-20
Small refactor for MarkItDown.convert_response function.

Abdujabbar:
I am trying to refactor this block of code:

https://github.com/microsoft/markitdown/blob/main/src/markitdown/_markitdown.py#L1396-L1419

As we can see here, we are using try/except, which requires manually closing/removing the file. 

I want to suggest using context managers, and use NamedTemporaryFile.

I think it's better to read, and support in the future the code.


Abdujabbar:

afourney:
Thanks for the PR.

The new 0.1.X branch has moved away from temporary files, so this PR has unfortunately become stale -- but I very much appreciate your effort here.

PR #173, opened 2024-12-20
docs: update README with Docker instructions and details

l-lumin:
Update the Docker instructions in the README, as the current ones are not working


afourney:

The Dockerfile has been reworked since (monorepo install paths, an `INSTALL_GIT` build arg, a non-root default user), and the README's Docker section was rewritten alongside it, so the instructions this was correcting are no longer the ones in the file.

PR #181, opened 2024-12-21
feat(docker): multi-stage builds for separate development and production environments

l-lumin:
Separate Dockerfiles for dev and prod.
The setup has been tested locally in both stages.


ahmadmy2312-max:
https://github.com/microsoft/markitdown/pull/181#issue-2753725111

sergiomoralesc33-cyber (approved):

sergiomoralesc33-cyber (approved):

afourney:

The Dockerfile is still single-stage, so the idea stands. The file has been rewritten since, though (monorepo install paths, `INSTALL_GIT`, non-root user), and `.devcontainer/` changed with it, so this branch no longer applies.

PR #189, opened 2024-12-21
Chore: precise typing for stream and json parameters

konverner:
Propositions for resolving TODOs on typing:

1. `IO[Any]` for `stream` parameter accommodates both binary and text streams.
2. `Union[dict, list, Any]` for `json` parameter accommodates the possible types of JSON-like data structures.

### DIFF
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -14,7 +14,7 @@ import tempfile
 import traceback
 import zipfile
 from xml.dom import minidom
-from typing import Any, Dict, List, Optional, Union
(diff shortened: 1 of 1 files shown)


afourney:
Thanks for the contribution. 

The main branch has been refactored, and streams are now properly type-hinted as BinaryIO. Json is still a problem, but I'm not sure how Union[Any, ...] is better than just Any.

In any case, I'll close this PR for now since the refactoring created any conflicts. If you want to re-open it and re-target it to the v0.0.X branch, I could probably merge this with fewer issues.

Issue #154, opened 2024-12-19
Convert video to markdown


gagb:

aviral-bhardwaj:
started working on it

KomalMaurya:
@aviral-bhardwaj is this issue still open? 
Can I contribute in it?

PR #193, opened 2024-12-21
feat(converter): add video converter.

absadiki:
This PR is a step towards resolving #154.

It introduces a `VideoConverter` class that converts videos to markdown by:

* Extracting metadata (if exiftool is installed)
* Performing speech transcription (if speech_recognition and pydub are installed)
* Generating a summary via a multimodal LLM from the transcription [This is optional and defaults to True if `llm_client` is configured]

### Notes:

* I believe checking the file type based on the extension is not ideal. There are many video extensions, and I think checking the mime_type would be a better approach, as it can cover a wider range of video files.
* I’m unsure about the testing strategy .. should we focus only on testing exiftool? Please share your thoughts on this.
* Additionally, I suggest refactoring `Mp3Converter` into a more general `AudioConverter`, as there are many audio extensions to consider. If you agree with this, I can submit a separate PR for it.


l-lumin:
could you add tests?

l-lumin:
could you add tests?

absadiki:
@l-lumin, could you provide a sample video file that is allowed to be uploaded to the repo?

l-lumin:
> @l-lumin, could you provide a sample video file that is allowed to be uploaded to the repo?

I think you can use the file you tested locally.If it's wrong, can change it later

absadiki:
@l-lumin, okey I created a sample video file using ffmpeg. I've added test for exiftool for now.
Maybe we can add tests for transcription as well, but #194 should be merged first.

hariappointy:
Will this be merged soon?

afourney:

Video is partly covered now: `AudioConverter` accepts `.mp4` and `video/mp4` for metadata extraction and speech transcription, and `--cu-file-types` can route video to Azure Content Understanding. A dedicated `VideoConverter` with LLM summarization is still a gap, but this branch targets the old single-file `_markitdown.py`.

Issue #74, opened 2024-12-17
### Audio Transcript: Error. Could not transcribe this audio.


caesarllj:
### Audio Transcript:
Error. Could not transcribe this audio.

Soulter:
Hi, can you provide more information about this?

caesarllj:
from markitdown import MarkItDown
markitdown = MarkItDown()  
    
result = markitdown.convert("******.mp3")
print(result.text_content)

.xlsx .pdf no problem,.mp3 format file error prompt

### Audio Transcript:
Error. Could not transcribe this audio.

tathagatankit:
I am also getting issues with audio files. I tired with .mp3 and .wav file format and getting the below issue:
![Image](https://github.com/user-attachments/assets/8419e184-fbba-4177-982f-807176ad6abf)

Additionally, I am using a Mac and noticed that exiftool was not installed during the pip installation. I’m unable to install this tool externally because macOS is blocking the installation and displaying a malware warning. Could someone assist with this? Your help would be greatly appreciated. Thank you!

caesarllj:
> I am also getting issues with audio files. I tried with .mp3 and .wav file format and getting the below issue: ![Image](https://github.com/user-attachments/assets/8419e184-fbba-4177-982f-807176ad6abf)
> 
> Additionally, I am using a Mac and noticed that exiftool was not installed during the pip installation. I’m unable to install this tool externally because macOS is blocking the installation and displaying a malware warning. Could someone assist with this? Your help would be greatly appreciated. Thank you!

I am in window env,the reason for above problems is that I am unable to use google service.

claudiocassimiro:
Have the same issue here, running in linux ubuntu

My error log:

➜ python3 test.py
Traceback (most recent call last):
  File "/home/claudio/Desktop/test.py", line 4, in <module>
    result = md.convert("SonhoGelado.mp3")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/claudio/miniconda3/lib/python3.12/site-packages/markitdown/_markitdown.py", line 260, in convert
    return self.convert_local(source, stream_info=stream_info, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/claudio/miniconda3/lib/python3.12/site-packages/markitdown/_markitdown.py", line 314, in convert_local
[...]

PR #194, opened 2024-12-21
fix(transcription): TRANSCRIPTION_CAPABLE should be iniztialized

absadiki:
Running `Markitdonw` on audio files without having `pydub` and `speech_recognition` installed would result in a NameError: name 'IS_AUDIO_TRANSCRIPTION_CAPABLE' is not defined.

This PR resolves that issue.

#74


afourney (approved):

afourney:
Looks good. Thanks.

Issue #133, opened 2024-12-19
Expose via Web API


lqdev:
Set up a Web API so users can use the library via a REST endpoint. 

This is also useful for Docker scenarios as well.

lqdev:
Feel free to assign this one to me.

gautam-e:
I started working on something like this but the moment I import `from markitdown import MarkItDown` I get the ugly warning:
```
python3.12/site-packages/starlette/routing.py:297: ResourceWarning: Unclosed file <tempfile.SpooledTemporaryFile object at 0x149021990>
  await self.app(scope, receive, send)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```
This is without even actually instantiating a MarkItDown object let alone using it. Is that a bug in the package?

gautam-e:
Here's a web app I made that works with Markitdown. I use it for my personal workflow: www.docx2md.com

luisquintanilla:
Great to see the community work and samples around this! 

@elbruno also put together a quick sample of how such an API might be consumed in a C# client application. 

https://github.com/elbruno/MarkItDownServer

Once #202 is merged, it'd be great to publish the Dockerfile as an image to the Microsoft Container Registry so you don't have to pull down the entire repo to use the container / server.

dezoito:
Any updates on this?

afourney:
Thanks for the suggestion. This repository is intended to provide Python libraries that can be incorporated into other systems, and additional web servers or REST APIs are now explicitly out of scope. A separate package or project depending on markitdown would be the appropriate home for this. I’m closing this as not planned.

PR #202, opened 2024-12-22
[Draft] Add Web API for MarkItDown

vs4vijay:
Related to #133

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     ffmpeg \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip install markitdown
+RUN pip install markitdown fastapi uvicorn
 
 # Default USERID and GROUPID
 ARG USERID=10000
@@ -20,4 +20,4 @@ ARG GROUPID=10000
 
 USER $USERID:$GROUPID
 
-ENTRYPOINT [ "markitdown" ]
+ENTRYPOINT ["uvicorn", "src.markitdown.api:app", "--host", "0.0.0.0", "--port", "8000"]
--- a/src/markitdown/api.py
+++ b/src/markitdown/api.py
@@ -19,9 +19,9 @@ async def convert(file: UploadFile = File(...)):
         markitdown = MarkItDown()
         result = markitdown.convert(temp_file_path)
 
-        output_file_path = f"/tmp/{os.path.splitext(file.filename)[0]}.md"
-        with open(output_file_path, "w") as output_file:
-            output_file.write(result.text_content)
+        # output_file_path = f"/tmp/{os.path.splitext(file.filename)[0]}.md"
+        # with open(output_file_path, "w") as output_file:
+        #     output_file.write(result.text_content)
 
         os.remove(temp_file_path)
 
--- a/src/markitdown/api.py
+++ b/src/markitdown/api.py
@@ -25,7 +25,8 @@ async def convert(file: UploadFile = File(...)):
 
         os.remove(temp_file_path)
 
-        return FileResponse(output_file_path, filename=f"{os.path.splitext(file.filename)[0]}.md")
+        # return FileResponse(output_file_path, filename=f"{os.path.splitext(file.filename)[0]}.md")
+        return {"markdown": result.text_content}
 
     except Exception as e:
         raise HTTPException(status_code=500, detail=str(e))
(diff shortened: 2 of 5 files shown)


vs4vijay:
In Testing

xingxingcan:
Thank you very much, we need api interface access in order to utilize automated processes, we are very optimistic about this project, thanks again to the great developers.

homanp:
I deployed a simple REST API, ya'll can try it out. Here are the specs: 

https://x.com/pelaseyed/status/1872326539208986903

Dockerfile
```diff
@@ -12,12 +12,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     ffmpeg \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip install markitdown
+RUN pip install markitdown fastapi uvicorn
```
GerardSmit, line 15:
```suggestion
RUN pip install markitdown fastapi uvicorn python-multipart
```

Without `python-multipart` the API won't start.

It's still weird that `markitdown` is being installed here. This means if the source gets changed and you want to check if the Docker image works you first need to publish the package to pip... It would be better if the current `src` folder would be included here.

Dockerfile
```diff
@@ -12,12 +12,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 
 # Default USERID and GROUPID
 ARG USERID=10000
 ARG GROUPID=10000
 
 USER $USERID:$GROUPID
 
-ENTRYPOINT [ "markitdown" ]
+ENTRYPOINT ["uvicorn", "src.markitdown.api:app", "--host", "0.0.0.0", "--port", "8000"]
```
GerardSmit, line 23: This doesn't work: `src/markitdown/api.py` doesn't exists, so the API will not start.

You must copy the `src` directory:

```suggestion
COPY src /src

ENTRYPOINT ["uvicorn", "src.markitdown.api:app", "--host", "0.0.0.0", "--port", "8000"]
```

Note that in `.dockerignore` that everything is ignored, so the `COPY` won't work.  
Modify the `.dockerignore` so the  `src/`-folder gets included:

```ignore
*
!/src
```

GerardSmit (commented):
I tried to run the API locally through Docker. I had to change the following things to make it work.

Dockerfile
markthepixel: When you install with pip install "fastapi[standard]" it comes with some default optional standard dependencies.

If you don't want to have those optional dependencies, you can instead install pip install fastapi.

ranma42:
I tried out this branch and found a couple of issues; locally I fixed them and rebased the branch on top of the current `main`.
You can find my current state here https://github.com/ranma42/markitdown/tree/add-web-api

src/markitdown/api.py
```diff
@@ -0,0 +1,32 @@
...
+        raise HTTPException(status_code=400, detail="No file uploaded")
+
+    try:
+        contents = await file.read()
+        temp_file_path = f"/tmp/{file.filename}"
+        with open(temp_file_path, "wb") as temp_file:
+            temp_file.write(contents)
+
+        markitdown = MarkItDown()
+        result = markitdown.convert(temp_file_path)
```
ranma42, line 20: when this is a long CPU-bound operation (for example when converting a long PDF) this can block the executor

src/markitdown/api.py
```diff
@@ -0,0 +1,32 @@
...
+app = FastAPI()
+
+@app.post("/convert")
+async def convert(file: UploadFile = File(...)):
+    if not file.filename:
+        raise HTTPException(status_code=400, detail="No file uploaded")
+
+    try:
+        contents = await file.read()
+        temp_file_path = f"/tmp/{file.filename}"
```
ranma42, line 15: if multiple requests are accepted concurrently with the same filename, they can conflict
(nb: AFAICT this currently does not occur because requests are handled sequentially)

Ahoo-Wang:
https://github.com/Ahoo-Wang/markitdown/tree/dev/packages/markitdown-api

https://github.com/Ahoo-Wang/markitdown/blob/dev/packages/markitdown-api/delpoy/markitdown-api.yml

Docker Image:

  - ahoowang/markitdown-api:0.1.2-0.0.12
  - ghcr.io/ahoo-wang/markitdown-api:0.1.2-0.0.12
  - registry.cn-shanghai.aliyuncs.com/ahoo/markitdown-api:0.1.2-0.0.12

image

afourney:

A serving layer has since landed in a different shape: `packages/markitdown-mcp` exposes MarkItDown over MCP (stdio, SSE, and streamable HTTP). Since this was still marked draft and predates the monorepo split, I'm closing it out.

PR #205, opened 2024-12-23
Enhance _markitdown.py to support embedding images in markdown

MauroDruwel:
This update enhances the `DocxConverter` class by adding functionality to extract and embed images when converting DOCX files to Markdown. The changes include:

- **`convert_image` method:** Handles image extraction, sanitizes the filename, and saves the image as a PNG in the specified output directory.
- **`convert` method:** Integrates the `convert_image` method with the Mammoth library's HTML conversion process, ensuring images are extracted and included in the final output.


MauroDruwel:

afourney:
Looks promising. Sanitizing filenames is a task that will come up a lot (and my already be implemented). I'm going to ask around for advice on how to handle this broadly and robustly (without regular expressions). Filenames may also have other restrictions (e.g., length etc.) on some OSs.

MauroDruwel:
Hi @afourney,

I've added the following improvements:

1. **Test cases**: I've included some test cases, which you can view in the [result here](https://maurodruwel.be/MarkItDownExample/).
2. **Filename formatting**: I've replaced spaces with underscores in filenames to ensure compatibility.
3. **Length limit checker**: A checker has been added to enforce a length limit for filenames.
4. **File existence check**: Now the script checks if the file already exists and appends a counter to avoid overwriting.
5. **Alt text formatting**: I've fixed an issue where newlines in the image `alt_text` were causing images not to show by removing them.

Let me know if you need any further adjustments!

MauroDruwel:
Hi @afourney, is there anything left that I need to do?

afourney:

Embedded-image handling has since landed as the `markitdown-ocr` plugin, which extracts text from images inside PDF, DOCX, PPTX and XLSX using LLM vision. It takes a different approach — extracting text rather than writing image files to an output directory — but covers much of what this was reaching for.

PR #220, opened 2024-12-26
feat(docker): improve dockerfile build

syaghoubi00:
Build from source instead of installing package.
Don't ignore files needed for building in dockerignore.
Set apt to be non-interactive.
Use generic nobody:nogroup instead of uid:gid of 10000.
Don't install recommended packages with git.


syaghoubi00:

PR #223, opened 2024-12-27
Fix markitdown.convert_stream handling of leading blanks

### LINKED ISSUE #222: [bug] markitdown._markitdown.UnsupportedFormatException
Here is the steps to reproduce the issue:

OS: Windows

Shell: PowerShell v7.4.6

```ps1
pip install markitdown
curl -s https://www.duotify.com -o a.htm
type a.htm | markitdown
```

The error message:

```log
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Python312\Scripts\markitdown.exe\__main__.py", line 7, in <module>
  File "C:\Python312\Lib\site-packages\markitdown\__main__.py", line 38, in main
    result = markitdown.convert_stream(sys.stdin.buffer)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\markitdown\_markitdown.py", line 1142, in convert_stream
    result = self._convert(temp_path, extensions, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\markitdown\_markitdown.py", line 1260, in _convert
    raise UnsupportedFormatException(
markitdown._markitdown.UnsupportedFormatException: Could not convert 'C:\Users\user\AppData\Local\Temp\tmpzn_zxenq' to Markdown. The formats [] are not supported.
```

doggy8088:
Fixes #222

Address issue with `markitdown.convert_stream` crashing on input with leading blank characters or line breaks.

* Modify `convert_stream` function in `src/markitdown/_markitdown.py` to strip leading blank characters or line breaks from the input stream using a new helper function `_strip_leading_blanks`.
* Add a test case in `tests/test_markitdown.py` to verify that `markitdown.convert_stream` handles input with leading blank characters or line breaks correctly.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/microsoft/markitdown/pull/223?shareId=aac1dfd3-227c-4348-a18b-4355dd2fe5f3).

### DIFF
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -1344,7 +1344,7 @@ class MarkItDown:
         result = None
         try:
             # Write to the temporary file
-            content = stream.read()
+            content = self._strip_leading_blanks(stream.read())
             if isinstance(content, str):
                 fh.write(content.encode("utf-8"))
             else:
@@ -1366,6 +1366,10 @@ class MarkItDown:
             os.unlink(temp_path)
 
         return result
+
+    def _strip_leading_blanks(self, content: bytes) -> bytes:
+        """Helper function to strip leading blank characters or line breaks from content."""
+        return content.lstrip()
 
     def convert_url(
         self, url: str, **kwargs: Any
--- a/tests/test_markitdown.py
+++ b/tests/test_markitdown.py
@@ -300,6 +300,15 @@ def test_markitdown_llm() -> None:
         assert test_string in result.text_content.lower()
 
 
+def test_markitdown_strip_leading_blanks() -> None:
+    markitdown = MarkItDown()
+
+    # Test input with leading blank characters
+    input_data = b"   \n\n\n<html><body><h1>Test</h1></body></html>"
+    result = markitdown.convert_stream(io.BytesIO(input_data), file_extension=".html")
+    assert "<h1>Test</h1>" in result.text_content
+
+
 if __name__ == "__main__":
     """Runs this file's tests from the command line."""
     test_markitdown_remote()
@@ -307,3 +316,4 @@ if __name__ == "__main__":
     test_markitdown_exiftool()
     test_markitdown_deprecation()
     test_markitdown_llm()
+    test_markitdown_strip_leading_blanks()


src/markitdown/_markitdown.py
```diff
@@ -1344,7 +1344,7 @@ def convert_stream(
         result = None
         try:
             # Write to the temporary file
-            content = stream.read()
+            content = self._strip_leading_blanks(stream.read())
```
afourney, line 1347: The stream might not be text -- in which case stripping characters could be very problematic. Suggest we move this to the inside of the if statement below.

src/markitdown/_markitdown.py
```diff
@@ -1367,6 +1367,10 @@ def convert_stream(
 
         return result
 
+    def _strip_leading_blanks(self, content: bytes) -> bytes:
```
afourney, line 1370: Do we need a helper function for lstrip?

afourney:
Thanks for the PR. Before we accept this, I would like to better understand why leading spaces are causing a crash. I suspect that the issue lies deeper in the logic for guessing the file format, and it will be triggered in other conditions as well.

tests/test_markitdown.py
```diff
@@ -300,10 +300,20 @@ def test_markitdown_llm() -> None:
         assert test_string in result.text_content.lower()
 
 
+def test_markitdown_strip_leading_blanks() -> None:
+    markitdown = MarkItDown()
+
+    # Test input with leading blank characters
+    input_data = b"   \n\n\n<html><body><h1>Test</h1></body></html>"
+    result = markitdown.convert_stream(io.BytesIO(input_data), file_extension=".html")
+    assert "<h1>Test</h1>" in result.text_content
```
afourney, line 309: This test will fail. The output will be Markdown, not HTML

afourney:
I've determine the problem, and am investigating other fixes that don't involve truncating the file. See https://github.com/microsoft/markitdown/issues/222#issuecomment-2569874433 for more.

afourney:
Fixed in #260 without modifying file.

PR #219, opened 2024-12-26
Add json

Gad:
the proposed PR addresses somehow issue #34.
Having not found a suitable python library, I added a JsonConverter class independent of the PlainTextConverter.
in a nutshell : 
- parse document with python json.load() 
- converts nested dictionaries into objects starting with the filename (e.g. json_org_example2.menu.id for {menu :{id : ...}} in json_org_example2.json
- converts JSON arrays into md ordered lists since json arrays are ordered 
- adds some MD formatting 
- adds back tips around key/value to avoid markdown interpreters using them as keywords   

hash test / pre-commit ran, Ok

tested against json.org examples. output available here : https://gist.github.com/Gad/4de412dabb73c6a20b0a098089226814

PR #251, opened 2025-01-03
JsonConverter for Converting JSON Files into Structured Markdown Files

ZeyuTeng96:
This PR implements a JsonConverter class that converts JSON files into Markdown files while preserving the dictionary structure of the data to the greatest extent possible during the conversion process.

- For JSON files with dict like:

image

The conversion result will be:
image

- For JSON files containing a list of dicts, such as the SFT fine-tuning data for LLMs:
image

The conversion result will display each element of the list in sequence. Additionally, users can customize the value of the list heading (default is Elem):
image

This conversion approach preserves the dictionary structure to the greatest extent possible, allowing converted .md files to be more intuitively readable and visually clear for users. Furthermore, when performing retrieval-augmented generation (RAG), LLMs are highly familiar with dict structure and can process them effectively.

### DIFF
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -1204,6 +1204,42 @@ class ZipConverter(DocumentConverter):
                 text_content=f"[ERROR] Failed to process zip file {local_path}: {str(e)}",
             )
 
+class JsonConverter(DocumentConverter):
+    """
+    Converts Jsons to Markdown. The output preserves the structure of the JSON file as closely as possible, while using Markdown syntax for readability.
+    """
+
+    def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
+        # Bail if not a Json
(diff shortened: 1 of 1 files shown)


ZeyuTeng96:

afourney:
Thanks for the PR! We appreciate it!

Both #219 and this PR propose alternative ways of converting JSON to Markdown.  
An alternative approach would also be to simply quote the json (perhaps pretty-printed) in a Markdown code block.

Ideally, we would have a clean way of registering/unregistering different converters, so that people can customize the library to their desire  (e.g., by picking from the 3 above-mentioned approaches). I think this is coming in the new year, and will discuss with my team when everyone is back.

Until then, I'm not sure I am ready to pick a default behavior yet (beyond treating it as plain text -- which is what was intended)

ZeyuTeng96:
ok

PR #169, opened 2024-12-20
Feature/ Add xls support

yeungadrian:
PR for Issue #137 

- Adding new XLS converter using same pandas -> html -> md flow, except with xlrd engine instead of openpyxl
- New test file, which copies the existing xlsx and converts it to xls
- Explicitly use openpxyl for xlsx files

PR #52, opened 2024-12-16
chore: xlsx improvements

hewliyang:
**updates Apr 3 2025**

added the following kwargs to Excel-like converters

```py
na_rep: Any = ""
remove_header_placeholders: bool = True
drop_empty_cols: bool = False
drop_empty_rows: bool = False
```

- only NaN-like (None, NaT) is scrubbed out by default to `na_rep=""`
- pandas header placeholders, `Unnamed:{i}` is scrubbed out by default to `na_rep=""`

sparsity options:
- fully empty columns can optionally be removed
- fully empty rows can optionally be removed

also refactored Xlsx and Xls parsers to be more DRY with a ExcelParser base class. this will make it easier to
- introduce more excel-like types
- modify the underlying parsing logic, to say use `calamine` instead. see #259 

**usage**

```py
from markitdown import MarkItDown

md = MarkItDown()

# todo: cli support for parser opts?
res = md.convert(
    "packages/markitdown/tests/test_files/test.xlsx",
    na_rep="🤗",
    remove_header_placeholders=True,
    drop_empty_cols=False,
    drop_empty_rows=False,
)

print(res.markdown)
```

```
## Sheet1
| Alpha | 🤗 | Beta | Gamma | Delta | 🤗 |
| --- | --- | --- | --- | --- | --- |
| 89.0 | 🤗 | 82 | 100.0 | 12.0 | 13.0 |
| 76.0 | 🤗 | 89 | 33.0 | 42.0 | 43.0 |
| 🤗 | 🤗 | 🤗 | 🤗 | 🤗 | 🤗 |
| 60.0 | 🤗 | 84 | 19.0 | 19.0 | 53.0 |
| 7.0 | 🤗 | 69 | 10.0 | 17.0 | 63.0 |
| 87.0 | 🤗 | 89 | 86.0 | 54.0 | 312.0 |
| 23.0 | 🤗 | 4 | 89.0 | 25.0 | 23.0 |
| 70.0 | 🤗 | 84 | 62.0 | 59.0 | 34.0 |
| 83.0 | 🤗 | 37 | 43.0 | 21.0 | 321.0 |
| 71.0 | 🤗 | 15 | 88.0 | 32.0 | 431.0 |
| 20.0 | 🤗 | 62 | 20.0 | 67.0 | 647.0 |
| 67.0 | 🤗 | 18 | 15.0 | 48.0 | 878.0 |
| 42.0 | 🤗 | 5 | 15.0 | 67.0 | 🤗 |
[...]

PR #259, opened 2025-01-03
combine xlsx and xls to excel, replace openpxyl/xlrd with calamine

yeungadrian:
Improvement to: https://github.com/microsoft/markitdown/pull/169

Swaps excel processing engine from openpyxl to python-calamine, converting XLSXConverter to more generic ExcelConverter

Python-Calamine (https://pypi.org/project/python-calamine/) supports .xls, .xlsx, .xlsm, .xlsb (also .ods)

Calamine is rust based and so should also provide a bump in performance. No stats available but have found it more reliable than openpyxl & xlrd from personal experience.

Argument against could be that is that pandas needs to be >=2.2 / python-calamine is not as well known.


hewliyang:
As of [Jan 19](https://pandas.pydata.org/docs/dev/whatsnew/v2.2.0.html), `calamine` is now supported as an engine in `pandas` - I don't see any reason not to just have one processor for `.xls, .xlsx, .xlsm, .xlsb , .ods` anymore

@afourney doesn't seem to be in favor of removing sparsity by default, so would have to give some flags as in #52.

PR #267, opened 2025-01-06
Set exiftool path explicitly.

afourney:
This PR disables the automatic detection of exiftool, instead requiring it to be set explicitly either via the `exiftool_path` parameter in the MarkItDown constructor or by reading from the 'EXIFTOOL_PATH' environment variable.

### DIFF
--- a/tests/test_markitdown.py
+++ b/tests/test_markitdown.py
@@ -277,9 +277,29 @@ def test_markitdown_local() -> None:
     reason="do not run if exiftool is not installed",
 )
 def test_markitdown_exiftool() -> None:
-    markitdown = MarkItDown()
+    # Test the automatic discovery of exiftool throws a warning
+    # and is disabled
+    try:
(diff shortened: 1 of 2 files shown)


gagb (approved):

tests/test_markitdown.py
```diff
@@ -341,8 +361,8 @@ def test_markitdown_llm() -> None:
 
 if __name__ == "__main__":
     """Runs this file's tests from the command line."""
-    test_markitdown_remote()
-    test_markitdown_local()
+    # test_markitdown_remote()
+    # test_markitdown_local()
     test_markitdown_exiftool()
-    test_markitdown_deprecation()
-    test_markitdown_llm()
```
keenranger, line 348: Hi, I noticed that the tests in test_markitdown_exiftool() were commented out in this commit. Was this an intentional change, or was it possibly included by mistake?

Issue #51, opened 2024-12-16
Why remove image dataurl, That`s useful


fab617:
```python
        # Remove dataURIs
        if src.startswith("data:"):
            src = src.split(",")[0] + "..."
```

When I convet a DOCX file, I noticed that every image lost. by trace code , docx converted to html properly and image store in html as dataurl, but when html convert to markdown, you removed that in def convert_img at line 120 whitout any config. Every image lost without any alert.

afourney:
Perhaps we can make this optional, but the reason they are removed is because we were feeding the text_content into LLMs (the primary motivation for development), and the base64-encoded images would overwhelm the context windows.

tccgogogo:
> 也许我们可以将其设为可选，但删除它们的原因是我们将text_content提供给 LLM（开发的主要动机），而 base64 编码的图像会压倒上下文窗口。



> Perhaps we can make this optional, but the reason they are removed is because we were feeding the text_content into LLMs (the primary motivation for development), and the base64-encoded images would overwhelm the context windows.

When I convert a PDF to Markdown using the following code, it only outputs the textual information without any image information.
md = MarkItDown(llm_client=client, llm_model="qwen-long")
result = md.convert("")
print(result.text_content)

fab617:
> Perhaps we can make this optional, but the reason they are removed is because we were feeding the text_content into LLMs (the primary motivation for development), and the base64-encoded images would overwhelm the context windows.

I see, maybe we could make another copy without images and feed it to LLMs, so user can get the full document.

afourney:
Fixed in #1140

PR #272, opened 2025-01-09
add options to keep data uris

VoidIsVoid:
It's useful to keep image data uris for later process.

see https://github.com/microsoft/markitdown/issues/51


VoidIsVoid:

PR #303, opened 2025-01-23
Add support for conversion via Document Intelligence

KennyZhang1:
This PR adds document intelligence support for markitdown. It allows users to pass in an endpoint to extract text using Document Intelligence instead of the built-in offline converters. If an endpoint is not provided, markitdown will default to using offline conversion


afourney (approved):
Fantastic contribution. Thanks.

gagb:
@KennyZhang1 it maybe worth adding an example in README. And a pointer on how to create an azure Doc Int endpoint.

PR #318, opened 2025-02-10
Cleanup and refactor, in preparation for plugin support.

afourney:
Cleanup and refactor, in preparation for plugin support.


gagb:
Can you please update the README?

packages/markitdown/README.md
gagb, line 1: This is a copy of root README, which does not seem right. This README need to just explain this package and point back to root README for more info.

packages/markitdown-sample-plugin/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```toml
+[project.entry-points."markitdown.plugin"]
+sample_plugin = "markitdown_sample_plugin"
+```
+
+Here, the value of `sample_plugin` can be any key, but should ideally be the name of the plugin. The value is the fully qualified name of the package implementing the plugin.
+
+Once the plugin package is installed (e.g., `pip install -e .`), MarkItDown will automatically discover register it for use.
+
```
gagb, line 60: Can you add an example of what code will a developer write to use this plugin? Presumably, they just will just execute `markitdown` command. But can you say that explicitly here? Also what happens with the Python API? An example of that here would also help.

afourney: There's no change to code or to the command.

I will add an option to turn plugins on or off. I'll add documentation about that.

packages/markitdown/README.md
afourney: It's a little challenging, because the README for markitdown is what get sent to PyPi... so I want it to be complete. Meanwhile the README for the repo, is what shows for github, so I want that to be complete too. Open to suggestions. Ideally, the PyPi readme would load from the root, but this is not possible.

gagb: Sounds good. They could be mirrors. But update install instructions and add details about plugins.

packages/markitdown-sample-plugin/README.md
gagb: Yes no change, but it would be useful to say that

PR #319, opened 2025-02-10
feat: support automatic language code determination with `guesslang`

lh0x00:
#### Summary
This PR enhances `_CustomMarkdownify` with a feature:

1. **Automatic Code Language Detection**:
   - Uses `guesslang` to detect programming languages in code snippets.
   - Provides a fallback if `guesslang` is not installed.

#### Changes
- Imported `Guess` from `guesslang` with a fallback.
- Added `code_language_callback` to detect code languages.

#### How to Test
1. Install `guesslang` via `pip install guesslang`.
2. Convert a document with code snippets.
3. Verify correct language detection.

Thank you for reviewing!

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,12 @@ result = md.convert("example.jpg")
 print(result.text_content)
 ```
 
+### Extensions
+
+#### Automatic Code Language Detection
+
+Install `guesslang` with the command `pip install guesslang` to enable automatic code language recognition, especially useful for converting documents from web pages.
+
 ### Docker
 
 ```sh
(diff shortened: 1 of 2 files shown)


afourney:
Love this idea, but I'm thinking of using [magika](https://github.com/google/magika) for type detection (See #1108). Since magika can do code language detection as well, would you consider updating this PR to use it instead (to avoid another dependency)?

You should be able to just do: 
```
from magika import Magika
m = Magika()
res = m.identify_bytes(b".... contents of the code block ...")
if res.status == "ok" and res.prediction.output.group in ["text", "code"]:
    language =  res.prediction.output.label
```

Or something like that.

lh0x00:
I rebase it with your `magika` branch, and update it as requested, it should be merged after `magika`. Thanks!

PR #324, opened 2025-02-11
Added priority argument to all converter constructors.

afourney:
Ensures that a priority value can be passed to all DocumentConverter types.


packages/markitdown/src/markitdown/_markitdown.py
```diff
@@ -103,6 +99,23 @@ def __init__(
         # Register the converters
         self._page_converters: List[DocumentConverter] = []
 
+        # Note: We have tight control over the order of built-in converters, but
```
gagb, line 102: Can we improve the discoverability of this by moving it to the doc string?

PR #326, opened 2025-02-12
feat: Add Whisper transcription support with speech_recognition fallback

AlexZhangji:
## Changes
- Added OpenAI Whisper support for audio transcription in `WavConverter` and `Mp3Converter`
- Implemented automatic fallback to speech_recognition if Whisper fails or non-OpenAI client passed in.
- Added audio transcription tests (commented out by default) with a short test.wav.


AlexZhangji:

tech-mayavi (approved):

PR #331, opened 2025-02-13
Add Support For PPTX Shape Groups (Fix in code design to not miss out on slide content)

C0dingMast3r:
There was currently 0 support for nested shapes, this aims to add that as well as a test case to support that in the test pptx


C0dingMast3r:

afourney:
Thanks for the PR. Could you possible create another to add a test string to the nested shapes. It would be great to include this in unit testing, so that we ensure we don't regress. @C0dingMast3r

Issue #65, opened 2024-12-16
Removing if condition at line 1074


Viddesh1:
https://github.com/microsoft/markitdown/blob/81e3f24acd0049a59cd2dcb2d01d0a98cc57c734/src/markitdown/_markitdown.py#L1066-L1075

```python
    def _append_ext(self, extensions, ext):
        """Append a unique non-None, non-empty extension to a list of extensions."""
        if ext is None:
            return
        ext = ext.strip()
        if ext == "":
            return
       
        extensions.append(ext)

``` 
Hello,  I am new to this project. Thanks for creating it.

Rather that if condition at line number 1074 we can directly added the ext to extensions.

afourney:
Hah, good catch. Looks like line 1074 is some debug code. I will review the PR, asap.

PR #861, opened 2025-02-16
add necessary imports

tanreinama:
in _get_llm_description function, base64 and mimetypes used, but not imported.

```
client = OpenAI()
md = MarkItDown(llm_client=client, llm_model="gpt-4o")
result = md.convert("example.jpg")
>>> NameError: name 'mimetypes' is not defined
```

in packages/markitdown/src/markitdown/converters/_image_converter.py line #65;

```
    def _get_llm_description(self, local_path, extension, client, model, prompt=None):                                          if prompt is None or prompt.strip() == "":
            prompt = "Write a detailed caption for this image."

        data_uri = ""
        with open(local_path, "rb") as image_file:
>>>            content_type, encoding = mimetypes.guess_type("_dummy" + extension)
            if content_type is None:
                content_type = "image/jpeg"
>>>            image_base64 = base64.b64encode(image_file.read()).decode("utf-8")
            data_uri = f"data:{content_type};base64,{image_base64}"
```

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_image_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_image_converter.py
@@ -4,6 +4,7 @@ from ._media_converter import MediaConverter
 import base64
 import mimetypes
 
+
 class ImageConverter(MediaConverter):
     """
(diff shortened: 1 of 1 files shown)


afourney:
Oooof, I was overly aggressive when refactoring it seems. Thanks for the fix.

Issue #81, opened 2024-12-17
UnboundLocalError: local variable 'res' referenced before assignment


aylitat:
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
[<ipython-input-2-84b22469b7a6>](https://localhost:8080/#) in <cell line: 4>()
      2 
      3 markitdown = MarkItDown()
----> 4 result = markitdown.convert("/content/TEST.pdf")
      5 print(result.text_content)

2 frames
[/usr/local/lib/python3.10/dist-packages/markitdown/_markitdown.py](https://localhost:8080/#) in _convert(self, local_path, extensions, **kwargs)
   1043                     error_trace = ("\n\n" + traceback.format_exc()).strip()
   1044 
-> 1045                 if res is not None:
   1046                     # Normalize the content
   1047                     res.text_content = "\n".join(

UnboundLocalError: local variable 'res' referenced before assignment

aylitat:
`# If we hit an error log it and keep trying
                try:
                    res = converter.convert(local_path, **_kwargs)
                except Exception:
                    error_trace = ("\n\n" + traceback.format_exc()).strip()

                if res is not None:
                    # Normalize the content
                    res.text_content = "\n".join(
                        [line.rstrip() for line in re.split(r"\r?\n", res.text_content)]
                    )
                    res.text_content = re.sub(r"\n{3,}", "\n\n", res.text_content)

                    # Todo
                    return res`
maybe here should some fix

FrancescoMasaia:
I'm also experiencing this error not on every pdf but just on some of them, and it doesn't finish the conversion to md file

sidesteps:
same as #40

obar1:
> same as [#40](https://github.com/microsoft/markitdown/issues/40)

sorry, why is this still open if this was merged ?

afourney:
This report applies to the original converter implementation. The conversion loop has since been substantially rewritten and no longer reads res after a failed conversion attempt in this way. I’m closing this as obsolete. Please open a new issue with a current traceback if a similar failure remains.

PR #1038, opened 2025-02-19
Fix UnboundLocalError in MarkItDown._convert

menezesandre:
Fixes https://github.com/microsoft/markitdown/issues/40, https://github.com/microsoft/markitdown/issues/81.
If the first converter raises an exception, then the `res` variable is not initialized and we get an `UnboundLocalError` when checking `if res is not None`.


adam-lagoda:
➕

afourney (approved):

afourney:
Thanks for the fix! Running tests and merging now.

PR #1072, opened 2025-02-28
Unable to convert HTML to Markdown

lh0x00:
Problem description:
- I encountered a case where I could not convert from HTML to Markdown.

Problem solved:
- Fixed the problem of the function inherited from `markdownify.MarkdownConverter` not having `current_tags` causing an error by using `kwargs`, also set the default value for `convert_as_inline`.

Extensions:
- Deduplicated the list of extensions.

### DIFF
--- a/src/markitdown/__main__.py
+++ b/src/markitdown/__main__.py
@@ -84,7 +84,9 @@ def main():
             )
         elif args.filename is None:
             raise ValueError("Filename is required when using Document Intelligence.")
-        markitdown = MarkItDown(exiftool_path=which_exiftool, docintel_endpoint=args.endpoint)
+        markitdown = MarkItDown(
+            exiftool_path=which_exiftool, docintel_endpoint=args.endpoint
+        )
     else:
         markitdown = MarkItDown(exiftool_path=which_exiftool)
 
--- a/src/markitdown/_markitdown.py
+++ b/src/markitdown/_markitdown.py
@@ -1703,7 +1703,7 @@ class MarkItDown:
     ) -> DocumentConverterResult:
         # Deduplicate the list of extensions
         extensions = list(set(extensions))
-        
+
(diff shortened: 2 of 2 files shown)


afourney:
Thanks. I temporarily fixed the issue by pinning the markdownify version, but your fix is the correct path to support markdownify 1.0.0.

Since Markdownify is used so extensively, I want to test this a little more before merging. But thanks for the fix!

lh0x00:
@afourney thanks, I just ran reformatted again. I'm using the live version via commit id for now, waiting for your merge notification.

afourney:
Hmm. It's failing to convert html still (docx gets rendered as html, then converted to markdown). When it fails, the backoff tries processing it as a zip file. Anyhow, something isn't quite right still (the solution in main, right now, is to pin the version, but let's see if we can fix this here)

lh0x00:
I found it ok, you can check it directly here: https://huggingface.co/spaces/lamhieu/docsifer
Screenshot 2025-02-28 at 15 12 56

afourney:
> I found it ok, you can check it directly here: https://huggingface.co/spaces/lamhieu/docsifer Screenshot 2025-02-28 at 15 12 56
> 
> ps: oh yeah I saw fails let me check it out.

You that's a cool UI. I'll check it out in the morning! 

In any case, I'm sleep deprived at the moment and not seeing the origin of this error--- but I know I fixed it before the refactor. I'll keep digging. Sorry for the delay.

afourney:
Ok, I found the bug and merged everything. This is important enough that I'll cut a release in the morning after a little more testing.

lh0x00:
thanks @afourney, I just realized it too but I use `extensions = list(dict.fromkeys(extensions))`, maybe your solution is better, thanks for merging and solving it, much appreciated. G9, bro.

PR #1077, opened 2025-02-28
Update README.md

SigireddyBalasai:
The old git clone command is raising an access rights error in my system 
image
so just changed the command


afourney:
It's because you haven't registered a public ssh key with Github. In any case, fair point.

gagb:
I am not sure this is necessary. Appreciate the PR.

PR #1079, opened 2025-02-28
[Draft] Exploring ways to allow Optional dependencies

afourney:


packages/markitdown/src/markitdown/converters/_pptx_converter.py
```diff
@@ -54,9 +64,21 @@ def convert(self, local_path, **kwargs) -> Union[None, DocumentConverterResult]:
         if extension.lower() != ".pptx":
             return None
 
-        md_content = ""
+        # Load the dependencies
+        if _dependency_exc_info is not None:
+            raise MissingDependencyException(
+                f"""{type(self).__name__} recognized the input as a potential .pptx file, but the dependencies needed to read .pptx files have not been installed. To resolve this error, include the optional dependency [pptx] or [all] when installing MarkItDown. For example:
```
gagb, line 70: You could consolidate this string into a var instead of repeating it?

afourney: Yeah. Good idea

afourney:
@Zahlii
@robfitzgerald
@casperdcl
@AdrianVollmer

LMK what you thing about this approach. Main thing is that I want the converts to print useful message when they think they could handle the file, but don't have the right dependencies.

(alternatively, you could simply not register the converters at all... but then I worry about discoverability for people not reading docs)

Zahlii:
I think this will work fine for me - I didn’t check all of the dependencies in detail (maybe there are some more that could be marked as optional?) but at least those that currently lead to CVE warnings are now optional.

afourney:
> I think this will work fine for me - I didn’t check all of the dependencies in detail (maybe there are some more that could be marked as optional?) but at least those that currently lead to CVE warnings are now optional.

There's certainly a balance here. Some are needed just to help identify what filetypes are in use... and those need to be around in all cases. Others help with some of the fall-back converters (plain text etc.). And more generally, html-related dependencies are used by many converters (because the libraries might output HTML rather than markdown). Plus it's generally nice to be able to convert web pages out of the box, I'd argue. I think I'm satisfied with this first cut add organization.

README.md
```diff
@@ -50,6 +51,28 @@ You can also pipe content:
...
+* `[all]` Installs all optional dependencies
+* `[pptx]` Installs dependencies for PowerPoint files
+* `[docx]` Installs dependencies for Word files
+* `[xlsx]` Installs dependencies for Excel files
+* `[xls]` Installs dependencies for older Excel files
+* `[pdf]` Installs dependencies for PDF files
+* `[outlook]` Installs dependencies for Outlook messages
+* `[az-doc-intel]` Installs dependencies for Azure Document Intelligence
+* `[audio-transcription]` Installs dependencies for audio transcription of wav and mp3 files
+* `[youtube-transcription]` Installs dependencies for fetching YouTube video transcription
```
casperdcl, line 74: Personal preference...

```suggestion
* `[audio]` Installs dependencies for audio transcription of wav and mp3 files
* `[youtube]` Installs dependencies for fetching YouTube video transcription
```

casperdcl (approved):
lgtm

README.md
afourney: Agree shorter is better. However, we can still get metadata for YouTube (title, video descriptions etc.) even if the transcription library is not installed. Likewise we can get metadata for audio files (runtime, track title, artist, album, etc) even when transcription is not enabled. To this end, I felt it was worth the added characters for precision.

casperdcl: Strongly disagree; developer precision has nothing to do with user experience.

AdrianVollmer:
I like it

afourney:
K, merging to main... but I'll hold of from releasing to PyPi or cutting a release until we've had a chance to shake-test this a little more, and address a few other potentially API-breaking changes in the works.

PR #1088, opened 2025-03-05
Update converter API, user streams rather than filepaths

afourney:
A very-large PR that updates DocumentConverters in a couple of major ways:

* split responsibilities of deciding if a file *can* be converted  `DocumentConverter.accepts()` vs performing the conversion itself `DocumentConverter.convert()`
* `file_stream`, and `stream_info` inputs everywhere  (rather than local_paths). *No temporary files are used anymore.*
* Mimetypes (in addition to file extensions) are considered when electing to convert a file 

Adapted from #1045 , Thanks @KennyZhang1


packages/markitdown/src/markitdown/_base_converter.py
```diff
@@ -0,0 +1,142 @@
...
+        plugins can register converters in any order. A converter's priority
+        field reasserts some control over the order of converters.
+
+        Plugins can register converters with any priority, to appear before or
+        after the built-ins. For example, a plugin with priority 9 will run
+        before the PlainTextConverter, but after the built-in converters.
+        """
+        self._priority = priority
+
+    def accepts(
```
gagb, line 80: Is there a reason you didn't make it a class method?

packages/markitdown/src/markitdown/_base_converter.py
```diff
@@ -0,0 +1,142 @@
...
+        make a determination (e.g., special converters for Wikipedia, YouTube etc).
+        Finally, it is conceivable that the `stream_info.filename` might be used to in cases
+        where the filename is well-known (e.g., `Dockerfile`, `Makefile`, etc)
+
+        NOTE: The method signature is designed to match that of the convert() method. This provides some
+        assurance that, if accepts() returns True, the convert() method will also be able to handle the document.
+
+        IMPORTANT: If this method advances the position in file_stream, it must also reset the position before
+        returning. This is because the convert() method may be called immediately after accepts().
+
```
gagb, line 99: I am not sure what this is.

afourney: `file_stream` is a file-like object, and every read operation advances its position internally (like file.read()). In some cases -- like the OutlookMessageConverter -- additional stream read operations are needed to decide if the file is indeed an outlook message (the mimeype and extension are overly generic). If you read from the stream in the `accept()` methodm you need to return to the original position before exiting to leave the stream intact. 

E.g., 
```python
cur_pos = file_stream.tell()
data = file_stream.read()
file_stream.seek(cur_pos)
```

packages/markitdown/src/markitdown/converters/_zip_converter.py
```diff
@@ -46,99 +56,59 @@ class ZipConverter(DocumentConverter):
     """
 
     def __init__(
-        self, priority: float = DocumentConverter.PRIORITY_SPECIFIC_FILE_FORMAT
+        self,
+        priority: float = DocumentConverter.PRIORITY_SPECIFIC_FILE_FORMAT,
+        *,
+        markitdown: Any,
```
gagb, line 62: What is this? Can you improve the documentation? Not sure why the type Any makes sense here?

packages/markitdown/src/markitdown/_base_converter.py
afourney: MarkItDown registers converters as instances (this -- in theory -- allows the converters to be configured to take important parameters like API keys, etc). The decision to accept a `file_stream` may depend on this configuration. In general `accepts` and `convert`, take exactly the same parameters by design, so that exactly the same information is available at both stages of execution.

packages/markitdown/src/markitdown/converters/_zip_converter.py
afourney: It should be type  MarkItDown. (It's a reference to allow recursive conversions of nested files)

gagb: Should we add that type?

packages/markitdown/src/markitdown/_base_converter.py
gagb: It would be great to add this sentence that you just said "In some cases -- like the OutlookMessageConverter -- additional stream read operations are needed to decide if the file is indeed an outlook message"

gagb (approved):
Fantastic PR. Thank you!

packages/markitdown/src/markitdown/_base_converter.py
gagb: I see that makes sense -- the decision to accept may depend on the instance specific parameters.

packages/markitdown/src/markitdown/converters/_zip_converter.py
afourney: It's a bit squirrely as it involves a circular import -- but I've since added it.

packages/markitdown/src/markitdown/_base_converter.py
afourney: Done.

PR #1104, opened 2025-03-07
feat: sort pptx shapes to be parsed in top-to-bottom, left-to-right order

richardye101:
Currently, markitdown parses pptx shapes in Z-order, the order in which shapes are stacked on top of each other starting from the back to the front. 

There are repos that parse pptx to markdown which read the shapes in a normal reading order (top-to-bottom, left-to-right order) like https://github.com/ssine/pptx2md/blob/39bef65b312035baeade932aad8d221e37daae5f/pptx2md/parser.py#L249.

There are also stackoverflow posts that explain how to implement this code: https://stackoverflow.com/questions/51999656/how-to-extract-text-from-powerpoint-text-boxes-in-their-order-within-the-presen

I've simply copied over what @ssine has created in his repo, as it's the cleanest implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ MarkItDown is a lightweight Python utility for converting various files to Markd
 At present, MarkItDown supports:
 
 - PDF
-- PowerPoint
+- PowerPoint (reading in top-to-bottom, left-to-right order)
 - Word
 - Excel
 - Images (EXIF metadata and OCR)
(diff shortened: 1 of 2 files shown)


richardye101:

afourney:
Very nice simple fix!

afourney:
Thanks! Nice and simple fix.

PR #1111, opened 2025-03-10
Added FastAPI server to handle file conversion to markdown

dev-myk:
This PR introduces a FastAPI server that allows users to upload a file, which is then converted to Markdown using the Markitdown library.

What I Added:
- A new server.py file implementing a FastAPI-based API.
- A /convert endpoint that accepts file uploads and returns the Markdown conversion result.
- Integration of the Markitdown library to process file contents.

How It Works:
- Users send a POST request to /convert with a file as input.
- The server reads the file content and passes it to Markitdown for conversion.
- The API returns a JSON response containing the Markdown text.

### DIFF
--- a/server.py
+++ b/server.py
@@ -1,0 +1,15 @@
+from fastapi import FastAPI, File, UploadFile
+import markitdown
+
+app = FastAPI()
+
+@app.post("/convert")
+async def convert_to_markdown(file: UploadFile = File(...)):
+    content = await file.read()  # Read the file content
+    markdown_text = markitdown.convert(content.decode("utf-8"))  # Convert to markdown
+    return {"markdown": markdown_text}
+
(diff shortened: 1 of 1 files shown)


afourney:
Thanks. I've been thinking about something like this -- especially for MCP use. But I've not yet decided how to proceed. I think I might advocate for a separate package. Let me think about this a little.

afourney:
There's now an [mcp server](https://github.com/microsoft/markitdown/tree/main/packages/markitdown-mcp)

A more general FastAPI server would also be useful, perhaps. But I think I would recommend hosting that as a separate repo/project similar to how plugins are handled currently.

Floger911 (approved):

afourney:

PR #1120, opened 2025-03-12
Refactored tests.

afourney:


afourney:
Linking the special casing of 'test_mskanji.csv' in the stream tests to: https://github.com/google/magika/issues/983

Confirmed via MD5 sum, and minimal test reproduction (see https://github.com/google/magika/issues/983) that the correct bytes are being sent.

packages/markitdown/tests/_test_vectors.py
```diff
@@ -0,0 +1,214 @@
+import dataclasses
+from typing import List
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class FileTestVector(object):
```
gagb, line 6: Curious why you call it a FileTestVector?

packages/markitdown/tests/test_cli_misc.py
gagb, line 1: Could we add a doc string saying which tests are considered misc?

packages/markitdown/tests/_test_vectors.py
afourney: When I called it [TestVector](https://en.wikipedia.org/wiki/Test_vector), it got caught up in pytests sweep of tests :)

packages/markitdown/tests/test_module_misc.py
gagb, line 1: Same. Add doc string specifying what is misc and which tests should be added to this module.

packages/markitdown/tests/test_module_vectors.py
```diff
@@ -0,0 +1,154 @@
...
+        url=test_vector.url,  # Mock where this file would be found
+    )
+    for string in test_vector.must_include:
+        assert string in result.markdown
+    for string in test_vector.must_not_include:
+        assert string not in result.markdown
+
+
+if __name__ == "__main__":
+    import sys
```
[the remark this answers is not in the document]
afourney: It's only needed as a harness when running outside of pytest, in the __main__ namespace. This is a common pattern I believe.

You can run the test by just doing `python test_module_vectors.py` ... which is helpful I find for debugging.

gagb (approved):

packages/markitdown/tests/test_module_vectors.py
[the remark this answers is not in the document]
gagb: Sorry, I am having trouble understanding.
What is the difference between adding an import in the __main__ vs top of the file?

Issue #1102, opened 2025-03-07
How get inage captioning in docx files?


DmitryDiTy:
Hey,
I tried to convert docx with images file to md, but It does not do captioning:

```python
from markitdown import MarkItDown
from openai import OpenAI

client = OpenAI(base_url="http://localhost:8000/v1", api_key="dummy")  # my local VLLM host
md = MarkItDown(llm_client=client, llm_model="microsoft/Phi-3.5-vision-instruct")

result = md.convert("file.docx")
print(result.text_content)
# .... ![](data:image/png;base64...) ....
```

What did I do wrong?

Thank you in advance for your reply!

tookdes:
Same issue here. A bug maybe, for .pptx or .jpg it works well.

BetterAndBetterII:
https://github.com/microsoft/markitdown/pull/1140  It has supported passing parameters keep_data_uri to preserve image information

@afourney

afourney:
I think to be consistent with pptx etc., the request is to have the images get automatically captioned (either with the alt-text from the Word doc itself, or LLM-generated).

This is indeed a discrepancy, and I will work to address it in a future PR.

klynwuu:
would recommend also allow to have <--- image -1 ---> with base64 code or URL to path , image place holders like Docling in converted markdown, so that at least we can apply vision LLM to get caption by reading information from these image place holders. Thanks

hxk1633:
What’s the status on this? I’m also not able get image captioning using LLMs for docx files.

edwin-mui:
Any updates so far? Automated image captioning on .docx, .pdf, would certainly be useful.

Issue #246, opened 2025-01-02
Images are not visible/useable after PPTX to md conversion


adamcvolz:
In the PPTX Converter convert function, this 'filename' placeholder is being placed inside of the content section for the md image object:
![Image](https://github.com/user-attachments/assets/5e2d8014-757e-4aed-9701-38352901d419)

Causing us to not have the ability to support images post process:
![Image](https://github.com/user-attachments/assets/b58e4535-b378-4508-af18-0d39edcd07a8)

In the shape.image object, we have content_type(file type), and blob(bytes). Could easily format a base64 encoded image here instead:

` img_b64_string = base64.b64encode(image_bytes).decode('utf-8') `
`markdown_image = f"data:{content_type};base64,{b64_string}"`
`md_content += (
                        "\n!["
                        + (alt_text if alt_text else shape.name)
                        + "](markdown_image)\n"
                    )`

This would allow us to process images post conversion. Not sure if it was decided b64 encoded bytes wasn't ideal, but that can easily be extracted/converted/uploaded remotely etc. to a user's needs, and could even pass a flag to make it optionally supported.

afourney:
Fixed in #1140

Issue #162, opened 2024-12-20
how to save image in the markdown


charliedream1:
The image in the markdown is represented as this: ![](data:image/png;base64...). How can I save images?

Misnad:
Can you show an example?

corolair-hazem:
up

infogulch:
I have documents with embedded images that I would like to convert to markdown, so this feature would be helpful for me.

Currently when exporting it exports like this (exact quote):

```
![LABEL - TMG.jpg](data:image/jpeg;base64...)
```

I could see two resolutions:

1. Complete the embedded image data `![](data:image/png;base64...)`. I think this would work display correctly when rendering markdown.
2. Add an option to export embedded images as files in a particular directory and create relative links to them.

oliviermills:
Same problem, base64 data shoul be in the data-url or file saved, neither are there.. just `![](data:image/png;base64...)`  ie.. "..." were the base64 image data should be

tjyangsg:
Same problem. I convert word to md, all the image embedded in the document convert to just `![](data:image/png;base64...)`

SnakeHacker:
Refer to [issues51](https://github.com/microsoft/markitdown/issues/51)


Modify the file packages/markitdown/src/markitdown/converters/_markdownify.py, comment out the following code.
```
# if src.startswith("data:"):
#     src = src.split(",")[0] + "..."
```

<img width="893" alt="Image" src="https://github.com/user-attachments/assets/9efcb51d-b8a9-4a2d-be58-28d4397db34d" />

Reinstall it and you're done.
```
pip install -e packages/markitdown
```

Now it works!

tjyangsg:
***@***.***>谢谢！

charliedream1:
What's the meaning of view on web? image can't be visit ed locally?



---Original---
From: ***@***.***&gt;
Date: Mon, Feb 17, 2025 16:24 PM
To: ***@***.***&gt;;
Cc: "Optimus ***@***.******@***.***&gt;;
Subject: Re: [microsoft/markitdown] how to save image in the markdown (Issue#162)


   
Refer to issues51
 
Modify the file packages/markitdown/src/markitdown/converters/_markdownify.py, comment out the following code.
 # if src.startswith("data:"): #     src = src.split(",")[0] + "..."  image.png (view on web) 
Reinstall it and you're done.
 pip install -e packages/markitdown  
Now it works!

—
[...]

Issue #58, opened 2024-12-16
Both the image in docx and pdf will not be converted to base64 encoded content


GamerNoTitle:
Hello, I've found this repo and it is awesome to use this to convert some document to markdown. But when I use this tool to convert `docx` or `pdf` to markdown, the image in the file cannot be convertted currectly.

```log
(venv) PS F:\Git\GamerNoTitle\Doc2Markdown> python
Python 3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from markitdown import MarkItDown
>>> handler = MarkItDown()
>>> docx_result = handler.convert(r"C:\Users\GamerNoTitle\Desktop\testdoc.docx")
>>> docx_result.text_content
'\n\n![卡通人物\n\n低可信度描述已自动生成](data:image/jpeg;base64...)\n\nThe content above is a image.\n\n'
>>> pdf_result = handler.convert(r"C:\Users\GamerNoTitle\Desktop\testdoc.pdf")
>>> pdf_result.text_content
'The content above is a image.\n\n'
```

The file as follows
- [testdoc.docx](https://github.com/user-attachments/files/18148927/testdoc.docx)
- [testdoc.pdf](https://github.com/user-attachments/files/18148928/testdoc.pdf)

The version of MarkItDown I used is 0.0.1a2, which is installed from pypi.org using `pip install markitdown`

afourney:
Better handling of images in documents is certainly something I would like to support, but just to be clear, what is your expected behavior here?

* Should the images be saved to disk, and referenced in Markdown?
* Should they be included inline as base64? (I would argue against this, since it makes the text unsuitable for a variety of downstream applications like indexing, or use in llms)
* Should the images be sent through the same pipeline as .jpg etc, where we extract tags and other metadata?

zrh535:
My preference would be for the first option. Save to disk and reference in Markdown. I've been the issues base64 can cause and would suggest avoiding that one as well. Perhaps it could be a flag to toggle between extract or LLM pipeline?

gagb:
Duplicate of #56

GamerNoTitle:
> Better handling of images in documents is certainly something I would like to support, but just to be clear, what is your expected behavior here?
> 
> * Should the images be saved to disk, and referenced in Markdown?
[...]

PR #1140, opened 2025-03-18
Add support for preserving base64 encoded images

BetterAndBetterII:
## Issue Description
Currently, the markitdown library automatically truncates base64-encoded image data when converting HTML, DOCX, and PPTX files. This behavior may not be ideal for certain use cases, especially when users want to preserve complete image data in the converted Markdown documents.

Related: 
https://github.com/microsoft/markitdown/issues/51
https://github.com/microsoft/markitdown/issues/58
https://github.com/microsoft/markitdown/issues/162
https://github.com/microsoft/markitdown/issues/246
https://github.com/microsoft/markitdown/issues/1102
https://github.com/microsoft/markitdown/issues/1139

## Solution
Add a new parameter `keep_data_uris` that allows users to control whether base64-encoded image data is preserved during the conversion process.

## Changes Made

1. **Modified `_CustomMarkdownify` class** (in `_markdownify.py`):
   - Added a new parameter `keep_data_uris` (default: `False`) to control whether to preserve base64 data URIs
   - Updated the `convert_img` method to respect this parameter when processing image sources

2. **Updated `HtmlConverter` class** (in `_html_converter.py`):
   - Modified the `convert` method to accept and pass through the `keep_data_uris` parameter from `kwargs`
   - Ensured this parameter is passed to the `_CustomMarkdownify` constructor

3. **Updated `DocxConverter` class** (in `_docx_converter.py`):
   - Modified the `convert` method to extract and pass the `keep_data_uris` parameter to the HTML converter

4. **Updated `PptxConverter` class** (in `_pptx_converter.py`):
   - Added conditional logic to handle images differently based on the `keep_data_uris` parameter
   - When `keep_data_uris=True`, encode image blobs as base64 and include them directly in the Markdown
   - When `keep_data_uris=False` (default), continue using placeholder filenames

5. **Updated auxiliary converters**:
   - **`BingSerpConverter`**: Added parameter forwarding for `keep_data_uris`
   - **`RssConverter`**: Added `__init__` method to store `kwargs` and modified `_parse_content` to use the parameter
   - **`WikipediaConverter`**: Added parameter forwarding for `keep_data_uris`

6. **Verified `YouTubeConverter`**:
   - Confirmed that despite importing `_CustomMarkdownify`, it does not actually use it for conversion

## Benefits

1. **User Control**: Users can now decide whether to include base64-encoded images in the output
2. **Consistency**: All converters now handle the parameter in a unified way
3. **Backward Compatibility**: The default behavior remains unchanged (truncating base64 data)

## Usage Example

```python
# To preserve base64 images when converting
result = converter.convert(
    file_stream=file,
    stream_info=stream_info,
    keep_data_uris=True
)

# To continue truncating base64 images (default behavior)
result = converter.convert(
    file_stream=file,
    stream_info=stream_info
)
```

This enhancement ensures a consistent approach to handling base64 image data across all converters in the markitdown library, providing flexibility while maintaining backward compatibility.


afourney:
Hey this looks great. Thanks for the contribution.

Would you be able to add some test cases to verify this behavior in the CI. Once we have reasonable test coverage, I think it's good to merge.

packages/markitdown/src/markitdown/converters/_bing_serp_converter.py
```diff
@@ -79,7 +79,9 @@ def convert(
             slug.extract()
 
         # Parse the algorithmic results
-        _markdownify = _CustomMarkdownify()
+        _markdownify = _CustomMarkdownify(
+            keep_data_uris=kwargs.get("keep_data_uris", False)
+        )
```
hoangquochung1110, line 84:
```suggestion
        _markdownify = _CustomMarkdownify(
            **kwargs
        )
```
By this way, we lay a "foundation" so that we can easily add new options to `_CustomMarkdownify` without changing too much in upstream objects

packages/markitdown/src/markitdown/converters/_markdownify.py
```diff
@@ -17,6 +17,7 @@ class _CustomMarkdownify(markdownify.MarkdownConverter):
 
     def __init__(self, **options: Any):
         options["heading_style"] = options.get("heading_style", markdownify.ATX)
+        self.keep_data_uris = options.pop("keep_data_uris", False)
```
hoangquochung1110, line 20: likewise, we can

```suggestion
        options["keep_data_uris"] = options.get("keep_data_uris", False)
```
by this way, we have:
- Default value in case the option is not supplied in upstream objects
- Keep the implementation consistent

What do you think ?

BetterAndBetterII: Absolutely! 🫡

BetterAndBetterII: I have modified it in the form of kwargs passing parameters. For the sake of uniformity, I used the kwargs passing parameter form in all places where _CustomMarkdownify is initialized, and also reserved adjustment space for adding more optional parameters later.

BetterAndBetterII:
> Hey this looks great. Thanks for the contribution.
> 
> Would you be able to add some test cases to verify this behavior in the CI. Once we have reasonable test coverage, I think it's good to merge.

I have added tests for the cli and module calls. I mainly focus on office files such as word and powerpoint. Since Excel uses the pandas library to read, it does not support reading any images, so it is not possible, and I do not plan to add tests to retain images.

packages/markitdown/src/markitdown/converters/_markdownify.py
BetterAndBetterII: Since the command line parameter parser must declare parameters before parsing, kwargs cannot be used, so __main__.py still uses a fixed actual parameter passing method.

afourney:
Ok, there are some formatting issues, and I want to make one small change to how the tests are organized. Let me do that really quickly. Stay tuned.

afourney:
@BetterAndBetterII I fixed some formatting issues, and reorganized tests a little. If this looks good to you, let me know and I'll merge it.

afourney (approved):

BetterAndBetterII:
> @BetterAndBetterII I fixed some formatting issues, and reorganized tests a little. If this looks good to you, let me know and I'll merge it.

great!

harsh-devx:
quick question here, is the support for `keep_data_uris` is not extended for pdf converter?

iocoker:
> quick question here, is the support for `keep_data_uris` is not extended for pdf converter?

I think the tool is predominantly built for `.docx.`, `.ppt` files. Even basic styling is ignored in `.pdfs` and no images. You will only get plain texts from pdfs

afourney:
> > quick question here, is the support for `keep_data_uris` is not extended for pdf converter?
> 
> I think the tool is predominantly built for `.docx.`, `.ppt` files. Even basic styling is ignored in `.pdfs` and no images. You will only get plain texts from pdfs

That is correct. Images in office files are their own entities and are easy to extract. PDFs, on the other hand, make things much hard (which is typically the case).

I have it on my to-do list to integrate or provide options to use much much better PDF parsers than what we currently ship (and indeed Azure Document Intelligence will already yield better results for PDFs). But until then, we basically are stuck with the text layer of the PDFs.

PR #1151, opened 2025-03-24
Make it easier to use AzureKeyCredentials with Azure Doc Intelligence

afourney:
Allow users to use key-based authentication to Azure, and not just Azure Active Directory.

Added finer-grained control over which of the supported types are sent to Doc Intel.


saibhaskerraju:
although this PR is merged,  I guess its not pubished. we are not seeing these changes in latest version. please check

PR #1161, opened 2025-03-28
Handle PPTX shapes where position is None

richardye101:
My [previous PR](https://github.com/microsoft/markitdown/pull/1104) for sorting powerpoint shapes relied on the `top` and `left` attributes of the pptx shapes, however I didn't account for shapes having `None` as their `top`, `left`, `height` or `width` attributes. We'll need to think about what to do with these shapes, as sometimes they can still contain text, however they don't exist on the slide. 

Currently, I've filtered out shapes with `None` in their positional attributes, however I'm open to changing how they're dealt with.

I think the options are:

1. Parse them into markdown in a "Hidden Text" section under the Notes
2. Ignore these shapes as I have written, as they don't appear on the powerpoint anyway so anyone reading it wouldn't know that text exists


richardye101:
From my research, the only shapes with `None` in their `top,left,height,width` attributes are of type `pptx.shapes.placeholder.SlidePlaceholder`. I'm not sure why the sort key would result in a stack overflow error though... 

To your point about listing the shapes last: It seems almost all the shapes with missing coordinates are slide titles using placeholder shapes inserted using a slide template. I've seen some instances where a placeholder for slide content is also missing coordiantes. Based on that, i would actually be inclined to list them first

afourney:
> From my research, the only shapes with `None` in their `top,left,height,width` attributes are of type `pptx.shapes.placeholder.SlidePlaceholder`. I'm not sure why the sort key would result in a stack overflow error though...
> 
> To your point about listing the shapes last: It seems almost all the shapes with missing coordinates are slide titles using placeholder shapes inserted using a slide template. I've seen some instances where a placeholder for slide content is also missing coordiantes. Based on that, i would actually be inclined to list them first

First also makes sense to me.

 You can click on the test logs to see the detailed error, but here's the trace:
image

richardye101:
I fixed the recursion error, was a mistake in copying the sorting code between grouped shapes and normal shapes

cfu94:
> I fixed the recursion error, was a mistake in copying the sorting code between grouped shapes and normal shapes

Hi, is there an ETA for merging this PR? Some of our PowerPoints currently cannot be converted with markitdown because of this bug

Issue #167, opened 2024-12-20
merged cell convert issue,both in excel and pptx


yutaixi:
The merged cells in Excel, after being converted to Markdown, are split, resulting in a table with new empty cells.
before convert
![Image](https://github.com/user-attachments/assets/3de20159-9bd7-4bf2-a7d9-404e9c0195aa)
after convert
<img width="871" alt="Image" src="https://github.com/user-attachments/assets/d86e454c-6ddf-4dac-9c26-939908c03be5" />

GaboEspadas:
Markdown tables do not allow for merged cells. This is not related to markitdown but the markdown language itself. 
HTML allows for this flexibility. It will be more flexible to have a xml-it tool besides the markdown.
Anyhow, I think markdown can also support custom html so perhaps the developers can eventually use the xml for tables.

BetterAndBetterII:
I also encountered similar problems when parsing tables. My work is for very large and complex tables with a lot of merged cells with a large span. 

If there is no option to parse merged cells, my table is basically unreadable. 

I added an optional option in pr #1165 to support parsing merged cells and headers in Excel and filling values ​​into child cells. 

Combined with the expansion of parent and child items of the table that I implemented myself, this can greatly improve LLM's understanding of the table.

![Image](https://github.com/user-attachments/assets/16ca6767-a6ce-4a65-a9b2-90c7c7c214b9)

![Image](https://github.com/user-attachments/assets/de9eecdd-fec8-4ab5-b423-efbbf35a6202)

cooleel:
Same issue here. Were you guys able to figure it out?

phamkhactu:
@wei12314 Have you tried? If yes, Do you need to custome code in some files which is committed by BetterAndBetterII 
 in 4 commits?

PR #1165, opened 2025-04-01
arg: fill_merged_cells. Optional to fill merged cells instead of NAN

BetterAndBetterII:
issue: #167

I encountered problems when parsing Excel tables. My work is for very large and complex tables with a lot of merged cells with a large span. 

If there is no option to parse merged cells, my table is basically unreadable. 

I added an arg fill_merged_cells to support parsing merged cells and headers in Excel and filling values ​​into child cells. 

Image

Image


BetterAndBetterII:
btw: how about considering creating a new way to test the various parameters of convert()?

ramnathk:
I'd love to see this getting merged in!

PR #1172, opened 2025-04-08
fix(README): correct pip install

FacerAin:
Corrected README by adding quotes in the pip command. 
While it's a small tweak, the original command fails in zsh environments.
Thanks 😊


afourney:
Thanks for catching the shell-quoting problem. Both installation-command corrections proposed here were incorporated in [#1175](https://github.com/microsoft/markitdown/pull/1175): the source-install command and the selective-extras example now quote the package argument.

Closing this as already addressed; thank you for the contribution

PR #1195, opened 2025-04-20
Fix #183: Blank cells in .xlsx would return NaN in the markdown.

### LINKED ISSUE #183: Blank cells in .xlsx would return NaN in the markdown.
I just thought Python's default `NaN` doesn't look visually good on the markdown results. There should be a way/an option for it to output an empty cell instead.

Simewu:
This minimal PR adds `na_rep=""` as an argument to `to_html` in pandas, resolving issue #183.

---

The discussions in #52 are working towards a solution that gives the user options, but since several weeks have passed and the issue is still in the base branch, a minimal fix may help.


fn5:
Hi, was there a reason this was closed unresolved?

Simewu:
Hello, I figured #52 was on track to a less hardcoded solution.

PR #1207, opened 2025-04-24
Fix python.lang.security.use-defused-xml-parse.use-defused-xml-parse--tmp-c83c1369-f7f5-4288-aea7-04caa47d28e4-packages-markitdown-src-markitdown-converter_utils-docx-math-omml.py

kira-offgrid:
This PR fixes python.lang.security.use-defused-xml-parse.use-defused-xml-parse--tmp-c83c1369-f7f5-4288-aea7-04caa47d28e4-packages-markitdown-src-markitdown-converter_utils-docx-math-omml.py.

### DIFF
--- a/packages/markitdown/pyproject.toml
+++ b/packages/markitdown/pyproject.toml
@@ -29,6 +29,8 @@ dependencies = [
   "markdownify",
   "magika~=0.6.1",
   "charset-normalizer",
+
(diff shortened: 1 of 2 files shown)


afourney:
Please elaborate on what issue is being resolved here. This project already takes a dependency on defusedxml. Is the concern that we're not enforcing a specific version?

PR #1176, opened 2025-04-09
Add CSV to Markdown table conversion - fixes #1144

erinshek:
## Description
This PR adds support for converting CSV files to Markdown tables. The converter handles various CSV formats and edge cases while maintaining proper table structure.

## Features
- Convert CSV files to properly formatted Markdown tables
- Support for text/csv and application/csv MIME types
- UTF-8 encoding with error replacement for invalid characters
- Preserve table structure with headers and data rows
- Handle edge cases:
  - Empty cells
  - Rows with mismatched column counts
  - Special characters in content

## Example
Input CSV:
```csv
Name,Age,City
John,30,New York
Alice,25,London
```

Output Markdown:
```markdown
| Name  | Age | City     |
|-------|-----|----------|
| John  | 30  | New York |
| Alice | 25  | London   |
```

### Command Line
```bash
markitdown input.csv > output.md
```

### Python Library
```python
from markitdown import MarkItDown

converter = MarkItDown()
result = converter.convert("input.csv")
print(result.markdown)
```

## Changes
- Add new `CsvConverter` class
- Register converter in `MarkItDown` class
- Add CSV converter to available converters
- Fix Azure Document Intelligence dependency handling

## Testing
- Tested with various CSV formats
- Verified UTF-8 encoding handling
- Checked edge cases with empty cells and mismatched columns
- Validated Markdown table formatting

## Related Issues
Closes #1144

PR #1215, opened 2025-04-26
Create csvconvertor.py

rhwang1226:


afourney:
Thank you for proposing CSV-to-Markdown support. A CSV converter was implemented and registered in [#1176](https://github.com/microsoft/markitdown/pull/1176), so CSV files already use the built-in conversion pipeline.

Any remaining CSV parsing or rendering edge cases should be reported against the existing converter.

Issue #1232, opened 2025-05-01
'YouTubeTranscriptApi' object has no attribute 'fetch' when try to convert youtube videos to transcription in md files


claudiocassimiro:
When running the code, I receive some informations but the transcript operation fail:

python3 test.py
Attempt 1 failed: 'YouTubeTranscriptApi' object has no attribute 'fetch'
Attempt 2 failed: 'YouTubeTranscriptApi' object has no attribute 'fetch'
Attempt 3 failed: 'YouTubeTranscriptApi' object has no attribute 'fetch'
Error fetching transcript: Operation failed after 3 attempts.
# YouTube

### Video Metadata
- **Keywords:** programming, tech, dev, developer, javascropt, reagindo algo, reacao, programador, programação, tecnologia, startup, notícias, mercado de tecnologia, análise e desenvolvimento de sistemas, ads, javascript, typescript, lucas montano, código fonte tv, gustavo guanabara, FILIPE DESCHAMPS, aws, google cloud, vps, vpn, iphone, android
- **Runtime:** PT14M57S

### Description
Se inscreva para a próxima chorumeet:
https://meetup.chorume.club

✅ PATROCINADOR OFICIAL DOS CORTES
AUVP - O melhor curso de investimentos para desenvolvedores
https://sard.ink/449LOMt

Post: (Post Original Excluído)

🤘 Seja membro e ajude manter o canal
   / @manodeyvin

⭐️ Me siga nas outras redes
https://chorume.dev

JoshClark-git:
You need to specify a language if there is no english transcript available for the video. In this instance your convert statement would look like `result = md.convert("https://www.youtube.com/watch?v=6bGEPfwS9NA",youtube_transcript_languages = ['pt'])`

PR #1241, opened 2025-05-08
FIX YouTube transcript errors

JoshClark-git:
In response to #1232 

If no transcript language specified by input, and English isn't an available language, default to first available transcript in alphabetical order. 

If specified transcript language isn't available, translate available transcript into specified language.


JoshClark-git:

afourney:
Fantastic. Thanks.

PR #1261, opened 2025-05-22
Replaced the code derived from `dwml` with an alternative impl

t3tra-dev:
## Description

This pull request removes the dependency on the **`dwml`** library by re-implementing the functionality to convert mathematical formulas (OMML) in DOCX files to LaTeX format. This change resolves license compatibility issues and aims to provide a cleaner, more maintainable implementation.

### Testing

- All existing tests pass without modification
- Verified no functionality changes

### Implementation Details

- removed the `dwml` library's associated code
- re-implemented the OMML to LaTeX conversion logic with an object-oriented approach
- introduced new modules for handling the conversion:
  - `latex_symbols.py`: manages mappings for symbols, functions, and LaTeX commands
  - `omml_elements.py`: defines classes representing OMML elements
  - `omml_parser.py`: parses OMML XML into a tree of `omml_elements.py` instances
  - `omml_to_latex.py`: orchestrates the conversion process
- updated `pre_process.py` to use the new conversion modules
- deleted `ThirdPartyNotices.md`, because the licensing issue has been resolved

### Migration Notes

This is a non-breaking change as all public APIs remain unchanged.


afourney:
Thanks for the work. 

Can you describe the perceived license compatibility issue? My understanding is that Apache 2 and MIT are compatible, provided we include the requisite information and attributions in the ThirdPartyNotices file (which we are happy to do).

afourney:
I see. Ok this makes sense to me. 

One reason I haven't published a new release in a while is that I was nervous about this extra document processing on docx. It all seems to work fine, but rewriting the zip contents of docs, in-flight, felt like it warranted extra caution. I'm satisfied now that the general approach and current code is stable, and will release it to the main channel. Once we've gained some broader experience with this, I will be more comfortable with this PR as well. I agree that the benefits you are proposing are worthwhile.

PR #1283, opened 2025-06-04
Added `keep_data_uris` support for Mcp.

Sergeant61:
Added `keep_data_uris` support for Mcp.

### DIFF
--- a/packages/markitdown-mcp/src/markitdown_mcp/__main__.py
+++ b/packages/markitdown-mcp/src/markitdown_mcp/__main__.py
@@ -16,11 +16,14 @@ import uvicorn
 # Initialize FastMCP server for MarkItDown (SSE)
 mcp = FastMCP("markitdown")
 
+# Global variable for keep_data_uris setting
+KEEP_DATA_URIS = False
+
 
 @mcp.tool()
-async def convert_to_markdown(uri: str, keep_data_uris: bool = False) -> str:
+async def convert_to_markdown(uri: str) -> str:
     """Convert a resource described by an http:, https:, file: or data: URI to markdown"""
-    return MarkItDown(enable_plugins=check_plugins_enabled()).convert_uri(uri, keep_data_uris=keep_data_uris).markdown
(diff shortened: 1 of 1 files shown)


afourney:
@Sergeant61 putting this argument here will require the model to turn the feature on each time it calls the tool.

(I'm also super conservative about how/when to expand the action space)

Would a server configuration approach work for your scenario? E.g. a config file, or command line argument for the server that turns this on for all calls?

Sergeant61:
Actually, we added it like that at the beginning, but my teammate and I thought that passing it as a parameter would create a more dynamic structure. In some files we want this, in some we don't, then shouldn't we restart the mcp server every time?

Sergeant61:
We discovered this (keep_data_uris) feature while reviewing PRs with my friend. We haven't examined all the features yet, but honestly, I’d like to see features like this added to the MCP server. How do you plan to incorporate such features into the MCP server? I can update the PR accordingly. However, it might be necessary to think about this in a broader scope.

afourney:
> We discovered this (keep_data_uris) feature while reviewing PRs with my friend. We haven't examined all the features yet, but honestly, I’d like to see features like this added to the MCP server. How do you plan to incorporate such features into the MCP server? I can update the PR accordingly. However, it might be necessary to think about this in a broader scope.

My sense is that we can create some config options, or similar, for the server that would set things like this on startup. My concern about adding it to the function definition that the LLMs see is that the LLMs then need to consider it, and guess the right action, on every LLM call. But if we just configure a flag when starting the server, you get deterministic behavior on that front, and it doesn't confuse the LLM.

But this only works if you downstream application ALWAYS or NEVER wants to save data uris. If it's case-by-case, then the current solution is needed. My gut feel, though, is that ALWAYS or NEVER cover most applications.

Sergeant61:
Okay then I will update the pr as you say as soon as possible.

Sergeant61:
You can now pass parameters via the command line.
For example:

```bash
markitdown-mcp --http --host 127.0.0.1 --port 3001 --keep-data-uris
```

PR #1285, opened 2025-06-06
Extending LLM usage for PDFs when the extracted text is empty after pdfminer

gjmveloso:
Initial work to attempt to use LLM to perform OCR operations within a PDF when `pdfminer` returns empty text

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_pdf_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_pdf_converter.py
@@ -530,7 +530,11 @@ class PdfConverter(DocumentConverter):
                     extension=".pdf",
                     feature="pdf",
                 )
+<<<<<<< HEAD
             ) from _dependency_exc_info[1].with_traceback(
+=======
+            ) from _dependency_exc_info[1].with_traceback(  # type: ignore[union-attr]
+>>>>>>> 117ffa2 (Extending LLM usage for PDFs where the extracted text was empty with pdfminer)
                 _dependency_exc_info[2]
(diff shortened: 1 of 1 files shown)


gjmveloso:

packages/markitdown/src/markitdown/converters/_pdf_converter.py
```diff
@@ -65,13 +64,30 @@ def convert(
...
+                original meaning. Only a valid Markdown-formatted output is allowed."""
+                markdown = llm_caption(
+                    file_stream,
+                    stream_info,
+                    client=llm_client,
+                    model=llm_model,
+                    prompt=llm_prompt,
+                )
+
+        return DocumentConverterResult(markdown=str(markdown))
```
dillonstreator, line 93: There is an issue of PDFs containing both mineable text and images that contain text. It would be nice to have a more sophisticated branching mechanism that accounts for this and/or allowing an API to override by the markitdown caller.

gjmveloso: Are you thinking on something like replacing the usage of `extract_text` with `extract_pages` and iterate over its non-text elements, like `LTImage` and `LTFigure`?

Layout system reference:
https://pdfminersix.readthedocs.io/en/latest/topic/converting_pdf_to_text.html#topic-pdf-to-text-layout

gjmveloso: Finally, done. Thoughts?

PR #1295, opened 2025-06-17
Add language parameter for the audio transcription

hchmch:
Add the possibility to select a language for the transcription


hchmch:

PR #1296, opened 2025-06-17
fix: Make magika an optional dependency

mohiuddin-khan-shiam:
## Problem
The `magika` package was a hard dependency, causing import failures if not installed, even when basic functionality would work with Python's built-in `mimetypes`.

## Changes
- Made `magika` import optional with graceful fallback
- Updated [MarkItDown](cci:2://file:///d:/Github/markitdown/packages/markitdown/src/markitdown/_markitdown.py:95:0-784:26) to handle missing `magika` gracefully
- Maintained all existing functionality when `magika` is available
- Falls back to standard `mimetypes` when `magika` is not available

## Impact
- CLI and library now work without `magika` installed
- Better out-of-the-box experience for users
- More robust error handling for file-type detection
- No breaking changes to existing functionality


josteink:
Tagging @afourney .

Would it be possible to consider this pull-request for merging as well? It would solve a lot of problems for a lot of users.

PR #1307, opened 2025-07-03
Add the option to export data URIs to files.

mike-matera:
Hello! 

This pull request builds on #1140 by adding an option, `--export-data-uris`, that allows users to specify a path prefix where embedded images will be saved. The document will be generated with references to the saved files. The rationale is to make it so converted documents are easier to edit because they don't contain mountains of base64 encoded data. 

I realize this might not quite be in line with the goal of making the data available to an LLM but it really seems like a missing feature. 

Please let me know what you think.


mike-matera:

DanielOverdevest:
Exactly what I needed! Thanks @mike-matera

akumar9-godaddy:
Does it work for pdf as well ?

Jeandoom:
need it, i want to know when it would be published

Jeandoom (commented):
please merge this into main branch and quickly publish it

Issue #1164, opened 2025-03-31
Support for AsyncOpenai client


shoang22:
Hi, is there support for using an async Openai client for calls involving the llm_client?

afourney:
Yeah, the inference code needs an overhaul in general. It's something I'm thinking about how to tackle all-up.

clemlesne:
In part duplicate of https://github.com/microsoft/markitdown/issues/13.

PR #1309, opened 2025-07-07
perf: Migrate logic to async

clemlesne:
For backward compatibility, converters are added new "_async" methods, in complement of the original (sync). Sync method is marked deprecated. Youtube converted is executed in a thread context (the lib isn't natively async) and OpenAI integration use the AsyncOpenAI class, also marking sync class as deprecated.

Relates to #13 and #1164.


Swipe4057:
any updates?

PR #1319, opened 2025-07-10
fix: correctly pass custom llm prompt parameter

stefan-rink:
This fix ensures that the llm_prompt parameter, used by the PPTX and image converters for annotation, is correctly passed down to the converter. Previously it was not being forwarded.


stefan-rink:

PR #1325, opened 2025-07-14
Fix #53: Preserve Excel cell formatting (currency, percentage, etc.) …

### LINKED ISSUE #53: Unable to extract currency from excel formatted cells

Image

Returns:

preetlimbani:
Fix #53: Preserve Excel cell formatting (currency, percentage, etc.) during conversion

- Use openpyxl to access cell formatting information
- Apply number formats (currency, percentage, thousands separator)
- Maintain visual representation from Excel in Markdown output
- Add tests for currency formatting preservation

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
@@ -1,16 +1,21 @@
 import sys
-from typing import BinaryIO, Any
+from typing import BinaryIO, Any, Optional, Union, List, Dict
 from ._html_converter import HtmlConverter
 from .._base_converter import DocumentConverter, DocumentConverterResult
 from .._exceptions import MissingDependencyException, MISSING_DEPENDENCY_MESSAGE
 from .._stream_info import StreamInfo
+import re
+from datetime import datetime
+from decimal import Decimal
 
 # Try loading optional (but in this case, required) dependencies
 # Save reporting of any exceptions for later
 _xlsx_dependency_exc_info = None
 try:
     import pandas as pd
-    import openpyxl  # noqa: F401
+    import openpyxl
+    from openpyxl.utils import get_column_letter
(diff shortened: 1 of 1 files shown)


preetpatel26:

packages/markitdown/src/markitdown/converters/_xlsx_converter.py
```diff
@@ -36,18 +41,29 @@
...
+    - Percentage formatting (e.g., 25.5%)
+    - Number formatting with thousands separators (e.g., 1,234,567)
+    - Date formatting (preserves display format from Excel)
+    - Custom number formats
+
+    The converter uses openpyxl to access cell formatting information that is not
+    available through pandas.read_excel() alone.
     """
 
     def __init__(self):
```
dan-i-am: Thanks for tackling this - preserving cell formatting is a useful enhancement and something I was recently looking into myself.

However, this PR's approach is a breaking change for downstream consumers who parse numeric values from the markdown output. Code doing float(cell_value) will break when it encounters $1,234.56 instead of 1234.56.

I'd suggest making this opt-in via a config object pattern rather than a boolean flag. This keeps the API extensible without accumulating kwargs over time:

```
@dataclass
class XlsxConfig:
    preserve_formatting: bool = False
    # future options can be added without breaking changes

class XlsxConverter(DocumentConverter):
    def __init__(self, config: XlsxConfig | None = None):
        self._html_converter = HtmlConverter()
        self._config = config or XlsxConfig()

```

Then in **_format_cell_value()**, check **self._config.preserve_formatting** before applying formatting.

This pattern:

- Maintains backward compatibility (default = current behavior)
- Allows users to opt-in explicitly
- Scales cleanly as more converter options are added
- Works well with IDE autocomplete and type hints

The same pattern could be extended to other converters (PdfConfig, DocxConfig, etc.) if MarkItDown wants a consistent configuration approach.

PR #1338, opened 2025-07-16
fix : typo in security.md

yannickzheng:
Fixed redundant punctuation in the security.md file.


chienyuanchang:
Thanks for taking the time to tidy this up. This change is a small wording/punctuation tweak in our support guidance, and we usually batch these kinds of doc-only edits together rather than merging them on their own. If you’d like to contribute further, a focused PR that fixes a user-facing instruction or adds missing project guidance would be a better fit.

PR #1345, opened 2025-07-17
Fix: Preserve hard breaks and new lines in PPTX

### LINKED ISSUE #1331: .pptx conversion doesn't add two spaces at the end of a line to preserve new lines
Converting [my presentation](https://github.com/stelios-c/sig_proc/blob/main/pydata_london2025/Basic%20time-series%20analysis%20in%20Pandas.pptx)

raw looks like this 
```markdown
<!-- Slide number: 1 -->
# Basic timeseries analysis in Pandas
Stelios Christodoulou
PyData London
June 2025

### Notes:

<!-- Slide number: 2 -->
```
which renders to 

Image

To render correctly you need two spaces at the end of the lines

Image

This is possibly related to other pptx issues such as #1087

claytomode:
# Hard Break Preservation and Improved PPTX New Line Handling

This pull request refactors the text content normalization process. It switches to using Python's `str.splitlines()` for more robust cross-platform newline handling and introduces logic to explicitly preserve two-space "hard breaks" in the text.

It then fixes #1331 by adding hard breaks to PPTX conversion


claytomode:

ghost:
Verified that str.splitlines() correctly

MaciejZet:
Hi! I reproduced an adjacent case on current `main`: direct Markdown (`.md` / `text/markdown`) input loses hard breaks because the shared normalizer strips trailing spaces. Your PR appears to cover the exact-two-space case, but its normalizer still strips lines ending in three or more spaces, which CommonMark also uses for hard breaks. I prepared a Markdown-input-scoped regression/fix that preserves two or more spaces without changing HTML output, and I’m holding a separate patch to avoid duplicating your normalizer change. Would you prefer that test/fix be added as a follow-up to this PR, or should we open a separate narrowly scoped PR? Thanks!

PR #1352, opened 2025-07-17
HTML| Update document intelligence file type handling

safen0s:
Added MIME types “text/html” and “application/xhtml+xml” so DocumentIntelligenceConverter can recognize HTML streams

Included “.html” in accepted file extensions to allow HTML files to be processed

Created new tests verifying that HTML files are accepted via extension and MIME type checks


safen0s:

PR #1376, opened 2025-07-23
Extract raster images from pdf

tsdicloud:
- extract ito a subir `img` either at pdf source location (default) or in a given dir
- the unambiguous image name is page_X_image_Y.<imageext>
- unit tests for default and custom directory


tsdicloud:
Sorry, wrong repo.

PR #1406, opened 2025-09-03
Add the HTTP option for running container

justinyoo:
The current MCP server doesn't show the HTTP option to run in a container. This PR is to add an instruction how to run the MCP server in a local container.


chienyuanchang:
Thanks for putting this together. We appreciate the documentation improvement, but this change is more of an additional usage note than a correction to existing instructions. If you'd like to contribute further, please open a new PR or issue with a change that fixes a concrete documentation error or adds a broadly useful improvement.

PR #1442, opened 2025-10-10
Improve output info to help user with Azure Document Intelligence

accelazh:
After I tried it out myself, this PR adds output info to help user when using Azure Document Intelligence.

Previously, even if Azure Document Intelligence client throws an exception, the converter would silently fall back to a more basic converter. No messages are printed to let user know. The output markdown file is still generated with the basic converter. It is very confusing to the user that, the output markdown file seems to be generated by Azure Document Intelligence, but actually NOT.

The typical confusing case is when user forgot to set AZURE_API_KEY. Below is a screenshot. You can see no messages are output.

image

This PR adds helpful output messages to clarify this situaiton for user. Warning messages are printed if Azure Document Intelligence client throws an exception. And if AZURE_API_KEY is forgotten, a hint message is also printed.

Below is the screenshot after this PR:

image


accelazh:

PR #1447, opened 2025-10-16
UI and Batch Processing

TOMMY-CODER28:
Added a Basic UI with Batch processing


afourney:
Thanks for your contribution. At this time, I'd like to keep this repo as mostly a python library. I am not ready to take on any UI etc. 

I encourage the authors to publish their own packages or repos that take MarkItDown as a dependency, and provide a UI if desired.

PR #1454, opened 2025-10-21
Feature/batch converter web UI

TOMMY-CODER28:
Added batch conversion and Frontend


afourney:

Issue #1456, opened 2025-10-24
Integrate full support for SpeechRecognition


irresi:
## Problem
1. audio time greater than [1 minute](https://github.com/Uberi/speech_recognition/pull/171)
 **google speech recognition** can accept file that is less or equal than **1 minutes** and only **50 requests** are allowed per day.
  google speech recognition(not google cloud stt) is almost deprecated at all
2. Exception is handled poorly
  If audio file exceeds 1 minute then it throws error, but user can't know why this error occured.
3. No Test for audio transcription

## Solution
- Integrate full support for [SpeechRecognition](https://github.com/Uberi/speech_recognition?tab=readme-ov-file)
  - Since this project pursues to be lightweight, avoid custom logics and mapping argument to speech_recognition would be great.
  - I will add **kwargs that contains engine, model, api_key, regions, ... which are required from SpeechRecognition library 
  and match it to   SpeechRecongition Library using
  ```
  recognize_method = getattr(recognizer, f"recognize_{engine}")
  ```
  - skip for options with offline dependency to be lightweight and align with image_converter
- Exception handling 
- Add test codes
  - skip for github_actions(CI)
## Related Issue
- #326 tried to add whisper directly, but as #1284 noted, mapping SpeechRecognition's support would be simpler and support more models as well.
- #1275 wants custom support for audio transcription as well.
- #74 If the time of the audio exceeds quota(60 seconds per request, 50 requests per day), 
recognize_google() from SpeechRecognition throws an error, but the user can't know where it came from.

## Suggestion
Asynchronous transcription and chunking would be great feature, 
but I am concerned that it might not align with the project's philosophy. "Lightweight"

PR #1457, opened 2025-10-24
feat: speech recognition support

irresi:
### Description
Support various engines and options using speech recognition

### Changes
`_markitdown.py`
- Add transcription_engine, transcription_kwargs

`_audio_converter.py`
- Gets engine and engine kwargs by finding option that starts with transcription

`_transcribe_audio.py`
- Refer supported engine for speech recognizer and throws exception if not supported
- Enhanced exception handling

`test_transcribe_engines.py`
- Made engine test and format test
- Check if output is string and len(output) > 0
- Skip if GitHub Actions or credentials are not provided.

### Related issue
#1456 
Reference : [example for speech recognition library](https://github.com/Uberi/speech_recognition/blob/master/examples/audio_transcribe.py)


irresi:

PR #1465, opened 2025-10-31
Remove Windows constraint

jjrdk:
The win32 constraint makes means that I cannot use the markdownit package outside of Windows. The onnxruntime package supports all 3 major OS, so the win32 constraint seems unnecessary.

### DIFF
--- a/packages/markitdown/pyproject.toml
+++ b/packages/markitdown/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
   "magika~=0.6.1",
   "charset-normalizer",
   "defusedxml",
-  "onnxruntime<=1.20.1; sys_platform == 'win32'",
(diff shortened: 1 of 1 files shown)


stonebig:
keeping "onnxruntime<=1.20.1" would keep it unusable for Python-3.14, on Windows 11.

afourney:
Thank you for looking into the ONNX Runtime constraint. We addressed the direct dependency pin in [#1551](https://github.com/microsoft/markitdown/pull/1551) by removing MarkItDown's onnxruntime<=1.20.1 entry entirely, rather than retaining the cap and applying it to additional platforms.

I'm closing this as superseded by that change. This only resolves the direct-pin change proposed here; broader Python/platform compatibility and transitive dependency constraints remain separate concerns

PR #1494, opened 2025-12-07
use UTF-8 encoding in unit test `test_output_to_file`

oxygen-dioxide:
On my Chinese windows, this test fails with the following error:
```
        with open(output_file, "r") as f:
>           output_data = f.read()
                          ^^^^^^^^
E           UnicodeDecodeError: 'gbk' codec can't decode byte 0xaa in position 61: illegal multibyte sequence

tests\test_cli_vectors.py:87: UnicodeDecodeError
```
because Markitdown writes to file in UTF-8, but the test tries to read it in local encoding


zashed (approved):

afourney:
Thanks for identifying the locale-dependent decoding failure in test_output_to_file. [#2351](https://github.com/microsoft/markitdown/pull/2351) incorporated the same correction: the test now reads the generated output with an explicit UTF-8 encoding.

The change is on main, so I'm closing this PR as already addressed. Thank you for contributing the fix.

Issue #1506, opened 2025-12-15
Chinese file names in zip file are garbled


oxygen-dioxide:
For example, in this file: 

[chinese-example.zip](https://github.com/user-attachments/files/24155422/chinese-example.zip)

The file names are in Chinese

<img width="1385" height="621" alt="Image" src="https://github.com/user-attachments/assets/5d3c071c-e58b-4b29-a1ea-28bcb415b3e0" />

However when reading the zip file with markitdown, the file names are garbled

```
Content from the zip file `chinese-example.zip`:

## File: ┴?╨█╛?╡τ╢?2017╟∩╞┌╓╨.txt

1、（a）证明静电场的唯一性定理
（b）证明无穷大空间中静电场的唯一性定理，要求ρ ~ 1/r^q，q ≥ q_c，求q_c
（c）利用叠加原理和电像法算电势分布

2、写出接地平面的Green函数，算一个电势分布，作轴外展开

3、从电偶极子的?δ表达式算电势分布、单极矩、偶极矩，然后把电偶极子放在接地平面附近，算能量、力（搞清楚镜像偶极子的方向，能量和力记得除以2）

4、用双线性组合构造Green函数，证明L G = δ，然后在一个接地长方体内，构造z方向的Green函数，然后构造三维的Green函数（把三个方向的一维Green函数相乘，记得乘4 π），然后算一个电势分布

## File: ╨φ╕?╚┘└φ┴?2016╟∩╞┌─?.txt

1.用虚功原理求杆子平衡时夹角
2.用拉格朗日方程推导冲量的效果
3.一个圆盘和一根杆子的小振动
4.用哈-雅方程解谐振子
5.证明自由定点刚体无章动
6.写出重力场中球对称刚体的拉格朗日方程，哈密顿量，泊松括号形式的正则方程
```

mizuresort:
Hi, I tested this with Japanese filenames ( since I'm a Japanese user), and I encountered the same issue.
I'll try working on a fix for this.

PR #1535, opened 2026-01-14
Fix ZIP filename encoding for Chinese and Japanese files

mizuresort:
This PR fixes garbled ZIP filenames for non-UTF8 encodings.

It normalizes ZIP entry names by trying multiple encodings:
- UTF-8
- CP932 (Japanese)
- GBK / BIG5 (Chinese)

It also skips macOS metadata files (e.g., __MACOSX/, ._ files)
to avoid showing non-user content.

This allows international filenames to be displayed correctly.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_zip_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_zip_converter.py
@@ -12,6 +12,24 @@ from .._exceptions import UnsupportedFormatException, FileConversionException
 if TYPE_CHECKING:
     from .._markitdown import MarkItDown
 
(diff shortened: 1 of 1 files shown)


mizuresort:

oxygen-dioxide:
This PR works for me. It fixes my issue #1506

mizuresort:
Thanks for linking your issue and confirming the fix!
I'm glad I could help with your issue :)

McgovernEmily:
I’m still getting familiar with encoding issues, but this makes sense. Is there a reason CP437 is used as the initial encoding before trying others?

PR #1537, opened 2026-01-15
Feat epub inline images

tosmart01:
## Background
Add EPUB image handling so embedded images are preserved during conversion.

## Changes
- Build a manifest path -> media-type map for MIME resolution.
- Inline local EPUB images as data URIs when `keep_data_uris=true`.
- Resolve relative/decoded paths and fall back to `mimetypes` when needed.
- Add an EPUB image conversion test using a sample file.

## Files
- packages/markitdown/src/markitdown/converters/_epub_converter.py
- packages/markitdown/tests/test_epub_images.py
- packages/markitdown/tests/test_files/test_epub_images.epub

## Testing
- packages/markitdown/tests/test_epub_images.py


tosmart01:

PR #1538, opened 2026-01-20
When converting CSV to Markdown, change to use pandas, as with XLSX and XLS.

komoroko:
## Problem

When a CSV file contains line breaks inside a cell, the generated Markdown table becomes corrupted after conversion.  
This type of CSV file is reasonably common in real-world and enterprise datasets, and supporting this case improves the robustness of the conversion process.

### Example CSV

```csv
名前,年齢,住所
佐藤太郎,30,東京
三木英子,25,大阪
髙橋淳,35,名古屋
鈴木一郎,"28

cellB5",福岡
```
In this case, the line break inside the quoted cell causes the Markdown table structure to break.

## Solution

To address this issue, the CSV-to-Markdown conversion was updated to use pandas-based HTML conversion, consistent with the existing handling for xlsx and xls files.

By adopting pandas-based HTML conversion for CSV-to-Markdown processing, multiline cells are handled correctly and the table structure is preserved.

## Result

- CSV files containing line breaks within individual cells are now converted into valid Markdown tables.
- Behavior is consistent with Excel (.xlsx, .xls) conversions
- Improves robustness of Markdown generation for CSV inputs


komoroko:
> @microsoft-github-policy-service agree

komoroko:

afourney:
Thank you for the multiline-CSV example and proposed fix. [#2266](https://github.com/microsoft/markitdown/pull/2266) addresses the reported table breakage by escaping Markdown-sensitive cell content and replacing embedded line breaks with spaces in both headers and data cells.

This preserves table integrity while keeping CSV conversion independent of pandas. It does not preserve visual multiline layout within a cell. The fix is in main and v0.1.8b1, so I'm closing this parser-replacement proposal. We may revisit this again in the future.

PR #1540, opened 2026-01-21
fix: Handle UnicodeDecodeError in PlainTextConverter (#1505)

### LINKED ISSUE #1505: Bug with Spanish symbols: PlainTextConverter threw UnicodeDecodeError with message: 'ascii' codec can't decode byte
Hi, I was trying to use markitdown's convert function with some files and I noticed that it threw an error for certain symbols. I've attached both the error and an example file.

Source used to generate bug:
```

md = MarkItDown()
result = md.convert("./e9aee1c13edc351190d4722c7bd288ca.txt")
print(result.text_content)
```

Error:
```
 - PlainTextConverter threw UnicodeDecodeError with message: 'ascii' codec can't decode byte 0xc3 in position 4101: ordinal not in range(128)
 ```

Example file:
[e9aee1c13edc351190d4722c7bd288ca.txt](https://github.com/user-attachments/files/24154768/e9aee1c13edc351190d4722c7bd288ca.txt)

s3ich4n:
When charset detection samples only the first `4096` bytes and detects `ascii`, but the file contains UTF-8 characters beyond that point, decoding fails with `UnicodeDecodeError`.

Added fallback to `charset_normalizer` when `UnicodeDecodeError` occurs, allowing proper handling of files with non-ASCII characters Spanish, Korean, Japanese, Chinese, etc.)
that appear after the 4096-byte sample.

fixes #1505


s3ich4n:

bpeloquinm-glitch (approved):

mtv-tuanva (approved):

Issue #1470, opened 2025-11-02
Python 3.14 support


carragom:
Hi there, can't install markitdown-mcp with `pipx install markitdown-mcp` in Fedora 43 with Python 3.14. The error is:

```
The conflict is caused by:
    magika 0.6.3 depends on onnxruntime>=1.17.0; python_full_version >= "3.10"
    magika 0.6.2 depends on onnxruntime>=1.17.0; python_full_version >= "3.10"
    magika 0.6.1 depends on onnxruntime>=1.17.0; python_full_version >= "3.10"

Additionally, some packages in these conflicts have no matching distributions available for your environment:
    onnxruntime

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict


PIP STDERR
----------
ERROR: Cannot install markitdown because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```
This error seems to be caused by onnxruntime not supporting 3.14 yet. Here is the microsoft/onnxruntime#26309. Hopefully this can be resolved soon.
Thanks.

josteink:
To me it seems really weird not being able to upgrade a markdown-library to a new Python version because somewhere down in the deep is a dependency which contains a full, binary AI engine.

To me, the use of magika (AI-based file-inference) seems really not necessary for 99% of the use-cases when most streams will be able to provide a file-extension or MIME/content-type.

I would love if markitdown could modularize away this magika dependency as well, as it has for lots of the converts.

I really want to be able to convert all the files, but I really don't want to depend on an AI to do it.

fmmoret:
onnxruntime RC candidate is out; might be close on this one

stonebig:
also, could this limit pre-entively  allow onnxruntime-1.24 for Python-3.14 (on windows) , as nothing else will be possible ? even if it fails

https://github.com/microsoft/markitdown/blob/main/packages/markitdown/pyproject.toml#L33

<img width="926" height="521" alt="Image" src="https://github.com/user-attachments/assets/fb2250af-843e-4e70-ae77-4825d07b5b7f" />

josteink:
> also, could this limit pre-entively allow onnxruntime-1.24 for Python-3.14 (on windows) , as nothing else will be possible ? even if it fails
> 
[...]

PR #1274, opened 2025-05-28
Pin `onnxruntime` on Windows

t-kalinowski:
closes #1266

Issue #1266, opened 2025-05-23
onnxruntime dll import error


t-kalinowski:
Attempting to run `convert` on Windows 11, I see:

```
DLL load failed while importing onnxruntime_pybind11_state: A dynamic link library (DLL) initialization routine failed.
```

It looks like markitdown is not the only package experiencing this issue:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4905#issuecomment-2902130299

I think the onnxruntime version might need to be pinned until this is fixed upstream.
(onnxruntime is a dependency of [magika](https://google.github.io/magika/))

cephyn:
Yup seeing this too!

joshuacwnewton:
Hi there! I see you've tagged our `spinalcordtoolbox` issue. Just to give some additional context, I had also reported this upstream here:

- https://github.com/microsoft/onnxruntime/issues/24843#issuecomment-2904854046

It seems to do with the version of MSVC Runtime that is installed. Try [updating](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#latest-microsoft-visual-c-redistributable-version).

afourney:
Thanks for the report. Let me see what I can pin to address this incompatibility.

t-kalinowski:
With some experimentation, I ended up pinning `"onnxruntime<=1.20.1"` in ragnar:

https://github.com/tidyverse/ragnar/blob/1d5f62fd005bdab467febb86d4c4455d7e22e0c5/R/ragnar-package.R#L24 

Happy to submit a PR.

afourney:
If it's not too much trouble, I'd love a PR can credit you. Otherwise, I'm happy to take this on later today.

If it's impacting Magika, we should perhaps let them know too.

PR #1355, opened 2025-07-17
feat(magika): make magika an optional dependency

SyedaAnshrahGillani:
Description:

  This pull request makes the magika library an optional dependency in the markitdown package.
  Previously, magika (and its dependency onnxruntime) was a mandatory installation, even for users who
  did not utilize its functionality.

  By making magika optional, users can now choose to install it only if they require its features,
  leading to a smaller and more efficient installation for most users.

  Changes:

   * packages/markitdown/pyproject.toml: Moved onnxruntime into a new optional dependency group magika.
   * packages/markitdown/src/markitdown/_markitdown.py:
       * Removed the direct import of magika.
       * Implemented a conditional import for magika within the MarkItDown class's __init__ method.
       * Added a check in _get_stream_info_guesses to ensure magika is available before attempting to use
         it, gracefully skipping its functionality if not installed.

PR #1551, opened 2026-02-06
Remove onnxruntime<=1.20.1 Windows pin

### LINKED ISSUE #1550: Remove onnxruntime<=1.20.1 Windows pin (onnxruntime 1.24.1 is out)
The `onnxruntime<=1.20.1; sys_platform == "win32"` pin was added in #1274 to work around a DLL import error (#1266).

The investigation in [microsoft/onnxruntime#24843](https://github.com/microsoft/onnxruntime/issues/24843) showed that the issue was partly caused by outdated `msvc-runtime` packages, though `onnxruntime>=1.22.0` also showed a genuine regression on `windows-2019` (which is now EOL). Newer Windows versions (`windows-2022`, `windows-2025`) work fine with `onnxruntime>=1.22.0`.

The pin now blocks Python 3.14 support (#1470) because `onnxruntime` 1.24.1 — which adds Python 3.14 wheels — exceeds the `<=1.20.1` upper bound.

Could this pin be removed or bumped? If there's concern about the `windows-2019` regression, the pin could at least be narrowed (e.g. to only affect older Python versions, or documented as a known issue for EOL Windows).

basnijholt:
## Summary

- Remove the `onnxruntime<=1.20.1; sys_platform == 'win32'` pin from `packages/markitdown/pyproject.toml`

## Context

The pin was added in #1274 to work around a DLL import error (#1266). The investigation in [microsoft/onnxruntime#24843](https://github.com/microsoft/onnxruntime/issues/24843) showed:

- The original reporter's issue was caused by an outdated `msvc-runtime` package (updating it fixed it)
- `onnxruntime>=1.22.0` has a genuine regression on the `windows-2019` runner, but `windows-2022` and `windows-2025` work fine
- `windows-2019` is EOL and scheduled for deprecation

The pin now blocks Python 3.14 support (#1470, #1550) because `onnxruntime` 1.24.1 — which adds Python 3.14 wheels — exceeds the `<=1.20.1` upper bound. The `onnxruntime` dependency comes from `magika`, which itself doesn't have this pin.

Fixes #1550

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 packages/markitdown/tests/test_files/** linguist-vendored
 packages/markitdown-sample-plugin/tests/test_files/** linguist-vendored
+
+# Treat PDF files as binary to prevent line ending conversion
+*.pdf binary
--- a/packages/markitdown/src/markitdown/__about__.py
+++ b/packages/markitdown/src/markitdown/__about__.py
@@ -1,4 +1,4 @@
 # SPDX-FileCopyrightText: 2024-present Adam Fourney <adamfo@microsoft.com>
 #
 # SPDX-License-Identifier: MIT
-__version__ = "0.1.5b1"
+__version__ = "0.1.5b2"
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -107,6 +107,13 @@ class MarkItDown:
         requests_session = kwargs.get("requests_session")
         if requests_session is None:
             self._requests_session = requests.Session()
+            # Signal that we prefer markdown over HTML, etc. if the server supports it.
+            # e.g., https://blog.cloudflare.com/markdown-for-agents/
+            self._requests_session.headers.update(
+                {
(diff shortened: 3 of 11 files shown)


josteink:
I'm not sure if this is beating a dead horse, but just iterating that this is an option too:

Making `magika` an optional dependency, which would eliminate all these issues.

There's even a PR for it: https://github.com/microsoft/markitdown/pull/1355

basnijholt:
@josteink, that would also be preferable to me but in the meantime this is a smaller change that unblocks a lot of people.

@afourney, perhaps you could take a look here? Essentially, no-one that is using markitdown can update to Python 3.14...

afourney:
Ooof good point. Let me see what I can do tonight to unblock

afourney:
Merged. I'll push a point release tonight

josteink:
*Knock knock*

Is this release still on?

My security tooling is starting to flag Python 3.13 as a vulnerability, and with this package still tied to obsolete packages which I don't need, I'm not able to make an upgrade.

afourney:
Yes this should be in the 0.1.5 release now on PyPi

PR #1553, opened 2026-02-12
Omml fixes

sdementen:
the omml convestion was failing silently on my document.
I ask copilot to fix all issues it could find and here is the PR


sdementen:

PR #1573, opened 2026-02-28
Guard against UNC path bypass in file URI parsing on Windows

RacerZ-fighting:
## Summary

`file_uri_to_path()` determines whether a `file:` URI refers to a local resource by
inspecting the parsed URI authority (`netloc`). However, a crafted URI such as
`file:////RemoteServer/Share/file` encodes a UNC path entirely within the **path component**,
causing `urlparse()` to return an empty `netloc` — bypassing the locality check — while
`url2pathname()` on Windows converts it into a valid UNC path (`\\RemoteServer\Share\file`).

This allows the application to initiate unintended outbound SMB connections when processing
such inputs, which is particularly concerning in automated or agent-driven workflows
(e.g., MCP-based systems) where an attacker could influence input data by hosting
attacker-controlled files on a remote server.

## Changes

- Add UNC path detection in the URI **path component** on Windows (`os.name == 'nt'`),
  rejecting URIs where the path resolves to a remote UNC share despite an empty `netloc`
- Simplify server name extraction logic (unify separators before split)
- Remove redundant emptiness check

### DIFF
--- a/packages/markitdown/src/markitdown/_uri_utils.py
+++ b/packages/markitdown/src/markitdown/_uri_utils.py
@@ -12,7 +12,26 @@ def file_uri_to_path(file_uri: str) -> Tuple[str | None, str]:
         raise ValueError(f"Not a file URL: {file_uri}")
 
     netloc = parsed.netloc if parsed.netloc else None
-    path = os.path.abspath(url2pathname(parsed.path))
+    path = url2pathname(parsed.path)
+
+    # On Windows, we need to guard against UNC path bypass attempts
+    # (where UNC paths are encoded in the URI path component to bypass netloc checks)
+    if os.name == 'nt':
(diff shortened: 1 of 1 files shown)


RacerZ-fighting:

kdjkdjkdj:
+1 from another Windows user — this is a real issue, not theoretical. `urlparse()` giving empty `netloc` on `file:////server/share/...` while `url2pathname()` still produces a valid UNC path is an unintended SMB egress vector.

Diff looks correct: `os.name == 'nt'` + `netloc is None` is the right scope, `':' not in potential_server` handles the drive-letter edge case, legitimate `file://server/share` URIs pass through. Fail-closed is the right default for this.

cc @afourney — small, isolated, CLA green. Happy to add a unit test if that unblocks review.

PR #1575, opened 2026-03-01
Send CLI validation errors to stderr

bernardassistant557-arch:
## Problem
When CLI validation fails (e.g., missing Document Intelligence endpoint), `markitdown` prints the error message to stdout. This makes it harder to distinguish valid output from errors and breaks common shell patterns that rely on stderr for failures.

## Why this change
Errors should flow to stderr so users can safely pipe/redirect stdout without losing error visibility.

## What changed
- Send `_exit_with_error` output to stderr.
- Added a CLI test that verifies the missing-docintel-endpoint error is emitted on stderr.

## Testing
- `python -m pytest packages/markitdown/tests/test_cli_misc.py`

### DIFF
--- a/packages/markitdown/src/markitdown/__main__.py
+++ b/packages/markitdown/src/markitdown/__main__.py
@@ -215,7 +215,7 @@ def _handle_output(args, result: DocumentConverterResult):
 
 
 def _exit_with_error(message: str):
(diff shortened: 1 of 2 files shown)


bernardassistant557-arch:

PR #1577, opened 2026-03-01
chore/1574-readme-typo-fix

### LINKED ISSUE #1574: Missing quotes in pypi.org
Installation instruction on pypi.org suggest to install package using:
`pip install markitdown[all]`. This doesn't work.

The readme file on github suggests:
`pip install 'markitdown[all]'`. This does work.

subtle but might make a difference.

david-hankinson:
Small typo issue raised
resolves #1574

Hello,

When I was installing this tool, I saw that the command wasn't right. As it stands, with -e and single quotes in the pip install, the command doesn't work. 

I'm on ubuntu 24.04 and the command 'pip install markitdown[all]' is what I needed to run to get it installed properly. 

Also noticed someone else has raised this issue, which i've linked.

This tool looks nice! I'll keep on using it and make PR's if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ To install MarkItDown, use pip: `pip install 'markitdown[all]'`. Alternatively, 
 ```bash
 git clone git@github.com:microsoft/markitdown.git
 cd markitdown
-pip install -e 'packages/markitdown[all]'
+pip install packages/markitdown[all]
 ```
 
 ## Usage


afourney:
Thanks for following up on #1574. The missing quotes in the package README/PyPI installation instructions were corrected in [#2223](https://github.com/microsoft/markitdown/pull/2223).

The top-level source-install example intentionally keeps editable mode and quotes around the extras argument. Since the underlying documentation issue is addressed.

PR #1588, opened 2026-03-06
Allow bytes-like inputs in MarkItDown.convert

### LINKED ISSUE #287: Please allow bytestream as an argument to convert
The convert method takes file, path. Is it possible to pass the bytestream to the same method.

lavish0000:
Closes #287

Summary:
- accept `bytes`, `bytearray`, and `memoryview` inputs in `MarkItDown.convert()`
- route bytes-like inputs through `io.BytesIO` so they follow the existing stream conversion path
- add a regression test covering all supported bytes-like inputs

Testing:
- `uv run --project packages/markitdown --with pytest,mypy python -m pytest packages/markitdown/tests/test_module_misc.py -k 'bytes_like or input_as_strings'`
- `uv run --project packages/markitdown --with ruff python -m ruff check packages/markitdown/src/markitdown/_markitdown.py packages/markitdown/tests/test_module_misc.py`
- `uv run --project packages/markitdown --with ruff python -m ruff format --check packages/markitdown/src/markitdown/_markitdown.py packages/markitdown/tests/test_module_misc.py`

Note:
- running targeted mypy against this package in this checkout still surfaces pre-existing unrelated issues in untouched files, so I limited type-check validation to the touched path plus the regression test.

### DIFF
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -251,14 +251,17 @@ class MarkItDown:
 
     def convert(
         self,
-        source: Union[str, requests.Response, Path, BinaryIO],
+        source: Union[
+            str, requests.Response, Path, BinaryIO, bytes, bytearray, memoryview
+        ],
         *,
         stream_info: Optional[StreamInfo] = None,
         **kwargs: Any,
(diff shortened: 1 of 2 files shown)


lavish0000:

lavish0000:

Issue #1534, opened 2026-01-14
A bug when llm_description got None response from LLM in PptxConverter


wangxb12:
Got below error when converting a .pptx file. 

Error when converting the document: File conversion failed after 1 attempts:
 - PptxConverter threw TypeError with message: sequence item 0: expected str instance, NoneType found

After investigation the llm_description got None response from LLM in line 120 of _pptx_converter.py and it will cause this issue in line 139.
<img width="1138" height="826" alt="Image" src="https://github.com/user-attachments/assets/58929bea-931a-4696-a078-80ce0ea80113" />

hshgogogo:
I plan to work on this bug and submit a PR with a fix and regression test.

PR #1623, opened 2026-03-18
Fix PPTX conversion when LLM caption is None

ReinerBRO:
Summary
- Handle None returned by llm_caption in PptxConverter.
- Normalize embedded PPTX descr alt text to a string before joining.
- Add regression test for PPTX conversion with llm_caption mocked to return None.

Why
Issue #1534 reports a TypeError in PPTX conversion when an LLM returns None for image captioning.

Testing
- .venv/bin/pytest packages/markitdown/tests/test_module_misc.py::test_markitdown_pptx_none_llm_description packages/markitdown/tests/test_module_misc.py::test_markitdown_llm_parameters -q


afourney:

PR #1626, opened 2026-03-20
Update SECURITY.md

sahabatmotortelukmakmur-cell:


chienyuanchang:
Thanks for putting this together. This change looks like formatting/polish only in `SECURITY.md`, so we usually keep those kinds of edits out of the main review queue unless they’re part of a larger documentation batch. If you’d like to contribute, please open a new PR with a substantive docs or code change and we’ll be glad to take a look.

PR #1663, opened 2026-04-04
feat: add EmlConverter with attachment support

### LINKED ISSUE #1662: EmlConverter: add support for converting email attachments
## Problem

PR #1633 adds `EmlConverter` which correctly extracts email headers and the message body. However, email attachments are not processed. A `.eml` file that contains an attached PDF, Word document, or image is converted to just the email body text, with no indication that attachments exist or what they contain.

For use cases like:
- Converting an inbox export to a searchable knowledge base
- Extracting information from automated report emails with attached spreadsheets
- Processing compliance email archives that include attached contracts

...the body text alone is not enough. The attachments often contain the most important content.

## Proposed solution

Extend `EmlConverter._get_body()` (or add a new `_get_attachments()` method) to:

1. Detect MIME parts that are attachments (i.e., `Content-Disposition: attachment`)
2. For each attachment, pass the raw bytes back through the main `MarkItDown` converter using the attachment filename to determine the appropriate converter
3. Append each converted attachment to the markdown output under a `## Attachment: filename.ext` heading

Attachments that cannot be converted (unknown format, conversion error) should be listed with a note rather than silently skipped.

## Example output

```markdown
# Email Message

**From:** sender@example.com
**Subject:** Q1 Report

## Content

Please find the Q1 report attached.

## Attachment: Q1_Report.xlsx

| Quarter | Revenue | Costs |
|---------|---------|-------|
| Q1 2026 | 1.2M    | 0.8M  |
```

## Additional context

The recursive converter call pattern (passing attachments back through `MarkItDown`) is already used in `ZipConverter` for processing zip contents. The same approach would work here with minimal new code.

VANDRANKI:
## Summary

Adds `EmlConverter` for converting RFC 822 `.eml` files to markdown.

This extends the gap identified in issue #1662. The converter handles:
- Email headers (From, To, Cc, Subject, Date)
- Body content (prefers `text/plain`, falls back to `text/html` with tag stripping)
- Attachments: each attachment is passed through the MarkItDown converter pipeline and rendered under its own heading in the output

## What this PR covers

- `converters/_eml_converter.py` - new `EmlConverter` class
- `converters/__init__.py` - registers the new converter in exports
- `_markitdown.py` - registers `EmlConverter(markitdown=self)` in `enable_builtins()`, matching the `ZipConverter` pattern for recursive conversion
- `tests/test_eml_converter.py` - 11 tests covering headers, body, attachment conversion, unsupported format fallback, and body/attachment isolation

## Attachment design

Attachment conversion follows the same pattern as `ZipConverter`: a `MarkItDown` instance is injected via the constructor and used to call `convert_stream()` on each attachment. This means any format markitdown supports (XLSX, PDF, DOCX, etc.) is automatically available for email attachments with no additional code.

If no `MarkItDown` instance is available (e.g., the converter is instantiated directly), attachments are listed by filename with a note rather than silently skipped.

## Example output

```markdown
# Email Message

**From:** sender@example.com
**To:** recipient@example.com
**Subject:** Q1 Report
**Date:** Sat, 04 Apr 2026 12:00:00 +0000

## Content

Please find the Q1 report attached.

## Attachments

### Q1_Report.xlsx

## Sheet1
| Quarter | Revenue |
|---------|---------|
| Q1 2026 | 1.2M    |
```

## Testing

All 11 tests pass. ruff check and ruff format pass with zero issues.

Closes #1662

### DIFF
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -35,6 +35,7 @@ from .converters import (
     ImageConverter,
     AudioConverter,
     OutlookMsgConverter,
+    EmlConverter,
     ZipConverter,
     EpubConverter,
     DocumentIntelligenceConverter,
@@ -200,6 +201,7 @@ class MarkItDown:
             self.register_converter(IpynbConverter())
             self.register_converter(PdfConverter())
             self.register_converter(OutlookMsgConverter())
+            self.register_converter(EmlConverter(markitdown=self))
             self.register_converter(EpubConverter())
             self.register_converter(CsvConverter())
 
--- a/packages/markitdown/src/markitdown/converters/__init__.py
+++ b/packages/markitdown/src/markitdown/converters/__init__.py
@@ -16,6 +16,7 @@ from ._pptx_converter import PptxConverter
 from ._image_converter import ImageConverter
 from ._audio_converter import AudioConverter
 from ._outlook_msg_converter import OutlookMsgConverter
+from ._eml_converter import EmlConverter
 from ._zip_converter import ZipConverter
 from ._doc_intel_converter import (
     DocumentIntelligenceConverter,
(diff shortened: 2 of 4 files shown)


VANDRANKI:

PR #1666, opened 2026-04-05
feat: add configurable safety limits to ZipConverter

### LINKED ISSUE #1661: Make ZipConverter safety limits configurable via constructor parameters
## Problem

PR #1628 added zip bomb protection to `ZipConverter` with three hardcoded module-level constants:

- `MAX_DECOMPRESSED_FILE_SIZE = 100 MB` per file
- `MAX_DECOMPRESSION_RATIO = 100:1`
- `MAX_TOTAL_DECOMPRESSED_SIZE = 500 MB` total

These defaults are reasonable for general use, but they are not configurable. This creates a real problem for legitimate use cases:

- **Scientific datasets**: ZIP archives commonly contain files well over 100 MB (genomics data, satellite imagery, simulation outputs).
- **Legal and financial document archives**: SEC EDGAR bulk data packages and court document bundles regularly exceed 500 MB total.
- **Internal tooling**: An organization running markitdown on known-safe internal archives has no way to raise the limits without monkey-patching the module.

Users who hit these limits get silent skipping with only a logger warning, and no way to know programmatically that their content was truncated.

## Proposed solution

Move the limits to constructor parameters on `ZipConverter` with the current values as defaults:

```python
class ZipConverter(DocumentConverter):
    def __init__(
        self,
        max_file_size: int = 100 * 1024 * 1024,
        max_ratio: int = 100,
        max_total_size: int = 500 * 1024 * 1024,
    ):
        self.max_file_size = max_file_size
        self.max_ratio = max_ratio
        self.max_total_size = max_total_size
```

This is a non-breaking change: the defaults stay the same, and users who need larger limits can pass them explicitly. Users who want to disable the limits entirely can pass `float("inf")`.

## Additional context

This was flagged during review of PR #1628 as a blocker before merge. Opening as a tracked issue so it does not get lost if the PR merges first.

VANDRANKI:
## Problem

The current `ZipConverter` has no limits on how many files it processes or how much data it extracts. A maliciously crafted zip file (zip bomb) can compress gigabytes of data into a small archive, causing OOM or denial of service when converted. There is also no protection against zip slip path traversal attacks.

## Changes

**`converters/_zip_converter.py`:**

- Add three constructor parameters with safe defaults:
  - `max_file_count` (default 100): stops processing when limit is reached, appends a notice
  - `max_file_size` (default 50 MB): skips individual files that exceed the limit with a per-file notice
  - `max_total_size` (default 200 MB): stops processing when cumulative uncompressed bytes would exceed the budget
- Guard against zip slip: skip entries whose name starts with `/` or contains `..` path components (cross-platform - handles both POSIX and Windows absolute paths)
- Iterate over `ZipInfo` objects instead of `namelist()` to read `file_size` before extracting
- Skip directory entries (names ending with `/`)

**`tests/test_zip_converter.py`** (new file): 10 tests covering all limit types, zip slip protection, and `accepts()`.

## Usage

```python
# Use defaults (100 files, 50 MB/file, 200 MB total)
md = MarkItDown()

# Tighter limits for untrusted input
from markitdown.converters import ZipConverter
converter = ZipConverter(
    markitdown=md_instance,
    max_file_count=20,
    max_file_size=5 * 1024 * 1024,   # 5 MB
    max_total_size=50 * 1024 * 1024, # 50 MB
)
```

## Tests

```
python -m pytest tests/test_zip_converter.py -v
# 10 passed
```

```
ruff check src/markitdown/converters/_zip_converter.py tests/test_zip_converter.py  # All checks passed
ruff format --check ...  # Already formatted
```

Closes #1661

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_zip_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_zip_converter.py
@@ -1,12 +1,11 @@
-import zipfile
 import io
 import os
-
-from typing import BinaryIO, Any, TYPE_CHECKING
+import zipfile
+from typing import Any, BinaryIO, TYPE_CHECKING
 
 from .._base_converter import DocumentConverter, DocumentConverterResult
+from .._exceptions import FileConversionException, UnsupportedFormatException
 from .._stream_info import StreamInfo
-from .._exceptions import UnsupportedFormatException, FileConversionException
 
 # Break otherwise circular import for type hinting
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


VANDRANKI:

PR #1668, opened 2026-04-05
fix: handle missing m:t text in OMML math runs (fixes #1512)

### LINKED ISSUE #1512: Doesn't convert the docx to md properly with the mathematical equations inside
I have fed it a docx file with the mathematical equations inside it and when I converted it to the md format, equations was completely missing from md file, just the text was visible. pypandoc library however did the conversion perfectly, with all the equations visible. So, is it a bug in the markitdown library or is this feature missing?

If missing, then I would request the moderators to add this!

Original docx -:
Image

With Markitdown -:
Image

With pypandoc -:
Image

Thanks for your consideration.

Regards,
Vineet Mangal

octo-patch:
Fixes #1512

## Problem

When converting a DOCX file with mathematical equations, if any `m:r` element inside an `oMath` block has no `m:t` child (for example, a formatting-only run containing only `m:rPr`), `ElementTree.findtext()` returns `None`. The `do_r` method in `omml.py` then does:

```python
for s in elm.findtext("./{ns}t".format(...)):  # iterates over None → TypeError
```

This `TypeError` propagates up through `_pre_process_math` → `pre_process_docx`, where a file-level `try/except` catches it and falls back to the raw, unmodified DOCX XML. Since mammoth cannot render OMML markup, **all equations in the document silently disappear** from the converted markdown output.

## Solution

1. **`omml.py` — `do_r`**: use `(elm.findtext(...) or "")` so an absent `m:t` element yields an empty string instead of crashing.

2. **`pre_process.py` — `_pre_process_math`**: wrap each individual equation replacement in its own `try/except` with a warning log. This means a single un-convertible equation no longer causes every other equation in the document to be lost.

## Testing

- Existing `test_docx_equations` continues to pass.
- New regression test `test_docx_equations_omit_empty_run` directly exercises an `m:r` with only `m:rPr` (no `m:t`) and asserts that the surrounding equation is still converted to `$x$`.

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
@@ -373,7 +373,7 @@ class oMath2Latex(Tag2Method):
         @todo \text (latex pure text support)
         """
         _str = []
-        for s in elm.findtext("./{0}t".format(OMML_NS)):
+        for s in (elm.findtext("./{0}t".format(OMML_NS)) or ""):
             # s = s if isinstance(s,unicode) else unicode(s,'utf-8')
             _str.append(self._t_dict.get(s, s))
         return escape_latex(BLANK.join(_str))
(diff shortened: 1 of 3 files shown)


afourney:
Missing <m:t> elements in OMML runs are now handled by #2189, with focused regression coverage. I’m closing this as superseded; please reopen if an empty math run still causes conversion to fail.

PR #1671, opened 2026-04-06
fix: sanitize subprocess call in _exiftool.py

orbisai0security:
## Summary
Fix critical severity security issue in `packages/markitdown/src/markitdown/converters/_exiftool.py`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | CRITICAL |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `packages/markitdown/src/markitdown/converters/_exiftool.py:41` |
| **CWE** | CWE-22 |

**Description**: The ExifTool converter passes user-controlled file paths to subprocess.run() without proper validation. While the code uses a list format (not shell=True), the local_path variable comes from user input and could contain malicious filenames. If the file path contains special characters or is manipulated through path traversal, it could lead to unintended file access or command execution through ExifTool's own command parsing.

## Changes
- `packages/markitdown/src/markitdown/converters/_exiftool.py`

## Verification
- [ ] Build not verified
- [ ] Scanner re-scan not performed
- [ ] LLM code review not performed

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_exiftool.py
+++ b/packages/markitdown/src/markitdown/converters/_exiftool.py
@@ -1,5 +1,7 @@
 import json
 import locale
+import os
+import shutil
 import subprocess
 from typing import Any, BinaryIO, Union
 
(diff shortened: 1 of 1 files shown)


VANDRANKI (commented):
Solid security improvement. Three things work together here:

1. `shutil.which()` + `os.realpath()` resolves the path to an absolute canonical path before it reaches subprocess. This prevents symlink-based path manipulation and catches cases where the path looks valid but does not resolve to an executable.

2. `--` separator before `-` in the exiftool invocation. Without this, a filename starting with `-` could be interpreted as a flag. Passing content via stdin with `-` was already safe, but the `--` makes the intent explicit.

3. `timeout=30` on the version check subprocess call, with `subprocess.TimeoutExpired` added to the except clause. The main subprocess call - is there a timeout on that too? Large files passed via stdin could hang indefinitely if exiftool stalls. Worth adding the same timeout there.

Otherwise LGTM.

PR #1673, opened 2026-04-07
fix: preserve HTML tables in Outlook .msg conversion

### LINKED ISSUE #1567: [outlook]: HTML Tables in outlook files
I have an outlook `.msg` file which contains HTML tables. 
How can I extract them to markdown? 

the program does not preserve the html table so it just gets dumped as plain text without formatting

octo-patch:
Fixes #1567

## Problem

The `OutlookMsgConverter` reads only the plain text body, discarding the HTML body. When a .msg file contains HTML tables, the plain text fallback strips all HTML formatting.

## Solution

Prefer the HTML body (PR_BODY_HTML) when it exists:

1. Try Unicode HTML stream `__substg1.0_1013001F` first
2. Fall back to binary HTML stream `__substg1.0_10130102`
3. Convert HTML to markdown via BeautifulSoup + _CustomMarkdownify (same as HtmlConverter)
4. Fall back to plain text body if no HTML body is present

No new dependencies introduced.

## Testing

Tested with a .msg file containing HTML tables. Before: unformatted plain text. After: proper markdown tables.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_outlook_msg_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_outlook_msg_converter.py
@@ -1,8 +1,10 @@
 import sys
 from typing import Any, Union, BinaryIO
+from bs4 import BeautifulSoup
 from .._stream_info import StreamInfo
 from .._base_converter import DocumentConverter, DocumentConverterResult
 from .._exceptions import MissingDependencyException, MISSING_DEPENDENCY_MESSAGE
+from ._markdownify import _CustomMarkdownify
 
 # Try loading optional (but in this case, required) dependencies
 # Save reporting of any exceptions for later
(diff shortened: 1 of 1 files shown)


VANDRANKI (commented):
The priority order (HTML body first, plain text fallback) is correct - MSG files that have an HTML body should produce better markdown than stripping it to plain text.

A few points:

**`script`/`style` removal** - good. Without this, CSS and JS in the HTML body would end up in the markdown output.

**`assert olefile is not None`** in `_get_binary_stream_html` - the existing `_get_stream_data` method uses `assert olefile is not None` too, so this is consistent with the existing pattern. But both methods are only called after the class-level import check, so the assertion should never fire in practice - it is really just a type-narrowing hint. A comment explaining this would help future readers.

**`BeautifulSoup` dependency** - is this already available in the optional deps for `OutlookMsgConverter`? If not, it needs to be added to the extras in `pyproject.toml` alongside `olefile`.

**Tests** - is there a test .msg file with an HTML body that exercises the new path? If not, a test covering table preservation would strengthen this.

The logic itself looks correct.

PR #1686, opened 2026-04-10
Add disabled_converters support for built-ins

### LINKED ISSUE #1665: No way to disable specific built-in converters without subclassing
## Problem

When you call `MarkItDown()`, all built-in converters are registered. There is no way to exclude specific converters without subclassing `MarkItDown` and overriding `enable_builtins`.

This creates friction in two common cases:

1. **Security hardening.** A deployment that serves untrusted files might want to disable `ZipConverter` (to prevent zip bombs even with the existing limits) or `AudioConverter` (to avoid network calls to Whisper APIs on untrusted audio files). Right now this requires a custom subclass.

2. **Dependency management.** Some converters require optional extras (`mammoth` for DOCX, `pptx` for PowerPoint, `speech_recognition` for audio). If a user installs `markitdown` without those extras, the converter registration silently does nothing. A user who wants to explicitly opt in to only a specific set of converters has no clean way to do that.

## Proposed API

```python
# Option A: exclusion list
md = MarkItDown(disabled_converters=["ZipConverter", "AudioConverter"])

# Option B: explicit inclusion (no built-ins except what you list)
md = MarkItDown(enable_builtins=False)
md.register_converter(PdfConverter())
md.register_converter(HtmlConverter())

# Option B already partially works - enable_builtins=False would just skip the auto-registration
```

Option A is the lower-friction change. The `enable_builtins` method could accept an `exclude` parameter:

```python
def enable_builtins(self, md: "MarkItDown", exclude: list[str] | None = None) -> None:
    ...
```

## Notes

- This is purely an API surface change with no impact on converter logic
- The `exclude` list could match on class name strings to avoid importing converter classes just to exclude them

txhno:
Closes #1665.

Adds a `disabled_converters` option to built-in registration and covers it with focused tests.

Validation:
- `pytest tests/test_module_misc.py -k "disabled_converters or enable_builtins_with_disabled_converters"`
- direct runtime check in the repo venv

### DIFF
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -144,6 +144,8 @@ class MarkItDown:
         This method should only be called once, if built-ins were initially disabled.
         """
         if not self._builtins_enabled:
+            disabled_converters = set(kwargs.get("disabled_converters") or [])
+
             # TODO: Move these into converter constructors
             self._llm_client = kwargs.get("llm_client")
             self._llm_model = kwargs.get("llm_model")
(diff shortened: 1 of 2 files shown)


VANDRANKI (approved):
Thanks for implementing this. The approach is clean and addresses what the issue asked for.

**`maybe_register` helper** - good pattern. Keeps the registration block readable without repeating the guard on every line.

**Class name strings** - `type(converter).__name__ not in disabled_converters` is the right check. Callers do not need to import converter classes, which was the core ask.

**Priority refactor** - I checked: `register_converter` defaults to `PRIORITY_SPECIFIC_FILE_FORMAT`, so assigning it explicitly to `RssConverter`, `WikipediaConverter`, etc. only makes the implicit default visible. No behavioral change.

**Tests** - both paths covered: `MarkItDown(disabled_converters=...)` constructor and `enable_builtins(disabled_converters=...)` directly. The `registered_converter_names` helper is reusable.

LGTM.

txhno:

txhno:
CLA is complete from my end now.

PR #1692, opened 2026-04-10
fix: suppress pydub RuntimeWarning when ffmpeg is not installed

### LINKED ISSUE #1685: [Bug]: RuntimeWarning from pydub: "Couldn't find ffmpeg or avconv" on Linux
## Describe the bug
When running **MarkItDown** on a Linux system, a `RuntimeWarning` is triggered by the `pydub` dependency. The warning indicates that `ffmpeg` or `avconv` is missing from the system path, which may cause audio-to-markdown conversion features to fail.

## To Reproduce
1. Install `markitdown` in a Python 3.12 virtual environment on a Linux distribution.
2. Run a conversion command or import the library in a script.
3. Observe the following warning in the console:
   ```text
   /markitdown/.venv/lib/python3.12/site-packages/pydub/utils.py:170: RuntimeWarning: Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work
     warn("Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work", RuntimeWarning)

### Expected behavior
The library should either:
- Document `ffmpeg` as a required system-level dependency for Linux users.
- Handle the missing dependency gracefully if audio processing is not being actively used in the current task.

### System Configuration
- **OS:** Linux
- **Python Version:** 3.12
- **Environment:** Virtual environment (`.venv`)

xor-xe:
## Description

When `pydub` is imported on a system without `ffmpeg` or `avconv` in `PATH`, it emits a `RuntimeWarning` at import time — even when no audio conversion is being performed. This affects any user who has the `audio-transcription` extras installed.

The existing `warnings.catch_warnings()` block in `_transcribe_audio.py` already suppresses `DeprecationWarning` and `SyntaxWarning` on import, but `RuntimeWarning` was not covered.

## Fix

Extend the existing filter block with a targeted filter for pydub's ffmpeg warning:

\`\`\`python
warnings.filterwarnings("ignore", message=".*ffmpeg.*", category=RuntimeWarning)
\`\`\`

Using `message=".*ffmpeg.*"` keeps the scope narrow — only the specific "Couldn't find ffmpeg or avconv" warning is silenced, not all `RuntimeWarning`s. If ffmpeg is actually missing when audio conversion is attempted, pydub will raise a proper error at that point anyway.

## Related

Fixes #1685

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_transcribe_audio.py
+++ b/packages/markitdown/src/markitdown/converters/_transcribe_audio.py
@@ -13,6 +13,7 @@ try:
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", category=DeprecationWarning)
         warnings.filterwarnings("ignore", category=SyntaxWarning)
+        warnings.filterwarnings("ignore", message=".*ffmpeg.*", category=RuntimeWarning)
         import speech_recognition as sr
         import pydub
 except ImportError:


VANDRANKI (commented):
Correct placement - the filter is inside the same `warnings.catch_warnings()` block that already suppresses `DeprecationWarning` and `SyntaxWarning` on import. Using `message=".*ffmpeg.*"` as the match pattern is specific enough to avoid accidentally suppressing unrelated RuntimeWarnings from other packages.

LGTM.

Mouli04 (approved):

Mouli04 (approved):

afourney:
Thanks for the fix. The same missing-FFmpeg warning is covered by #1985; please reopen if #1985 does not address the warning you observed.

PR #1696, opened 2026-04-10
fix: preserve underline formatting in DOCX to Markdown conversion

### LINKED ISSUE #35: Underline not preserved
I get that there is no syntax for underline in Markdown, but maybe an optional setting for `<u> .... </u> ` ?

nitinbhatia-dev:
## Summary

Closes #35

Underlined text in `.docx` files was silently dropped during conversion. Two things were missing:

- **Mammoth** was not configured to emit `<u>` HTML tags for underlined runs (it requires an explicit `style_map` rule)
- **`_CustomMarkdownify`** had no `convert_u` handler, so even if `<u>` tags arrived they would have been stripped

## Changes

- `_docx_converter.py`: Set `u => u` as the default Mammoth `style_map` so underlined runs are emitted as `<u>` elements. Any user-supplied `style_map` is appended and takes precedence.
- `_markdownify.py`: Add `convert_u` to `_CustomMarkdownify` that renders `<u>text</u>` inline HTML — the only way to express underline in CommonMark.

## Test plan

- [ ] Convert a `.docx` file containing underlined text and verify the output contains `<u>...</u>` wrapping the underlined spans
- [ ] Verify bold and italic text still convert correctly (`**bold**`, `*italic*`)
- [ ] Verify passing a custom `style_map` kwarg still works and takes precedence over the default
- [ ] Run the existing test suite: `hatch test`

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_docx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_docx_converter.py
@@ -75,7 +75,16 @@ class DocxConverter(HtmlConverter):
                 _dependency_exc_info[2]
             )
 
-        style_map = kwargs.get("style_map", None)
+        # Default style map preserves underline formatting via HTML <u> tags,
+        # since Markdown has no native underline syntax. User-supplied style_map
+        # is appended so its rules take precedence over the defaults.
+        default_style_map = "u => u"
+        user_style_map = kwargs.get("style_map", None)
+        style_map = (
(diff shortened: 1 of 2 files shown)


nitinbhatia-dev:

afourney:
Thanks for working on underline preservation. The same behavior is covered by #2017. Please re-open, or comment if issues remain.

PR #1709, opened 2026-04-11
Update reference model on README.md

p4cs-974:


chienyuanchang:
Thanks for the contribution. This update changes example model names in the README, but it doesn't appear to fix an incorrect instruction or add user-facing functionality for MarkItDown itself.

We generally batch small documentation polish like this with related changes, so we won't merge this one as-is. If you'd like to contribute again, a focused PR with a concrete documentation fix or a MarkItDown feature update would be a better fit.

PR #1711, opened 2026-04-11
pyproject.toml

saiganesh47:


chienyuanchang:
Thanks for putting this together. This change is just formatting and metadata cleanup in the sample plugin's `pyproject.toml`, so we tend to reserve PRs like this for a batch of small maintenance updates. If you'd like to contribute, a change with clearer user impact or a bug fix would be a better fit.

PR #1717, opened 2026-04-12
fix(markitdown-ocr): make PyMuPDF an optional dependency to fix AGPL licensing concern

### LINKED ISSUE #1675: Use of Pymupdf
Hi, I was reviewing the [Markitdown-OCR](https://github.com/microsoft/markitdown/blob/main/packages/markitdown-ocr/README.md) plug in readme, and saw that it used Pymupdf to address some edge cases, and this dependency was confirmed when reviewing the [pyproject.toml](https://github.com/microsoft/markitdown/blob/main/packages/markitdown-ocr/pyproject.toml) for that plugin. 

Pymupdf uses an [AGPL3 license](https://github.com/pymupdf/PyMuPDF?tab=AGPL-3.0-1-ov-file#readme), and my understanding is that use of that (directly or transitively) would impact the licensing of any apps using it. If this understanding is correct, would it be possible to either:

* Provide a version of the plugin that excludes the Pymupdf dependency (knowing that there will be some edge cases it can no longer handle)

or

* Update documentation to make the implications of use of this plugin more clear. Currently, the README doesn't disclose this, and it's particularly confusing because MarkItDown's own license (and that of the plugin) are MIT.

Thank you for your time,
Noah

octo-patch:
Fixes #1675

## Problem

`markitdown-ocr` declared `PyMuPDF>=1.24.0` as a **required** dependency, but PyMuPDF is licensed under **AGPL-3.0**. Since the plugin itself is MIT-licensed, this mismatch was not disclosed anywhere. Any user who installed `markitdown-ocr` silently acquired an AGPL transitive dependency, which can affect the licensing requirements of applications that distribute the software.

PyMuPDF is only used in one place: a fallback path in `_pdf_converter_with_ocr.py` that handles malformed PDFs that `pdfplumber` cannot open (e.g. truncated EOF). This is an edge case, not the primary conversion path.

## Solution

- Move `PyMuPDF` from `dependencies` to an optional `[pymupdf]` extra in `pyproject.toml`
- Add an `[all]` convenience extra that bundles both `[llm]` and `[pymupdf]`
- Add a clear license notice in `README.md` explaining the AGPL implications and showing how to install with or without PyMuPDF

The existing `import fitz` call is already inside a `try/except`, so the fallback path degrades gracefully when PyMuPDF is not installed — no code changes needed.

## Testing

- `pip install markitdown-ocr` installs without PyMuPDF; standard PDF/DOCX/PPTX/XLSX conversion works normally
- `pip install 'markitdown-ocr[pymupdf]'` enables the malformed-PDF fallback
- `pip install 'markitdown-ocr[all]'` pulls in both openai and PyMuPDF extras

### DIFF
--- a/packages/markitdown-ocr/README.md
+++ b/packages/markitdown-ocr/README.md
@@ -24,6 +24,28 @@ The plugin uses whatever OpenAI-compatible client you already have. Install one 
 pip install openai
 ```
 
+### Optional: Malformed PDF fallback (PyMuPDF)
+
+`markitdown-ocr` includes a fallback for PDFs that `pdfplumber` cannot open (e.g. truncated or malformed files). This fallback uses [PyMuPDF](https://pymupdf.readthedocs.io/) (`fitz`), which is licensed under **AGPL-3.0**. PyMuPDF is **not installed by default** to avoid imposing AGPL requirements on users who do not need it.
+
+To enable the malformed-PDF fallback:
+
+```bash
+pip install 'markitdown-ocr[pymupdf]'
+```
+
+Or install everything at once:
+
+```bash
+pip install 'markitdown-ocr[all]'
+```
+
+> **License notice:** Including `PyMuPDF` (via the `[pymupdf]` or `[all]` extras) adds an AGPL-3.0
+> dependency to your project. If you distribute software that links PyMuPDF, you must comply with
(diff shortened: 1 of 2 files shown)


manuel1618:
+1 and thanks!

PR #1719, opened 2026-04-12
Add support for .vtt / WebVTT subtitle files

### LINKED ISSUE #1682: Support .vtt / WebVTT subtitle files

leul8:
## Summary

Closes #1682

Adds a dedicated `VttConverter` for WebVTT subtitle files (`.vtt` / `text/vtt`), producing clean readable Markdown instead of raw subtitle markup.

## What it does

- Recognizes `.vtt` files and `text/vtt` MIME type explicitly
- Strips the `WEBVTT` header, cue identifiers, and timestamp lines
- Formats `<v Speaker>` voice spans as `Speaker: text`
- Strips other inline tags (`<b>`, `<i>`, `<c.class>`, etc.)
- Skips `NOTE`, `STYLE`, and `REGION` blocks
- Collapses consecutive duplicate lines (common in rolling-caption files)

## Optional timestamp mode

Pass `vtt_include_timestamps=True` to preserve start times:

```python
md.convert("transcript.vtt", vtt_include_timestamps=True)

### DIFF
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -39,6 +39,7 @@ from .converters import (
     EpubConverter,
     DocumentIntelligenceConverter,
     CsvConverter,
+    VttConverter,
 )
 
 from ._base_converter import DocumentConverter, DocumentConverterResult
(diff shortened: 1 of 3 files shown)


leul8:

PR #1723, opened 2026-04-12
Fix multiple converter bugs: RSS crash, PPTX text loss, CSV pipes, Wikipedia regex, ExifTool encoding

hobostay:
## Summary

This PR fixes 5 bugs found across multiple converters:

1. **RSS converter crash** (`_rss_converter.py`): `NameError` when an RSS feed has no `<title>` element in `<channel>`. `md_text` was never initialized before the conditional assignment, causing a crash when `channel_title` is `None`. Fixed by initializing `md_text = ""` and using `+=` consistently.

2. **PPTX text frame loss** (`_pptx_converter.py`): `elif shape.has_text_frame` after `if shape.has_chart` caused text content to be silently dropped for shapes that have both a chart and a text frame. Changed `elif` to `if` so both chart and text content are extracted.

3. **CSV malformed tables** (`_csv_converter.py`): Pipe characters (`|`) in CSV cell values were not escaped, producing broken markdown tables with extra columns. Added `escape_pipe()` helper to escape `|` as `\|`.

4. **Wikipedia language subdomain regex** (`_wikipedia_converter.py`): The regex `[a-zA-Z]{2,3}` rejected valid Wikipedia subdomains with hyphenated or multi-character language codes (e.g., `be-tarask`, `zh-classical`, `roa-rup`, `map-bms`). Changed to `[a-zA-Z0-9-]+` to match all valid Wikipedia language subdomains.

5. **ExifTool encoding** (`_exiftool.py`): Used `locale.getpreferredencoding(False)` to decode ExifTool JSON output, which can fail on non-UTF-8 systems (e.g., Windows with `cp1252`). Since ExifTool always outputs UTF-8 JSON, hardcoded `"utf-8"` and removed unused `locale` import.

## Test plan

- [x] All 159 existing test vectors pass
- [x] 7 new tests added in `test_bug_fixes.py` covering:
  - RSS feed with no channel title
  - RSS feed with no title or description
  - CSV with pipe characters in cells
  - Wikipedia standard subdomains (en)
  - Wikipedia hyphenated subdomains (be-tarask)
  - Wikipedia multi-character subdomains (zh-classical)
  - Wikipedia numeric subdomains (roa-rup)
- [x] Total: 203 tests passed, 0 failures

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -57,8 +57,13 @@ class CsvConverter(DocumentConverter):
         # Create markdown table
         markdown_table = []
 
+        def escape_pipe(cell: str) -> str:
+            return cell.replace("|", "\\|")
+
         # Add header row
-        markdown_table.append("| " + " | ".join(rows[0]) + " |")
+        markdown_table.append(
+            "| " + " | ".join(escape_pipe(cell) for cell in rows[0]) + " |"
+        )
 
         # Add separator row
         markdown_table.append("| " + " | ".join(["---"] * len(rows[0])) + " |")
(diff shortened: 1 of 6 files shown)


afourney:
Thanks for the contribution.

These look like legitimate fixed, but should be separated into separate focused PRs. Please also check the current main brain to verify the issues are still present. A number of them may have already been corrected in other PRs.

Issue #1738, opened 2026-04-13
fix packages/markitdown/src/markitdown/_uri_utils.py


llA1ll:
- 
- 
- from urllib.parse import urlparse, unquote_to_bytes
+ from urllib.parse import urlparse, unquote, unquote_to_bytes

-    path = os.path.abspath(url2pathname(parsed.path))
+    decoded_path = unquote(parsed.path)
+    path = os.path.abspath(url2pathname(decoded_path))

Before the change, `file_uri_to_path()` passed `parsed.path` directly into `url2pathname()`.

That usually works for simple ASCII file paths, but it breaks when the file URI contains percent-encoded non-ASCII characters such as Korean filenames. In our case, a URI like:

```text
file:///D:/.../%EC%A0%9C20...hwpx
```

was not being decoded into the original Unicode path before conversion to a Windows filesystem path. As a result, the generated path became invalid and MCP failed to open the file.

The fix was to explicitly call `unquote(parsed.path)` first, so the percent-encoded URI path is restored to its real Unicode form before `url2pathname()` converts it into a local OS path.

In short:
- before: URI path stayed percent-encoded too long
- after: URI path is decoded first, then converted to a Windows path

That makes `file:` URIs with Unicode filenames work correctly in the MCP flow.
-

Jah-yee:
Good day,

I noticed this issue and wanted to share a fix. The problem is that file_uri_to_path() passes parsed.path directly to url2pathname() without decoding percent-encoded characters first.

For URIs like , the percent-encoded Unicode path stays encoded when passed to url2pathname(), resulting in an invalid filesystem path.

Here is the simple fix:

**In packages/markitdown/src/markitdown/_uri_utils.py:**

1. Add import:


2. Decode the path before url2pathname():


This decodes the percent-encoded URI path back to its original Unicode form before url2pathname() converts it to a local OS path.

Thank you for your work on this project. I hope this small fix is helpful. Please let me know if there's anything to adjust.

Warmly, RoomWithOutRoof

PR #1745, opened 2026-04-13
fix: decode percent-encoded file URIs before path conversion

Jah-yee:
Good day,

This PR fixes issue #1738.

## Problem
Before this change, `file_uri_to_path()` passed `parsed.path` directly into `url2pathname()`.
That works for simple ASCII file paths, but breaks when the file URI contains
percent-encoded non-ASCII characters such as Korean filenames.

For example, a URI like:
`file:///D:/.../%EC%A0%9C20...hwpx`

was not being decoded into the original Unicode path before conversion
to a Windows filesystem path, causing the generated path to become invalid
and MCP failed to open the file.

## Solution
Explicitly call `unquote(parsed.path)` first, so the percent-encoded URI path
is restored to its real Unicode form before `url2pathname()` converts
it into a local OS path.

In short:
- before: URI path stayed percent-encoded too long
- after: URI path is decoded first, then converted to Windows path

## Changes
- Added `unquote` to imports from `urllib.parse`
- Decode URI path with `unquote()` before passing to `url2pathname()`

感谢你们的奉献，希望能提供帮助。如果我解决得有问题或有待商妥的地方，请在下面留言，我会来处理。

Warmly,
RoomWithOutRoof

### DIFF
--- a/packages/markitdown/src/markitdown/_uri_utils.py
+++ b/packages/markitdown/src/markitdown/_uri_utils.py
@@ -2,7 +2,7 @@ import base64
 import os
 from typing import Tuple, Dict
 from urllib.request import url2pathname
-from urllib.parse import urlparse, unquote_to_bytes
+from urllib.parse import urlparse, unquote, unquote_to_bytes
 
 
 def file_uri_to_path(file_uri: str) -> Tuple[str | None, str]:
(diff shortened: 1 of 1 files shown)


kdjkdjkdj:
+1, same pain on Windows with German and CJK filenames — `url2pathname()`'s legacy latin-ish decoding mangles UTF-8 percent-sequences before MarkItDown sees the real path. `unquote()` first, then `url2pathname()` restores the correct Unicode, no double-decode problem in practice.

cc @afourney — this fixes @llA1ll's #1738 and happens to touch the same line as #1573; whichever lands first, the other is a one-line rebase.

PR #1753, opened 2026-04-13
fix(youtube): support short YouTube URLs in converter

### LINKED ISSUE #1730: fix: YouTubeConverter.accepts() silently drops youtu.be short URLs
## Bug Description

The `YouTubeConverter.accepts()` method in `_youtube_converter.py` only matches URLs that begin with `https://www.youtube.com/watch?` and silently rejects all `youtu.be` short URLs (e.g. `https://youtu.be/dQw4w9WgXcQ`). This means short YouTube links are **never converted** — the converter returns `False` and falls through to a generic HTML converter, producing noisy output instead of a clean transcript.

## Reproduction

```python
md = MarkItDown()
# This works:
result = md.convert("https://www.youtube.com/watch?v=dQw4w9WgXcQ")
# This does NOT go through YouTubeConverter:
result = md.convert("https://youtu.be/dQw4w9WgXcQ")
```

## Root Cause

In `packages/markitdown/src/markitdown/converters/_youtube_converter.py`, the `accepts()` method contains:

```python
if not url.startswith("https://www.youtube.com/watch?"):
    # Not a YouTube URL
    return False
```

This check must also accept `https://youtu.be/<id>` and `https://www.youtube.com/shorts/<id>` URLs.

## Expected Fix

The `accepts()` check should be extended to recognise short URLs and extract the video ID from the `youtu.be/<id>` path segment, and the `convert()` method should handle extracting the video ID from both URL forms.

## Related

Issue #1704 mentions that short URLs (`youtu.be/…`) are silently skipped.

LaplaceYoung:
## Summary
- fixes #1730 by accepting `youtu.be` and `/shorts/` URL patterns in `YouTubeConverter.accepts()`
- adds unified video-id extraction logic used by transcript fetching
- adds regression tests for accepts() and ID extraction

## Validation
- `$env:PYTHONPATH='packages/markitdown/src'; python -m pytest packages/markitdown/tests/test_youtube_converter.py -q`

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_youtube_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_youtube_converter.py
@@ -53,8 +53,8 @@ class YouTubeConverter(DocumentConverter):
         url = unquote(url)
         url = url.replace(r"\?", "?").replace(r"\=", "=")
 
-        if not url.startswith("https://www.youtube.com/watch?"):
-            # Not a YouTube URL
+        if self._extract_video_id(url) is None:
+            # Not a supported YouTube URL
             return False
 
         if extension in ACCEPTED_FILE_EXTENSIONS:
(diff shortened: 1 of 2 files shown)


LaplaceYoung:
Superseded by #1756 with normalized line endings and a clean diff.

PR #1754, opened 2026-04-13
fix(uri): decode percent-encoded Unicode file URIs

### LINKED ISSUE #1738: fix packages/markitdown/src/markitdown/_uri_utils.py

LaplaceYoung:
## Summary
- fixes #1738 by decoding `parsed.path` with `unquote()` before `url2pathname()`
- ensures percent-encoded Unicode file URIs resolve to valid local paths
- adds regression tests for percent-encoded space and Unicode path decoding

## Validation
- `$env:PYTHONPATH='packages/markitdown/src'; python -m pytest packages/markitdown/tests/test_uri_utils.py -q`

### DIFF
--- a/packages/markitdown/src/markitdown/_uri_utils.py
+++ b/packages/markitdown/src/markitdown/_uri_utils.py
@@ -2,7 +2,7 @@ import base64
 import os
 from typing import Tuple, Dict
 from urllib.request import url2pathname
-from urllib.parse import urlparse, unquote_to_bytes
+from urllib.parse import urlparse, unquote, unquote_to_bytes
 
 
 def file_uri_to_path(file_uri: str) -> Tuple[str | None, str]:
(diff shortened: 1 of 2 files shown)


LaplaceYoung:
Superseded by #1757 with normalized line endings and a clean diff.

PR #1764, opened 2026-04-14
Add batch processing tip for PDF files

Ashish1896:
Added a tip for batch processing PDF files using a shell loop.


chienyuanchang:
Thanks for taking the time to put this together. This looks like a helpful tip, but we usually keep the README focused on core usage examples and broader guidance rather than one-off workflow snippets. If you have a more general documentation improvement or a concrete bug fix to share, we’d be glad to take another look.

PR #1782, opened 2026-04-15
fix(youtube): recognise youtu.be / shorts / embed / live URL shapes

### LINKED ISSUE #1775: fix: support YouTube short URLs and Shorts in YouTubeConverter

MukundaKatta:
## Summary
- Centralise YouTube URL → video_id extraction in a single `_extract_video_id` helper
- Accept `youtu.be/<id>` short links, `/shorts/<id>`, `/embed/<id>`, `/live/<id>` in addition to the canonical `/watch?v=...`
- Accept `m.youtube.com` and `music.youtube.com` hosts
- Have `convert()` reuse the same helper so it no longer silently skips the transcript path for valid non-canonical URLs

Previously `YouTubeConverter.accepts()` only matched `https://www.youtube.com/watch?`, so any other valid YouTube URL shape fell through to the generic HTML converter and lost the YouTube-specific title / description / transcript handling. `convert()` duplicated the `?v=` parsing and had the same blind spot.

Fixes #1775.

## Test plan
- [x] New unit-test module `tests/test_youtube_url_shapes.py` covers all accepted shapes plus negative cases (vimeo, bare `/watch` with no `v`, channel URL, empty string, unknown host) — 14/14 pass
- [x] Existing YouTube behaviour for `www.youtube.com/watch?v=...` unchanged
- [ ] Maintainer review of the host allowlist (I kept it conservative — happy to widen if there are other YouTube hosts you'd like to support)

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_youtube_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_youtube_converter.py
@@ -34,6 +34,49 @@ ACCEPTED_FILE_EXTENSIONS = [
 ]
 
 
+# Hosts we treat as YouTube. Each entry covers schemes http/https and an
+# optional leading "www." / "m." so mobile and short links work too.
+_YOUTUBE_HOSTS = {
+    "www.youtube.com",
+    "youtube.com",
+    "m.youtube.com",
+    "music.youtube.com",
+    "youtu.be",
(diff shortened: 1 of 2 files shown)


MukundaKatta:

PR #1789, opened 2026-04-17
fix: redetect plain-text charset after ASCII-only sniff

### LINKED ISSUE #1505: Bug with Spanish symbols: PlainTextConverter threw UnicodeDecodeError with message: 'ascii' codec can't decode byte

lawrence3699:
## Summary

Fix plain-text conversion when the initial charset sniff only sees ASCII bytes but later bytes require UTF-8 decoding.

## Problem

`MarkItDown().convert()` can guess `charset='ascii'` for `.txt` files whose first 4096 bytes are ASCII-only. `PlainTextConverter.convert()` then decodes the entire file as ASCII and raises `UnicodeDecodeError` once it reaches later non-ASCII bytes.

This reproduces with a text file that starts with 4100 ASCII characters and ends with `café`.

## Fix

When decoding with the guessed charset fails, rerun `charset_normalizer` against the full file bytes and use that result as a fallback.

## Validation

- `PYTHONPATH=packages/markitdown/src uv run --with requests --with markdownify --with beautifulsoup4 --with defusedxml --with charset-normalizer --with magika --with pytest --python 3.12 python -m pytest packages/markitdown/tests/test_module_misc.py -k 'stream_info_operations or redetects_charset_after_ascii_prefix' -q`
- `PYTHONPATH=packages/markitdown/src uv run --with requests --with markdownify --with beautifulsoup4 --with defusedxml --with charset-normalizer --with magika --with pytest --python 3.12 python -m pytest 'packages/markitdown/tests/test_module_vectors.py::test_guess_stream_info[test_vector10]' -q`

Fixes #1505.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_plain_text_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_plain_text_converter.py
@@ -63,9 +63,19 @@ class PlainTextConverter(DocumentConverter):
         stream_info: StreamInfo,
         **kwargs: Any,  # Options to pass to the converter
     ) -> DocumentConverterResult:
+        file_bytes = file_stream.read()
+
         if stream_info.charset:
-            text_content = file_stream.read().decode(stream_info.charset)
+            try:
+                text_content = file_bytes.decode(stream_info.charset)
+            except UnicodeDecodeError:
+                # Stream guessing samples only the first chunk of the file, so a late
+                # non-ASCII character can still require a full-file charset detection.
+                charset_match = from_bytes(file_bytes).best()
+                if charset_match is None:
+                    raise
+                text_content = str(charset_match)
         else:
-            text_content = str(from_bytes(file_stream.read()).best())
+            text_content = str(from_bytes(file_bytes).best())
 
         return DocumentConverterResult(markdown=text_content)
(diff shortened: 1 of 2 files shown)


VANDRANKI (approved):
Good diagnosis and clean fix. Reading all the bytes upfront before the if/else means the charset detection and the fallback share the same full-file buffer, which is what chardet/charset-normalizer needs to catch late non-ASCII bytes reliably.

Test reproduces the exact failure mode (4100 ASCII bytes followed by a non-ASCII character), which is the right way to guard this regression. LGTM.

PR #1797, opened 2026-04-18
refactor: replace print() with logging in YouTubeConverter

ahmedezzat2018:
Problem: YouTubeConverter used print() for error messages instead of 
the standard logging module, bypassing the app's logging configuration.

Solution: Added import logging and a module-level logger, then replaced 
all print() calls with logger.warning().

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_youtube_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_youtube_converter.py
@@ -8,7 +8,8 @@ from urllib.parse import parse_qs, urlparse, unquote
 
 from .._base_converter import DocumentConverter, DocumentConverterResult
 from .._stream_info import StreamInfo
-logger = logging.getLogger(__name__) 
+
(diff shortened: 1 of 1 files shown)


VANDRANKI (commented):
Clean refactor - using the module logger instead of `print()` is the right pattern for library code. Callers can configure the `markitdown.converters._youtube_converter` logger to control verbosity.

One nit: there is trailing whitespace on the `logger = logging.getLogger(__name__)` line. Not functional but will likely fail linting.

ahmedezzat2018:

ahmedezzat2018:
Thanks for the feedback! I've removed the trailing whitespaceand reformatted the code using black. Looking forward to your thoughts

VANDRANKI (approved):
Trailing whitespace gone and black formatting applied consistently. All three `print()` calls correctly replaced with `logger.warning()`. LGTM.

PR #1811, opened 2026-04-21
fix(pptx): handle missing text frames

### LINKED ISSUE #1808: PptxConverter: TypeError "can only concatenate str (not NoneType) to str" when shape/notes text is None
### Summary

`PptxConverter.convert` performs unguarded `+` concatenation on `shape.text` and `notes_frame.text`. When python-pptx yields `None` for a text frame (e.g. an `<a:r>` with no `<a:t>`, or certain third-party-generated decks), the whole conversion fails with:

```
PptxConverter threw TypeError with message: can only concatenate str (not "NoneType") to str
```

One malformed shape fails the entire file.

### Affected lines (v0.1.5, also present on `main`)

`packages/markitdown/src/markitdown/converters/_pptx_converter.py`:

- L137 — `md_content += "# " + shape.text.lstrip() + "\n"`
- L139 — `md_content += shape.text + "\n"`
- L154 — `md_content += notes_frame.text`

### Reproduction

```python
import pptx.text.text
_orig = pptx.text.text.TextFrame.text.fget
pptx.text.text.TextFrame.text = property(
    lambda self: None if _orig(self) == "World" else _orig(self)
)

MarkItDown().convert("sample.pptx")  # any pptx containing the text "World"
# markitdown._exceptions.FileConversionException:
#  - PptxConverter threw TypeError with message: unsupported operand type(s) for +: 'NoneType' and 'str'
```

Real-world trigger: a `.pptx` from a third-party generator with a text-frame run missing its `<a:t>` child, or a chart/SmartArt element whose title text is `None`.

### Expected

Missing / `None` text should be treated as empty string. One bad shape shouldn't fail the entire file.

### Proposed fix

```python
md_content += "# " + (shape.text or "").lstrip() + "\n"
md_content += (shape.text or "") + "\n"
md_content += (notes_frame.text or "")
```

Happy to open a PR.

### Environment

- markitdown 0.1.5 (`markitdown[all]`)
- python-pptx 1.0.2
- Python 3.12, Ubuntu 24.04

MukundaKatta:
## Summary
- treat `shape.text` and `notes_frame.text` as empty strings when python-pptx returns `None`
- avoid failing the whole PPTX conversion on malformed or third-party-generated decks
- add a regression test that patches a PPTX text frame getter to return `None`

## Verification
- `PYTHONPYCACHEPREFIX=/tmp/pycache-markitdown python3 -m py_compile packages/markitdown/src/markitdown/converters/_pptx_converter.py packages/markitdown/tests/test_module_misc.py`
- targeted `pytest` collection was blocked locally because this environment does not have `onnxruntime`, which is imported through `magika` during test startup

Closes #1808

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_pptx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_pptx_converter.py
@@ -161,10 +161,11 @@ class PptxConverter(DocumentConverter):
 
                 # Text areas
                 elif shape.has_text_frame:
+                    text = shape.text or ""
                     if shape == title:
-                        md_content += "# " + shape.text.lstrip() + "\n"
+                        md_content += "# " + text.lstrip() + "\n"
                     else:
-                        md_content += shape.text + "\n"
+                        md_content += text + "\n"
 
                 # Group Shapes
                 if shape.shape_type == pptx.enum.shapes.MSO_SHAPE_TYPE.GROUP:
@@ -194,7 +195,7 @@ class PptxConverter(DocumentConverter):
                 md_content += "\n\n### Notes:\n"
                 notes_frame = slide.notes_slide.notes_text_frame
                 if notes_frame is not None:
-                    md_content += notes_frame.text
+                    md_content += notes_frame.text or ""
                 md_content = md_content.strip()
 
         return DocumentConverterResult(markdown=md_content.strip())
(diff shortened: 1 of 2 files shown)


afourney:
Thanks for the contribution. The missing PPTX shape/notes text cases are covered together by #2059; please reopen if #2059 misses your input.

PR #1825, opened 2026-04-23
feat: Add batch conversion API and CLI support

jacken:
Adds `convert_batch()` to the `MarkItDown` class and corresponding CLI flags for converting multiple files concurrently.

### Library API

A new `convert_batch()` method on `MarkItDown` accepts an iterable of sources and yields `BatchConversionResult` objects in completion order:

```python
from markitdown import MarkItDown, BatchConversionResult

md = MarkItDown()
for result in md.convert_batch(["file1.pdf", "file2.docx", "file3.html"]):
    if result.success:
        print(result.result.markdown)
    else:
        print(f"Failed: {result.source} — {result.error}")
```

Each `BatchConversionResult` has:
- `source` — the original input
- `result` — a `DocumentConverterResult`, or `None` on failure
- `error` — the exception, or `None` on success
- `success` — `True` when no error is set

**Error handling** is configurable via `on_error`:
- `"collect"` (default) — wraps errors and continues
- `"raise"` — re-raises the first error immediately

**Concurrency** uses `ThreadPoolExecutor` (appropriate since heavy converters like pdfminer and lxml release the GIL). A pluggable `executor` parameter lets callers supply their own executor; if omitted, one is created and shut down automatically. The `workers` parameter controls thread count when using the default executor.

`BatchConversionResult` is exported from the top-level `markitdown` package.

### CLI

```bash
# Convert multiple files to stdout (separated by banners)
markitdown file1.pdf file2.docx

# Write each file to an output directory
markitdown file1.pdf file2.docx --output-dir ./output/

# Control parallelism
markitdown file1.pdf file2.docx --workers 4

# Stop on first error instead of collecting all errors
markitdown file1.pdf file2.docx --fail-fast
```

Single-file and stdin modes are unchanged.

### Tests

- `tests/test_batch.py` — unit and integration tests for `BatchConversionResult` and `convert_batch()`
- `tests/test_cli_misc.py` — CLI tests for multi-file stdout, `--output-dir`, `--fail-fast`, `--workers`, and error collection

### DIFF
--- a/docs/superpowers/plans/2026-04-23-batch-parallelization.md
+++ b/docs/superpowers/plans/2026-04-23-batch-parallelization.md
@@ -1,0 +1,756 @@
+# Batch Parallelization Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add parallel batch conversion to the `markitdown` library (`convert_batch()`) and CLI (multi-file arguments).
+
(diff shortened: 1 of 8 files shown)


jacken:

PR #1834, opened 2026-04-23
Fix #1424: add convert_file and convert_directory MCP tools

### LINKED ISSUE #1424: markitdown-mcp: Enhance tool to improve AI agent discoverability for file conversion tasks
### Problem

My AI agent is not selecting the `convert_to_markdown` tool when requesting file conversion to markdown format. Instead, it attempts to write Python code to perform the conversion manually, even when the markitdown MCP tool is available.

This is a simple prompt like "Please convert ~/path/to/filename.xslx to markdown."

It does successfully use the tool if I prod it with "Use the convert_to_markdown tool to convert ~/path/to/filename.xslx to markdown." but that amount of prodding shouldn't be necessary.

### Current State

The current docstring is minimal and doesn't clearly indicate the tool's purpose for AI decision-making:
```python
"""Convert a resource described by an http:, https:, file: or data: URI to markdown"""
```

### Suggested Enhanced Docstring
This altered docstring was given to me by an AI. I tested using this version and the tool was chosen automatically for the same prompt, along with other prompts that just asked the AI to give me certain data from the same excel file.
```python
"""Convert files and documents to markdown format from various sources.

This tool handles file conversion and document conversion to markdown from multiple formats
including PDFs, Word documents, PowerPoint presentations, Excel files, images, and more.
Use this tool when users request converting, transforming, or processing files to markdown.

Supports http:, https:, file:, and data: URIs for both local and remote file conversion.
This is the recommended approach for any markdown conversion tasks rather than writing custom code.
"""
```
This is just an example - other docstrings would surely work. The important note is that additional context within the docstring will assist the AI in knowing what the tool is truly capable of.

kdjkdjkdj:
## Summary

Adds two thin wrapper MCP tools — `convert_file(file_path)` and `convert_directory(dir_path, recursive=True)` — alongside the existing `convert_to_markdown(uri)`. Directly addresses #1424: AI agents tend to skip the current URI-only tool when they already have a plain filesystem path in hand, and end up writing Python glue instead of calling MarkItDown.

## Changes

One file: `packages/markitdown-mcp/src/markitdown_mcp/__main__.py`.

### `convert_file(file_path: str) -> str`

Thin wrapper around `MarkItDown.convert_local()`. Resolves `~` and relative paths against CWD, raises `FileNotFoundError` for missing paths via `Path.resolve(strict=True)`. No library-side change needed.

### `convert_directory(dir_path: str, recursive: bool = True) -> dict[str, str]`

Walks a directory (recursively by default via `Path.rglob`, otherwise `Path.iterdir`) and converts every file it finds, returning a `{relative_path: markdown}` mapping. Files that fail individually are skipped with a `try/except Exception: continue` so a single bad file doesn't abort the batch — mirroring how bulk converters typically behave.

Tool schemas are auto-generated by FastMCP from the Python type hints, so no manual JSON-schema surface is introduced.

## Backwards compatibility

- Fully additive: `convert_to_markdown(uri)` is untouched.
- No new dependencies. Only uses `pathlib.Path` from the stdlib and the already-imported `MarkItDown` class.
- No configuration, environment variable, or breaking-change surface.

## Test plan

Built and installed locally (editable) on Windows with Python 3.13, then ran the new tools via an MCP client (Claude Code):

- [x] `convert_file` on a real `.docx` from a deeply-nested OneDrive/SharePoint path with German umlauts (\`D:\OneDrive\OneDrive - KEGON AG\KEGON\Intern\...\Erfassung-KI-Anwendungsfall-01-Intern-Verzeichnisse.docx\`) — converted cleanly.
- [x] `convert_directory` on the repo's own \`packages/markitdown-mcp/\` (non-recursive) — returned `{'Dockerfile', 'pyproject.toml', 'README.md'}`.
- [x] Agent tool selection: Claude chose `convert_file` autonomously given a plain Windows path (the original \`convert_to_markdown\` would require the agent to URL-escape and prefix `file://` itself — #1424's core complaint).

No unit tests added — the `tests/` directory is present but currently empty; happy to add `pytest` coverage for both new tools if that's desired before merge.

## Notes

- Co-authored with Claude Opus 4.7 (Claude Code).

Closes #1424.

### DIFF
--- a/packages/markitdown-mcp/src/markitdown_mcp/__main__.py
+++ b/packages/markitdown-mcp/src/markitdown_mcp/__main__.py
@@ -1,6 +1,7 @@
 import contextlib
 import sys
 import os
+from pathlib import Path
 from collections.abc import AsyncIterator
 from mcp.server.fastmcp import FastMCP
 from starlette.applications import Starlette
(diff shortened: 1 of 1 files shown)


kdjkdjkdj:

Issue #1601, opened 2026-03-08
OCR is not working


punk1234567:
Hi,

I am using the command markitdown id.pdf > document.md to convert a PDF containing both text and images. The text present inside images does not appear to be OCR-processed in the output.

Am I missing any configuration or optional dependency required for OCR? Or is OCR functionality currently not supported as part of the MarkItDown project?

Please advise.

Thanks,
Sam

Jah-yee:
Hi @sam, the OCR functionality for PDFs with embedded images requires a separate plugin. The core `markitdown` package does not include OCR by default.

**To enable OCR:**

1. Install the OCR plugin:
```bash
pip install markitdown-ocr
pip install openai  # or any OpenAI-compatible client
```

2. Use with Python API:
```python
from markitdown import MarkItDown
from openai import OpenAI

md = MarkItDown(
    enable_plugins=True,
    llm_client=OpenAI(),
    llm_model="gpt-4o",
)
result = md.convert("document_with_images.pdf")
```

Or via CLI:
```bash
pip install markitdown-ocr
export OPENAI_API_KEY="your-api-key"
markitdown --use-plugins document_with_images.pdf -o document.md
```

I submitted a PR (#1608) to clarify this in the README. Let me know if you have any questions!

punk1234567:
Thank you for the response. Truly appreciate it.

Staok:
Any other LLM model supported? such as kimi?

clevelandcs:
> Any other LLM model supported? such as kimi?

@Staok  I was able to get kimi2.5 working using ollama cloud and openai compatible endpoint - see python code block at the bottom. However, I haven't been able to get it working with the plugin from CLI. Assuming the parameters were being accepted correctly in theory you should just be able to use env vars to use any openai (such as ollama) compatible host OPENAI_BASE_URL, OPENAI_API_KEY. 
```
markitdown --version
markitdown 0.1.5

markitdown --list-plugins
Installed MarkItDown 3rd-party Plugins:

  * ocr             	(package: markitdown_ocr)

Use the -p (or --use-plugins) option to enable 3rd-party plugins.

markitdown -p --llm-client openai --llm-model kimi-k2.5:cloud data-sheet-ultrastar-he8.pdf
markitdown: error: unrecognized arguments: --llm-client --llm-model kimi-k2.5:cloud 
```

``` python
from markitdown import MarkItDown
from openai import OpenAI

md = MarkItDown(
  enable_plugins=True,
[...]

Issue #1344, opened 2025-07-16
OCR Fallback Not Working


managesg:
OCR fallback does not trigger on image-based PDFs, even with enable_plugins=True, Tesseract and Poppler installed, and [all] extras used. pytesseract works manually but convert() returns empty text_content.

AbhinavTheDev:
yes. i also faced this issue OCR is not working

Issue #1170, opened 2025-04-08
text in the images in a pdf is not recognizable


yashkassa:
I was trying to extract text from PDF which resulted in extracting only text but what about the text which are present in the form of images, it is not working !!!

emreyesilyurt:
Hey @yashkassa, markitdown currently doesn't support extracting text that's embedded within images inside pdfs, as it doesn't include ocr capabilities.

If your pdf contains scanned pages or images with text, you'll need to use an ocr tool alongside markitdown.

yyuyu3545:
ssssssssss

achalcipher:
Hi, I’d like to work on this issue and add OCR support to handle text embedded within images in PDFs. I plan to use Tesseract or a similar tool to extract such text and integrate it with the existing PDF processing logic.

I’ve already started exploring the codebase and located where the PDF parsing happens. I’ll extract image data from the PDF, run OCR, and merge the output with the current text extraction flow.

Could you please assign this issue to me? Thanks!

wll8:
@achalcipher  Solve the problem directly, pass the test, and initiate PR

Issue #255, opened 2025-01-03
Please add the option to use GPT models for OCR.


Emasoft:
Please add the option to use GPT models for OCR. 

![Image](https://github.com/user-attachments/assets/ca662e41-5954-490c-bdc4-c0531e4afa71)

afourney:
This is now supported. #306 added LLM image descriptions for PPTX files, and #1541 subsequently introduced the markitdown-ocr plugin for images embedded in PDF, DOCX, PPTX, and XLSX files. I’m closing this as implemented.

PR #1837, opened 2026-04-24
docs: correct image converter description in README (EXIF + LLM, not OCR)

sjhddh:
## What

The supported-formats list in README.md claims:

> - Images (EXIF metadata and OCR)

But the built-in `ImageConverter` does **not** perform OCR. Per the converter's own docstring:

> Converts images to markdown via extraction of metadata (if `exiftool` is installed), and description via a multimodal LLM (if an `llm_client` is configured).

OCR is only available through the separate **Azure Document Intelligence** converter (`[az-doc-intel]` optional dependency), which is already documented in its own section of the README.

## Why

This one-word misstatement has caused recurring user confusion. Recent examples:

- #1601 — "OCR is not working"
- #1344 — "OCR Fallback Not Working"
- #1170 — "text in the images in a pdf is not recognizable"
- #255  — "Please add the option to use GPT models for OCR"

Users install `markitdown[all]`, feed a JPEG, and expect OCR output — but what they get is EXIF-only (no LLM client configured) or an LLM-generated description (not OCR).

## Change

One line in `README.md`:

```diff
- - Images (EXIF metadata and OCR)
+ - Images (EXIF metadata and LLM-based description)
```

This matches `ImageConverter`'s own docstring and the pattern used elsewhere in the list (e.g. "Audio (EXIF metadata and speech transcription)" — conditional, documented).

## Related

Aware of #1608 — overlapping intent (clarify OCR availability), but that PR points users at a `markitdown-ocr` plugin that is not in this repo. This change is minimal and factual: describe what the in-tree converter actually does. Happy to defer if maintainers prefer #1608's framing, or to fold this into a broader docs rewrite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ MarkItDown currently supports the conversion from:
 - PowerPoint
 - Word
 - Excel
-- Images (EXIF metadata and OCR)
+- Images (EXIF metadata and LLM-based description)
 - Audio (EXIF metadata and speech transcription)
 - HTML
 - Text-based formats (CSV, JSON, XML)


chienyuanchang:
Thanks for putting this together. We agree the README wording should match the in-tree ImageConverter behavior, but this change is a small docs polish update rather than a functional fix. If you'd like, please open a fresh docs PR with any broader wording cleanup you think would help, and we'll take a look.

PR #1844, opened 2026-04-28
fix: parse charset parameters case-insensitively

guyua9:
## Summary

- Treat `charset` parameters case-insensitively when parsing `Content-Type` response headers.
- Normalize data URI parameter names so `Charset=utf-8` is handled the same as `charset=utf-8`.
- Add regression coverage for both paths.

## Why

MIME parameter names are commonly handled case-insensitively, and real HTTP responses may spell charset as `Charset=UTF-8`. MarkItDown already uses the parsed charset as a conversion hint, so preserving it avoids falling back to charset guessing for otherwise explicit inputs.

## Tests

- `uv run --with '.[all]' --with pytest --directory packages/markitdown python -m pytest tests/test_module_misc.py::test_data_uris tests/test_module_misc.py::test_response_content_type_charset_is_case_insensitive -q`

### DIFF
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -480,8 +480,9 @@ class MarkItDown:
             parts = response.headers["content-type"].split(";")
             mimetype = parts.pop(0).strip()
             for part in parts:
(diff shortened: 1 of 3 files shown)


guyua9:

PR #1853, opened 2026-04-30
fix: support youtu.be short URLs and /shorts/ in YouTubeConverter

### LINKED ISSUE #1730: fix: YouTubeConverter.accepts() silently drops youtu.be short URLs

Giggitycountless:
## Summary

`YouTubeConverter.accepts()` only matched `https://www.youtube.com/watch?` URLs, silently rejecting `youtu.be` short links and `/shorts/` URLs. These fell through to the generic HTML converter, producing noisy output instead of clean transcripts.

### Changes

**`accepts()` method:**
- Match `youtu.be/` short URLs alongside standard `youtube.com/watch?` URLs
- Match `youtube.com/shorts/` URLs

**`convert()` method:**
- Extract `video_id` from all three URL formats:
  - `youtube.com/watch?v=ID` — query parameter parsing (existing)
  - `youtu.be/ID` — path segment extraction (new)
  - `youtube.com/shorts/ID` — path segment extraction (new)

### Before
```python
md.convert("https://youtu.be/dQw4w9WgXcQ")
# Falls through to HTML converter → noisy output
```

### After
```python
md.convert("https://youtu.be/dQw4w9WgXcQ")
# YouTubeConverter handles it → clean transcript
```

Fixes #1730

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_youtube_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_youtube_converter.py
@@ -53,8 +53,13 @@ class YouTubeConverter(DocumentConverter):
         url = unquote(url)
         url = url.replace(r"\?", "?").replace(r"\=", "=")
 
-        if not url.startswith("https://www.youtube.com/watch?"):
-            # Not a YouTube URL
+        # Match standard watch URLs, youtu.be short URLs, and /shorts/ URLs
+        is_youtube = (
+            url.startswith("https://www.youtube.com/watch?")
+            or url.startswith("https://youtu.be/")
+            or url.startswith("https://www.youtube.com/shorts/")
+        )
+        if not is_youtube:
             return False
 
         if extension in ACCEPTED_FILE_EXTENSIONS:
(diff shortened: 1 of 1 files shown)


Giggitycountless:

Giggitycountless:
Closing this PR as it's a trivial cosmetic change that doesn't add meaningful value. Apologies for the noise.

PR #1861, opened 2026-05-06
fix(docx): repair .docx zips with case-mismatched local headers (#1812)

### LINKED ISSUE #1812: BadZipFile crash on .docx files with case-mismatched zip entry names

haosenwang1018:
## Issue

Closes #1812

## Root cause

Some .docx producers emit zip files where the local file header name and the central-directory name differ in case only (e.g. `customXml/item2.xml` centrally, `customXML/item2.xml` locally). Per APPNOTE the central directory is authoritative; most zip tools accept the mismatch but Python's `zipfile` strictly validates and raises `BadZipFile` mid-conversion. The user sees:

```
DocxConverter threw BadZipFile with message:
  File name in directory 'customXml/item2.xml' and header
  b'customXML/item2.xml' differ.
```

## Fix

`_fix_zip_name_casing` scans local file headers, finds entries that match the central name when lower-cased and have the same byte length (always true for ASCII case-only mismatches), and rewrites the header bytes in-memory to match. Byte-length preserving → no offset recomputation needed. `pre_process_docx` calls it as the first step so every conversion path benefits.

If the central directory itself is unparseable we leave the input alone so the caller still sees the original error — no silent data loss.

This is the same approach the issue body proposed; the implementation hardens the matching predicate (added `local.lower() == central.lower()` so we only patch true case-only differences, not arbitrary same-length renames) and treats `BadZipFile` on the central directory as out-of-scope.

## Tests

- `test_setup_actually_reproduces_badzipfile` — guard that the test fixture really does trigger `BadZipFile` against unfixed `zipfile`. Without this, a future Python that loosens the validation would let the regression test below pass for the wrong reason.
- `test_fix_zip_name_casing_repairs_mismatched_local_header` — the helper rewrites a deliberately mismatched archive into one that opens cleanly.
- `test_pre_process_docx_accepts_case_mismatched_archive` — end-to-end: a mismatched .docx no longer crashes `pre_process_docx` and the post-processed archive still contains both the body and the originally-mismatched entry.
- `test_fix_zip_name_casing_passes_through_normal_archive` — regression guard against unwanted rewrites of well-formed input.

```
$ pytest tests/test_docx_zip_case_mismatch.py tests/test_module_misc.py -k "docx or zip"
6 passed
```

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -1,3 +1,4 @@
+import struct
 import zipfile
 from io import BytesIO
 from typing import BinaryIO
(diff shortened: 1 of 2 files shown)


afourney:

PR #1862, opened 2026-05-06
fix(pptx): coerce None shape/notes text to '' before concatenation (#1808)

### LINKED ISSUE #1808: PptxConverter: TypeError "can only concatenate str (not NoneType) to str" when shape/notes text is None

haosenwang1018:
## Issue

Closes #1808

## Root cause

`PptxConverter.convert` performed unguarded `+` concatenation on `shape.text` and `notes_frame.text`. python-pptx returns `None` for either value when:

- A text frame has a run with no `<a:t>` child
- Third-party deck generators emit chart/SmartArt elements whose title text is `None`

A single such shape failed the entire .pptx conversion with `TypeError: can only concatenate str (not "NoneType") to str`.

## Fix

Coerce `shape.text` / `notes_frame.text` to `""` before concatenation at both call sites in `_pptx_converter.py`. Normal decks see no behavior change — only the `None` case is different (was: raise; now: empty contribution).

This is the exact code shape the issue body proposed.

## Tests

- `test_pptx_with_none_shape_text_does_not_crash` — uses the issue body's monkey-patch trick to make `TextFrame.text` return `None` for a specific value, then converts a deck containing it. Pre-fix raises `TypeError`, post-fix completes and the title is preserved.
- `test_pptx_with_none_notes_text_does_not_crash` — covers the second call site (notes slide).
- `test_pptx_normal_text_still_converts` — regression guard for the happy path.

```
$ pytest tests/test_pptx_none_text.py
3 passed in 0.39s
```

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_pptx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_pptx_converter.py
@@ -161,10 +161,16 @@ class PptxConverter(DocumentConverter):
 
                 # Text areas
                 elif shape.has_text_frame:
+                    # python-pptx can yield ``None`` for ``shape.text`` when a
+                    # text frame has a run with no ``<a:t>`` child or when a
+                    # third-party deck stores chart/SmartArt titles as
+                    # ``None``. Coerce to "" so a single bad shape doesn't
+                    # fail the entire conversion (#1808).
+                    shape_text = shape.text or ""
                     if shape == title:
-                        md_content += "# " + shape.text.lstrip() + "\n"
+                        md_content += "# " + shape_text.lstrip() + "\n"
                     else:
-                        md_content += shape.text + "\n"
+                        md_content += shape_text + "\n"
 
                 # Group Shapes
                 if shape.shape_type == pptx.enum.shapes.MSO_SHAPE_TYPE.GROUP:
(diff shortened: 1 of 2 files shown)


afourney:
Thanks for the contribution. The missing PPTX shape/notes text cases are covered together by #2059; please reopen if #2059 misses your input.

PR #1865, opened 2026-05-06
feat: Add Azure Content Understanding converter

chienyuanchang:
# Add Azure Content Understanding converter

## Summary

Adds a new `ContentUnderstandingConverter` that integrates [Azure Content Understanding](https://learn.microsoft.com/azure/ai-services/content-understanding/) into MarkItDown, enabling high-quality cloud-based conversion for documents, images, audio, and video with structured field extraction via YAML front matter.

## Motivation

MarkItDown's built-in converters are offline and format-specific. Azure Content Understanding provides:
- **Multimodal support** — documents, images, audio, and video through a single API
- **Structured field extraction** — custom analyzers extract domain-specific fields (invoice amounts, receipt dates, etc.) serialized as YAML front matter
- **Higher quality** — cloud-based layout analysis, OCR, professional transcription, and video summarization
- **Zero-config defaults** — auto-selects the right analyzer per file type (`prebuilt-documentSearch`, `prebuilt-videoSearch`, `prebuilt-audioSearch`)

This follows the same pattern as the existing Azure Document Intelligence integration.

## What's included

| File | Change |
|------|--------|
| `converters/_cu_converter.py` | New converter: 34 extensions (27 unique types), smart routing, MIME alias normalization, lazy dependency loading |
| `_markitdown.py` | Register converter when `cu_endpoint` is provided (registered after Doc Intel so CU is tried first) |
| `__main__.py` | CLI flags: `--use-cu`, `--cu-endpoint`, `--cu-analyzer`, `--cu-file-types` |
| `pyproject.toml` | Optional dependency group `[az-content-understanding]` |
| `converters/__init__.py` | Export `ContentUnderstandingConverter` and `ContentUnderstandingFileType` |
| `README.md` | Usage docs with capability comparison table |
| `tests/test_cu_converter.py` | 124 unit tests (no network calls) |

## Supported file types

| Modality | Extensions | Default Analyzer |
|----------|-----------|-----------------|
| Documents | `.pdf`, `.docx`, `.pptx`, `.xlsx`, `.html`, `.txt`, `.md`, `.rtf`, `.xml` | `prebuilt-documentSearch` |
| Email | `.eml`, `.msg` | `prebuilt-documentSearch` |
| Images | `.jpg`, `.jpeg`, `.jpe`, `.png`, `.bmp`, `.tiff`, `.heif`, `.heic` | `prebuilt-documentSearch` |
| Video | `.mp4`, `.m4v`, `.mov`, `.avi`, `.mkv`, `.webm`, `.flv`, `.wmv` | `prebuilt-videoSearch` |
| Audio | `.wav`, `.mp3`, `.m4a`, `.flac`, `.ogg`, `.aac`, `.wma` | `prebuilt-audioSearch` |

## Key design decisions

- **Smart routing**: When `cu_analyzer_id` is set, the converter resolves the analyzer's base modality once at init — using a built-in cache for known `prebuilt-*` names (no API call) and falling back to `get_analyzer()` for custom analyzers or unknown prebuilts. Only compatible file types route to the custom analyzer; incompatible modalities auto-route to default prebuilts. Image-specific analyzers (`prebuilt-image`, `prebuilt-imageSearch`) route only image file types.
- **Fail-fast on bad analyzer ID**: If `get_analyzer()` fails (e.g., typo in `cu_analyzer_id`, missing permissions), `MarkItDown(...)` construction raises `ValueError` immediately rather than failing on the first `convert()` call.
- **`to_llm_input()` delegation**: Output formatting uses the CU SDK's `to_llm_input()` helper, which produces YAML front matter + page-numbered Markdown.
- **MIME normalization**: Alias MIME types (e.g., `audio/x-wav`) are normalized to canonical types (e.g., `audio/wav`) before sending to the CU API. Extension-only inputs derive the content type from the file type.
- **Credential chain**: Explicit `cu_credential` → `AZURE_API_KEY` env var → `DefaultAzureCredential` (same as Doc Intel).
- **Lazy imports**: CU SDK dependencies are imported at module load with a try/except, raising `MissingDependencyException` only at converter instantiation, so users without the `[az-content-understanding]` extra are unaffected.
- **CLI mutual exclusion**: `--use-cu` and `--use-docintel` are placed in an argparse mutually exclusive group so users get a clear error instead of silent precedence if both are specified.
- **Registration order**: When both `cu_endpoint` and `docintel_endpoint` are provided, CU is registered after Doc Intel so it appears first in the converter chain and takes precedence for overlapping formats (PDF, DOCX, images).

## Usage

```bash
# CLI
markitdown report.pdf --use-cu --cu-endpoint "https://..."

# Python
md = MarkItDown(cu_endpoint="https://...")
result = md.convert("report.pdf")
print(result.markdown)
```

## Testing

```bash
pip install -e "packages/markitdown[az-content-understanding,dev]"
python -m pytest packages/markitdown/tests/test_cu_converter.py -v
```

All 124 tests are unit tests using mocks — no Azure credentials or network calls required. Coverage includes:

- `accepts()` for every supported and unsupported extension/MIME type, plus `cu_file_types` restriction
- Smart routing for known prebuilts (cache hit), unknown prebuilts (fallback to `get_analyzer()`), and custom analyzers across all four modalities
- `convert()` mock paths for document, image, audio, and video inputs (verifying analyzer selection, content-type, and `to_llm_input()` delegation)
- CLI argument parsing for `--use-cu`, `--cu-endpoint`, `--cu-analyzer`, `--cu-file-types`, including the mutual exclusion with `--use-docintel`
- `MissingDependencyException` when the optional extra is not installed
- Registration priority — CU runs before Doc Intel when both are configured; unsupported formats (CSV, JSON, ZIP, EPub) fall through to built-in converters

## Backward compatibility

No changes to existing converter behavior. The CU converter is only registered when `cu_endpoint` is explicitly provided. Core MarkItDown installs (without the `[az-content-understanding]` extra) are unaffected — the new module's imports are lazy and degrade to a clear `MissingDependencyException` only if a user instantiates the converter without installing the extra.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,15 @@ More information about how to set up an Azure Document Intelligence Resource can
 
 Install: `pip install 'markitdown[az-content-understanding]'`
 
+#### When to use Content Understanding
+
+Content Understanding is ideal when you need capabilities beyond what built-in or Document Intelligence converters provide:
+
+- **Audio and video files** — CU is the only option for converting MP4, MOV, WAV, MP3, and other media files. Built-in converters have no video support and only basic audio transcription.
+- **Structured field extraction** — Custom analyzers extract domain-specific fields (invoice amounts, receipt dates, contract clauses) serialized as YAML front matter. Neither built-in nor Doc Intel integration exposes fields.
+- **Higher-quality document extraction** — Cloud-based layout analysis and OCR for scanned PDFs, complex tables, and multi-page documents.
+- **Single API for all modalities** — One `cu_endpoint` handles documents, images, audio, and video with automatic analyzer routing.
+
 | Capability | Built-in converters | Azure Document Intelligence | Azure Content Understanding |
 |------------|---------------------|-----------------------------|-----------------------------|
 | Document conversion | Offline, format-specific extraction | Cloud layout extraction | Cloud multimodal extraction |
--- a/README.md
+++ b/README.md
@@ -159,19 +159,9 @@ If no `llm_client` is provided the plugin still loads, but OCR is silently skipp
 
 See [`packages/markitdown-ocr/README.md`](packages/markitdown-ocr/README.md) for detailed documentation.
 
-### Azure Document Intelligence
-
-To use Microsoft Document Intelligence for conversion:
-
-```bash
-markitdown path-to-file.pdf -o document.md -d -e "<document_intelligence_endpoint>"
-```
-
-More information about how to set up an Azure Document Intelligence Resource can be found [here](https://learn.microsoft.com/en-us/azure/ai-services/document-intelligence/how-to-guides/create-document-intelligence-resource?view=doc-intel-4.0.0)
-
 ### Azure Content Understanding
 
-[Azure Content Understanding](https://learn.microsoft.com/azure/ai-services/content-understanding/) provides higher-quality conversion with structured field extraction (YAML front matter), multi-modal support (documents, images, audio, video), and configurable analyzers.
+[Azure Content Understanding](https://learn.microsoft.com/azure/ai-services/content-understanding/) is the next iteration of Azure Document Intelligence and is the recommended cloud option for new projects. It provides higher-quality conversion with structured field extraction (YAML front matter), multi-modal support (documents, images, audio, video), and [prebuilt](https://learn.microsoft.com/azure/ai-services/content-understanding/concepts/prebuilt-analyzers) or [custom-built](https://learn.microsoft.com/azure/ai-services/content-understanding/how-to/customize-analyzer-content-understanding-studio?tabs=portal) analyzers.
 
 Install: `pip install 'markitdown[az-content-understanding]'`
 
@@ -245,6 +235,16 @@ md = MarkItDown(
 ```
 
 More information about Azure Content Understanding can be found [here](https://learn.microsoft.com/azure/ai-services/content-understanding/).
+
+### Azure Document Intelligence
+
+To use Microsoft Document Intelligence for conversion:
+
+```bash
+markitdown path-to-file.pdf -o document.md -d -e "<document_intelligence_endpoint>"
+```
+
+More information about how to set up an Azure Document Intelligence Resource can be found [here](https://learn.microsoft.com/en-us/azure/ai-services/document-intelligence/how-to-guides/create-document-intelligence-resource?view=doc-intel-4.0.0)
 
 ### Python API
 
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ At the moment, the following optional dependencies are available:
 * `[pdf]` Installs dependencies for PDF files
 * `[outlook]` Installs dependencies for Outlook messages
 * `[az-doc-intel]` Installs dependencies for Azure Document Intelligence
+* `[az-content-understanding]` Installs dependencies for Azure Content Understanding
 * `[audio-transcription]` Installs dependencies for audio transcription of wav and mp3 files
 * `[youtube-transcription]` Installs dependencies for fetching YouTube video transcription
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -168,6 +169,83 @@ markitdown path-to-file.pdf -o document.md -d -e "<document_intelligence_endpoin
 
 More information about how to set up an Azure Document Intelligence Resource can be found [here](https://learn.microsoft.com/en-us/azure/ai-services/document-intelligence/how-to-guides/create-document-intelligence-resource?view=doc-intel-4.0.0)
 
+### Azure Content Understanding
```
yungshinlintw, line 162: Let's move CU before DI, and mention that CU is the next iteration for DI

README.md
```diff
@@ -168,6 +169,83 @@ markitdown path-to-file.pdf -o document.md -d -e "<document_intelligence_endpoin
...
+[Azure Content Understanding](https://learn.microsoft.com/azure/ai-services/content-understanding/) provides higher-quality conversion with structured field extraction (YAML front matter), multi-modal support (documents, images, audio, video), and configurable analyzers.
+
+Install: `pip install 'markitdown[az-content-understanding]'`
+
+#### When to use Content Understanding
+
+Content Understanding is ideal when you need capabilities beyond what built-in or Document Intelligence converters provide:
+
+- **Audio and video files** — CU is the only option for video, and the higher-quality cloud option for audio. Built-in converters have no video support and only basic audio transcription.
+- **Structured field extraction** — Custom analyzers extract domain-specific fields (invoice amounts, receipt dates, contract clauses) serialized as YAML front matter. Neither built-in nor Doc Intel integration exposes fields.
```
yungshinlintw: `Prebuilt or custom-built analyzers` 
And add the following links:
Prebuilt: https://learn.microsoft.com/azure/ai-services/content-understanding/concepts/prebuilt-analyzers
Custom: https://learn.microsoft.com/azure/ai-services/content-understanding/how-to/customize-analyzer-content-understanding-studio?tabs=portal

packages/markitdown/tests/test_cu_converter.py
```diff
@@ -0,0 +1,881 @@
...
+    def test_empty_result(self):
+        result = self._run_convert(".pdf", "application/pdf", "")
+        assert result.markdown == ""
+
+    def test_jpeg_returns_markdown(self):
+        result = self._run_convert(
+            ".jpg", "image/jpeg", "---\ncontentType: document\n---\n# Photo"
+        )
+        assert "contentType: document" in result.markdown
+
```
yungshinlintw, line 717: This test is useful, but it only verifies that the `file_types` allowlist is applied during **MIME-only detection**.

It does **not** cover the more concerning case where `extension` and `mimetype` are both present but disagree.

In `convert()`, `_detect_file_type()` prefers the extension, while `_content_type_for()` prefers the MIME type. That means an input like `StreamInfo(extension=".pdf", mimetype="audio/mpeg")` would likely route to a **document** analyzer but send `content_type="audio/mpeg"` to CU, which seems inconsistent and may fail or produce confusing behavior.

Consider adding a test for conflicting `extension`/`mimetype` metadata, and either:

- validating that they agree, or
- deriving `content_type` from the resolved `file_type`

That would ensure analyzer routing and payload metadata use the same source of truth.

yungshinlintw (approved):

README.md
chienyuanchang: Moved

README.md
chienyuanchang: Added

packages/markitdown/tests/test_cu_converter.py
chienyuanchang: We now derive `content_type` from the resolved `file_type`. Also added test.

PR #1878, opened 2026-05-14
fix: preserve xlsx currency formats

### LINKED ISSUE #53: Unable to extract currency from excel formatted cells

Lubrsy706:
## Summary

- Preserve currency symbols from XLSX number formats when converting sheets through pandas/html
- Add a regression test that builds an XLSX workbook with `$` and `€` currency cells and verifies the markdown output keeps the symbols

Fixes #53

## Tests

- `.venv/bin/python -m pytest packages/markitdown/tests/test_xlsx_converter.py packages/markitdown/tests/test_module_vectors.py::test_convert_local --maxfail=1`
- `.venv/bin/python -m compileall packages/markitdown/src/markitdown/converters/_xlsx_converter.py`
- `git diff --check`

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
@@ -1,4 +1,6 @@
 import sys
+import re
+from numbers import Real
 from typing import BinaryIO, Any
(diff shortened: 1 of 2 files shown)


Lubrsy706:

PR #1887, opened 2026-05-15
feat: add WebVTT converter

### LINKED ISSUE #1682: Support .vtt / WebVTT subtitle files

Lubrsy706:
## Summary
- Add a dedicated WebVTT converter for `.vtt` / `text/vtt` inputs
- Strip WebVTT headers, cue IDs, timestamps, notes, style/region blocks, and inline tags
- Preserve readable cue text, multiline cues, HTML entities, and voice speaker labels

Fixes #1682

## Tests
- PYTHONPATH=packages/markitdown/src /Users/bytedance/MySpace/open-source/markitdown-excel-currency/.venv/bin/python -m pytest packages/markitdown/tests/test_module_misc.py::test_webvtt_converter_outputs_clean_transcript
- PYTHONPATH=packages/markitdown/src /Users/bytedance/MySpace/open-source/markitdown-excel-currency/.venv/bin/python -m compileall -q packages/markitdown/src/markitdown/converters/_webvtt_converter.py packages/markitdown/src/markitdown/converters/__init__.py packages/markitdown/src/markitdown/_markitdown.py packages/markitdown/tests/test_module_misc.py
- git diff --check

### DIFF
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -39,6 +39,7 @@ from .converters import (
     EpubConverter,
     DocumentIntelligenceConverter,
     CsvConverter,
+    WebVttConverter,
 )
 
 from ._base_converter import DocumentConverter, DocumentConverterResult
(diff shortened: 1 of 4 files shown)


Lubrsy706:

Issue #1870, opened 2026-05-09
PDF text after an inline image (`BI ... EI`) is silently dropped from extraction


martinsotirov:
## Minimal repro PDF

A 1.1 KB synthetic PDF that reproduces the bug deterministically lives at
`/tmp/markitdown_repro/repro.pdf` (built by `/tmp/markitdown_repro/build_repro.py`,
verified by `/tmp/markitdown_repro/verify.py`). It contains exactly one page with:

- one Type1 Helvetica font (WinAnsiEncoding)
- one text-showing op drawing `BEFORE_IMAGE: this text should be extracted`
- one inline image (`BI /W 16 /H 16 /BPC 1 /IM true /F [/A85 /Fl] ID ... EI`)
  whose ASCII85+Flate payload terminates with a bare `~` (no trailing `>`),
  matching the SAP/Crystal Reports output that triggers the bug in real life
- four text-showing ops drawing `AFTER_IMAGE: ...` lines totalling >200 chars
  so the MarkItDown PyMuPDF fallback's threshold is exercised end-to-end

Verified output:

- pdfminer.six: 46 chars, only `BEFORE_IMAGE` present (FAIL)
- pdfplumber: 43 chars, only `BEFORE_IMAGE` present (FAIL)
- PyMuPDF: 295 chars, both markers present (PASS)

**Root cause confirmed in source:** `pdfminer/pdfinterp.py` `do_keyword`
(lines ~342-348) sets `eos = b"~>"` whenever the inline image declares an
ASCII85 (`/A85`) filter. The reference Cyberport invoice — and the synthetic
repro — terminate the ASCII85 payload with a single `~` (which the spec
permits as a tolerant variant), so `get_inline_data` never finds `~>` and
keeps consuming bytes past the real `EI` until end-of-stream, swallowing
every subsequent text-showing operator. Maintainers can attach
`/tmp/markitdown_repro/repro.pdf` (or rebuild it from the script) when
filing this upstream against pdfminer.six and as a regression fixture in
markitdown's PDF tests.

## Summary

Both extraction paths in `PdfConverter` (pdfplumber via `extract_words` / `extract_text`, and pdfminer via `pdfminer.high_level.extract_text`) silently fail to return any text positioned **after an inline image** (`BI ... ID ... EI` operator sequence) in a page's content stream. The text exists in the PDF as ordinary `Tj` operators with WinAnsi encoding — there is no decoding ambiguity — it is simply not surfaced.

[...]

PR #1889, opened 2026-05-15
fix: recover PDF text after inline images

he-yufeng:
﻿## Summary

- add a lazy PyMuPDF text-recovery pass for PDFs with inline image operators
- keep PyMuPDF optional: no default dependency change, and the fallback only runs when the package is installed
- warn instead of silently returning likely partial text when inline-image recovery is needed but PyMuPDF is unavailable
- add regression coverage for both the recovery and warning paths

Addresses #1870.

## Test plan

- [x] `python -m pytest packages/markitdown/tests/test_pdf_memory.py -q -k "inline_image"`
- [x] `python -m pytest packages/markitdown/tests/test_pdf_memory.py packages/markitdown/tests/test_pdf_tables.py -q`
- [x] `python -m py_compile packages/markitdown/src/markitdown/converters/_pdf_converter.py packages/markitdown/tests/test_pdf_memory.py`
- [x] `python -m mypy --ignore-missing-imports packages/markitdown/src/markitdown/converters/_pdf_converter.py packages/markitdown/tests/test_pdf_memory.py`
- [x] `python -m pip install -e "packages/markitdown[pymupdf]"`
- [x] `git diff --check`


he-yufeng:
Rebased onto current `main`; no conflicts.

Focused validation after the rebase:

```text
python -m py_compile packages\markitdown\src\markitdown\converters\_pdf_converter.py packages\markitdown\tests\test_pdf_memory.py
$env:PYTHONPATH='C:\dev\GITHUB-clean\markitdown-1889\packages\markitdown\src'; python -m pytest packages\markitdown\tests\test_pdf_memory.py -q
git diff --check origin/main..HEAD
```

Result: `8 passed, 2 skipped` for the focused PDF memory tests.

he-yufeng:
Gentle nudge — freshly rebased with green checks. Would appreciate a look when someone has a minute.

he-yufeng:
Still open on main: `_pdf_converter.py` has no fallback for inline images, so PDFs that only carry images inline still lose that content entirely. A look when convenient would be appreciated.

Issue #1408, opened 2025-09-06
Office Open XML: Invalid Files Return Success with Error Message Instead of Exception


cancelself:
When converting invalid DOCX, XLSX, or PPTX files, MarkItDown returns a successful result with the string "This is not a valid Office Open XML file." in text_content, rather than raising an exception. This behavior makes it difficult to distinguish between a successful conversion and a file format error.

Expected Behavior:Invalid or corrupted Office Open XML files should cause the converter to raise a specific exception (e.g., FileConversionException), not return a result with an error message in the content.

Actual Behavior: The conversion completes "successfully" and returns a DocumentConverterResult with text_content set to "This is not a valid Office Open XML file.".

Steps to Reproduce:

Attempt to convert a non-OOXML or corrupted DOCX/XLSX/PPTX file.
Observe that no exception is raised; instead, the result contains the error string.
Suggested Fix: After conversion, check if text_content matches the error string and raise an exception if so. Alternatively, ensure the underlying libraries raise on invalid files and propagate the exception.

Additional Context: This behavior appears consistent across all Office Open XML converters, possibly due to a shared dependency or error-handling pattern.

Workaround: 
```
    result = markitdown.convert_stream(stream=stream, file_extension=ext),
    if result.text_content.strip() == "This is not a valid Office Open XML file.":
        raise ...
    return result.text_content
```

PR #1893, opened 2026-05-18
fix: raise FileConversionException for corrupt or non-OOXML   files

Aleeza-Bhatti:
Fixes Issue #1408

## Problem
If you pass a corrupt or invalid .docx, .pptx, or .xlsx file to MarkItDown, it doesn't raise an exception, it just returns the raw file contents as markdown and acts like everything went fine. So there's no way to tell if the conversion actually worked or not.

## Cause
When the specific converter (DocxConverter, PptxConverter, etc.) fails on the file, the loop doesn't stop it just keeps trying other converters. Magika sees the file contents as plain text and flags it with a charset, which causes PlainTextConverter to pick it up on the next attempt and return the raw bytes as if it were a successful conversion.

## Fix
Added a flag in _convert() that gets set when a specific-format converter accepts a file and fails. When that happens, the loop stops before any generic converter gets a chance to run. Normal conversions on valid files are not affected.

## Tests
Added test_invalid_ooxml_raises_exception which writes a plain-text file with each of the three extensions and checks that FileConversionException is raised with the right converter in the error. All existing tests still pass.


Aleeza-Bhatti:

Issue #1894, opened 2026-05-18
bug: IpynbConverter.accepts() raises UnicodeDecodeError on non-ASCII files (French PDFs, etc.)


echavet:
## Description

`IpynbConverter.accepts()` reads the raw file stream and decodes it to check if the file is a Jupyter notebook. When the decode fails (non-ASCII bytes in the file), the exception propagates uncaught and **crashes the entire conversion pipeline** — even though the file has nothing to do with Jupyter notebooks.

## Traceback (production environment)

```
File "markitdown\_markitdown.py", line 601, in _convert
    _accepts = converter.accepts(file_stream, stream_info, **_kwargs)
  File "markitdown\converters\_ipynb_converter.py", line 36, in accepts
    notebook_content = file_stream.read().decode(encoding)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 43916: ordinal not in range(128)
```

The affected file was a standard **French-language PDF invoice** containing UTF-8 encoded characters (`é`, `è`, `à`... → bytes starting with `0xc3`).

## Root Cause

In `accepts()`, when the MIME type matches `application/json`, the code decodes the full file stream content:

```python
# _ipynb_converter.py — current code, unsafe
encoding = stream_info.charset or "utf-8"  # was "ascii" in older versions
notebook_content = file_stream.read().decode(encoding)
```

If the file contains bytes that cannot be decoded (wrong charset, binary content), a `UnicodeDecodeError` propagates uncaught and kills the full conversion chain. The semantic intent of `accepts()` is simply to return `True` or `False` — it should **never raise**.

## Expected Behavior

`accepts()` should return `False` gracefully when the file cannot be decoded, instead of raising an exception.

## Proposed Fix

Wrap the decode block in a `try/except (UnicodeDecodeError, ValueError)`:

```python
for prefix in CANDIDATE_MIME_TYPE_PREFIXES:
    if mimetype.startswith(prefix):
        cur_pos = file_stream.tell()
        try:
            encoding = stream_info.charset or "utf-8"
            notebook_content = file_stream.read().decode(encoding)
            return (
                "nbformat" in notebook_content
                and "nbformat_minor" in notebook_content
            )
        except (UnicodeDecodeError, ValueError):
            # File contains non-decodable bytes — definitely not a notebook
            return False
        finally:
            file_stream.seek(cur_pos)
```

## Impact

[...]

PR #1851, opened 2026-04-30
feat: add progress_callback support for real-time conversion tracking

echavet:
## Summary

Adds a `progress_callback` mechanism to MarkItDown converters, enabling callers to receive real-time progress updates during document conversion.

## Problem Solved

When converting large documents (e.g. a 213-page PDF), `convert()` blocks for several minutes with no way to report progress to the user. Server-side applications, worker processes, and UIs cannot display meaningful feedback during long conversions.

## Design

Two new public types in `_base_converter.py`:

- **`ConversionProgress`** — a frozen `@dataclass` carrying `current`, `total`, `unit` (page/slide/chapter/sheet), and `source` (converter class name).
- **`ProgressCallback`** — a `@runtime_checkable` `Protocol` that any callable matching `(ConversionProgress) -> None` satisfies (structural subtyping / duck typing).

The callback is passed through the existing `**kwargs` chain (`convert()` → `_convert()` → `converter.convert()`). Converters that do not support progress simply ignore the kwarg.

## Converters Updated

| Converter | Unit | Loop |
|---|---|---|
| `PdfConverter` | `page` | per-page in pdfplumber loop |
| `PptxConverter` | `slide` | per-slide |
| `EpubConverter` | `chapter` | per-spine-item |
| `XlsxConverter` | `sheet` | per-sheet |
| `XlsConverter` | `sheet` | per-sheet |

## Usage Example

```python
from markitdown import MarkItDown, ConversionProgress

def on_progress(p: ConversionProgress) -> None:
    print(f"[{p.source}] {p.current}/{p.total} {p.unit}s")

md = MarkItDown()
result = md.convert("large_report.pdf", progress_callback=on_progress)
# Output:
# [PdfConverter] 1/213 pages
# [PdfConverter] 2/213 pages
# ...
```

## Backward Compatibility

- ✅ All existing functionality unchanged
- ✅ Callback is optional — omitting it preserves current behavior
- ✅ No new dependencies
- ✅ No changes to existing method signatures
- ✅ `graphrag-input` and other consumers work without modification

## Type of Change

- [x] New feature (non-breaking change)

## Testing

- All existing tests pass (no modifications needed)
- Manual verification with PDF, PPTX, and EPUB files
- `ConversionProgress` dataclass and `ProgressCallback` Protocol validated at runtime

PR #1895, opened 2026-05-18
fix: IpynbConverter.accepts() raises UnicodeDecodeError on non-ASCII files (e.g. French PDFs)

echavet:
## Summary

`IpynbConverter.accepts()` can raise `UnicodeDecodeError` when processing files with non-ASCII content (e.g. PDFs with French or other accented text). This propagates uncaught through the conversion pipeline and crashes the entire conversion — even though the file has nothing to do with Jupyter notebooks.

## Problem

The engine in `_markitdown.py` only protects `accepts()` calls against `NotImplementedError`:

```python
_accepts = False
try:
except NotImplementedError:
    pass
```

Any other exception raised by `accepts()` propagates up and crashes the caller. `IpynbConverter.accepts()` reads the file stream and decodes it to look for `nbformat` markers. If the file contains bytes that cannot be decoded (binary content, wrong charset), a `UnicodeDecodeError` is raised.

## Traceback (production environment — Windows Server)

```
File "markitdown\converters\_ipynb_converter.py", line 36, in accepts
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 43916: ordinal not in range(128)
```

The affected file was a standard French-language PDF invoice. The byte `0xc3` is the first byte of a multi-byte UTF-8 sequence representing accented characters (`é`, `è`, `à`, etc.).

## Root Cause

`accepts()` is a predicate — per the contract defined in `_base_converter.py`, it **must** return `bool` and **never raise**. The decode block was not protected against encoding failures:

```python
# Before — unsafe
notebook_content = file_stream.read().decode(encoding)  # 💥 raises on binary/wrong-charset files
return "nbformat" in notebook_content and "nbformat_minor" in notebook_content
```

Note: `stream_info.charset` can be explicitly set to `"ascii"` by MIME/charset detection for certain file types, making the issue reproducible even when `"utf-8"` is the fallback.

## Fix

Wrap the decode + check block in `try/except (UnicodeDecodeError, ValueError)` and return `False` on failure. A file that cannot be decoded is definitively not a Jupyter notebook.

```python
# After — safe
try:
    return (
        "nbformat" in notebook_content
        and "nbformat_minor" in notebook_content
    )
except (UnicodeDecodeError, ValueError):
    # File contains non-decodable bytes — definitely not a notebook
    return False
```

The `finally: file_stream.seek(cur_pos)` block is preserved, ensuring the stream position contract is always respected.

## Compatibility

- ✅ No behavior change for valid `.ipynb` files or JSON streams
- ✅ No new dependencies
- ✅ No changes to method signatures
- ✅ Respects the `accepts()` contract: returns `bool`, resets stream position

## Related

- Fixes issue: #1894
- This fix is independent of, and compatible with, PR #1851 (`progress_callback` feature)

## Type of Change

- [x] Bug fix (non-breaking change — adds error handling only)

### DIFF
--- a/packages/markitdown/src/markitdown/__init__.py
+++ b/packages/markitdown/src/markitdown/__init__.py
@@ -8,7 +8,12 @@ from ._markitdown import (
     PRIORITY_SPECIFIC_FILE_FORMAT,
     PRIORITY_GENERIC_FILE_FORMAT,
 )
-from ._base_converter import DocumentConverterResult, DocumentConverter
+from ._base_converter import (
+    DocumentConverterResult,
+    DocumentConverter,
+    ConversionProgress,
+    ProgressCallback,
+)
 from ._stream_info import StreamInfo
 from ._exceptions import (
     MarkItDownException,
@@ -23,6 +28,8 @@ __all__ = [
     "MarkItDown",
     "DocumentConverter",
     "DocumentConverterResult",
+    "ConversionProgress",
+    "ProgressCallback",
     "MarkItDownException",
     "MissingDependencyException",
     "FailedConversionAttempt",
--- a/packages/markitdown/src/markitdown/_base_converter.py
+++ b/packages/markitdown/src/markitdown/_base_converter.py
@@ -1,5 +1,45 @@
-from typing import Any, BinaryIO, Optional
+from dataclasses import dataclass
+from typing import Any, BinaryIO, Optional, Protocol, runtime_checkable
+
 from ._stream_info import StreamInfo
+
+
+@dataclass(frozen=True)
+class ConversionProgress:
+    """Progress update emitted by a converter during document processing.
+
+    Attributes:
+        current: Current unit number (1-indexed page, slide, chapter…).
+        total:   Total number of units (0 if unknown).
+        unit:    Semantic label for the unit being processed
+                 (``"page"``, ``"slide"``, ``"chapter"``, ``"sheet"``).
+        source:  Name of the converter class emitting the progress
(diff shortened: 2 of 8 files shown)


serejaris:
Duplicate triage note: I checked #1894 against current `main` and can reproduce the crash with a minimal non-PDF fixture:

```python
from io import BytesIO
from markitdown.converters._ipynb_converter import IpynbConverter
from markitdown._stream_info import StreamInfo

stream = BytesIO("Bonjour, Ça marche très bien.".encode("utf-8"))
info = StreamInfo(mimetype="application/json", extension=".bin", charset="ascii")
IpynbConverter().accepts(stream, info)
```

Current result:

```text
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 ...
```

I also compared the related PRs:

- this PR includes the `_ipynb_converter.py` guard for #1894;
- #1910 was a narrower closed PR with the same guard plus a regression test;
- this branch also includes unrelated `progress_callback` changes across several converters.

If maintainers prefer a minimal bugfix PR, the smallest scope seems to be only:

- `packages/markitdown/src/markitdown/converters/_ipynb_converter.py`
- `packages/markitdown/tests/test_module_misc.py`

with a regression using the `BytesIO(...), StreamInfo(mimetype="application/json", charset="ascii")` case above. I’m not opening another PR immediately to avoid adding a duplicate while this one is open.

afourney:
The notebook-probing failure from issue #1894 has now been resolved by the focused fix in #1929. I’m closing this as superseded; please reopen if that problem remains on current main.

PR #1899, opened 2026-05-21
Docx And PDF Ocr support parallel mode by using max_workers args

jiejieling:
Docx And PDF Ocr support parallel mode by using max_workers args.
When max_workers > 1  use parallel ocr


jiejieling:

PR #1903, opened 2026-05-22
feat: extract embedded PPTX images to a directory

### LINKED ISSUE #56: PPTX: Extract images
Currently, the script only extracts placeholders, i.e.,

```md
![Content Placeholder 16](ContentPlaceholder16.jpg)
```

It would be nice if the tool would export the images (to the current folder or a folder passed as argument).
Thus, one could see the images in the markdown preview as well.

Image

Ylsssq926:
## What

Adds an `extract_images_to` option (Python API) and a corresponding `--extract-images-to` CLI flag to the PPTX converter. When set, embedded PPTX images are written to the supplied directory (deduplicated by content SHA1) and the markdown output references them via `<extract_images_to>/<filename>`, instead of emitting placeholder filenames that point at non-existent files.

Closes #56

## Why

Today the PPTX converter emits markdown like

    alt

but `Picture4.jpg` is never written to disk, so the link is dead. Issue #56 asked for the images to be exported so that markdown preview tools can actually render them. This change makes that opt-in via `extract_images_to`, while keeping the default behaviour completely unchanged for backward compatibility.

## How

- New `extract_images_to: Optional[str]` kwarg on the PPTX converter; when set, images are written to that directory.
- New `--extract-images-to DIR` CLI flag in the same style as the existing `--keep-data-uris`.
- Filenames follow `slide{NN}_{shape_idx}_{sanitized_name}.{ext}` with Unicode-safe sanitisation so non-ASCII shape names don't collapse to a single character.
- Identical images (matched by `image.sha1`) are written once and referenced by all slides that embed them.
- Disk-side filename collisions (e.g. when reusing the same output directory across runs) get a `_2`, `_3` suffix; existing files on disk are never silently overwritten.
- Markdown links use `posixpath.join(extract_images_to, filename)` with forward slashes, so the output is safe on Windows too.
- Extension detection is layered: `image.ext` -> `image.content_type` -> `mimetypes.guess_extension` -> `image.filename` -> `bin`, each step wrapped in `try/except` because `python-pptx` raises `ValueError` on SVG/EMF/ICO when resolving `ext`.
- If a write fails for any reason we fall back to the legacy placeholder behaviour and emit a one-time `warnings.warn` rather than crashing the whole conversion.
- When both `keep_data_uris=True` and `extract_images_to` are passed, `keep_data_uris` wins (base64 inline) and no files are written.

## Backward compatibility

The default code path is untouched: if neither flag is set, the converter still emits the same `Picture4.jpg`-style placeholder that the existing `_test_vectors.py` regression tests assert on. All seven existing PPTX vector tests continue to pass.

## Tests

Added in `packages/markitdown/tests/test_module_misc.py`:

- `test_pptx_extract_images_to_dir` - basic write-and-reference.
- `test_pptx_extract_images_to_relative_dir` - verifies relative paths in the markdown link, no absolute-path leak.
- `test_pptx_extract_images_does_not_overwrite_existing` - asserts pre-existing files in the target directory get the `_2` suffix instead of being clobbered.
- `test_pptx_extract_images_dedup` - programmatically generates a PPTX with the same image embedded three times and asserts exactly one file is written and the markdown references it multiple times.
- `test_pptx_extract_images_keep_data_uris_takes_precedence` - verifies the precedence rule.

Added in `packages/markitdown/tests/test_cli_misc.py`:

- `test_extract_images_to` - CLI smoke test via subprocess.

Local results: all 6 new tests pass and the 11 existing PPTX vector tests (`test_*[test_vector3]`) still pass, confirming no regression in the default placeholder behaviour. `black` passes on the touched files.

## Notes

- README is untouched in this PR. Happy to send a follow-up doc PR once the API shape is approved here, since the wording depends on whether you'd like the markdown link to be the user-supplied path verbatim (current behaviour) or something more clever (e.g. relative to the markdown output file). I went with the simplest option that matches issue #56's framing.
- Unsupported image types (SVG/EMF/WMF) are still written to disk with their real extension, but downstream markdown renderers may not display them. That seems strictly better than the status quo of writing nothing at all; happy to add a doc note if you'd like.

### DIFF
--- a/packages/markitdown/src/markitdown/__main__.py
+++ b/packages/markitdown/src/markitdown/__main__.py
@@ -110,6 +110,16 @@ def main():
         help="Keep data URIs (like base64-encoded images) in the output. By default, data URIs are truncated.",
     )
 
+    parser.add_argument(
+        "--extract-images-to",
+        metavar="DIR",
+        default=None,
+        help=(
+            "Extract embedded images (currently supported for PPTX) to the given directory and "
+            "reference them by relative filename in the output. Ignored when --keep-data-uris is set."
+        ),
+    )
+
     parser.add_argument("filename", nargs="?")
     args = parser.parse_args()
 
@@ -191,10 +201,14 @@ def main():
             sys.stdin.buffer,
             stream_info=stream_info,
             keep_data_uris=args.keep_data_uris,
+            extract_images_to=args.extract_images_to,
         )
     else:
         result = markitdown.convert(
-            args.filename, stream_info=stream_info, keep_data_uris=args.keep_data_uris
+            args.filename,
+            stream_info=stream_info,
+            keep_data_uris=args.keep_data_uris,
+            extract_images_to=args.extract_images_to,
         )
 
     _handle_output(args, result)
--- a/packages/markitdown/src/markitdown/converters/_pptx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_pptx_converter.py
@@ -5,6 +5,7 @@ import os
 import io
 import re
 import html
+import posixpath
 import warnings
 
 from typing import BinaryIO, Any, Dict, Optional, Set, Tuple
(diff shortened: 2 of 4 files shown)


Ylsssq926:

PR #1906, opened 2026-05-22
fix: let Document Intelligence use SDK default API version

### LINKED ISSUE #1904: DocumentIntelligenceConverter should not set a default api_version

he-yufeng:
Fixes #1904.

## Summary
- stop forcing the older Document Intelligence preview API version by default
- only pass api_version to DocumentIntelligenceClient when the caller supplies one
- add focused constructor tests that do not call Azure

## Verified
- .venv\Scripts\python.exe -m pytest packages\markitdown\tests\test_docintel_html.py -q
- .venv\Scripts\python.exe -m py_compile packages\markitdown\src\markitdown\converters\_doc_intel_converter.py packages\markitdown\tests\test_docintel_html.py
- git diff --check

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_doc_intel_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_doc_intel_converter.py
@@ -134,7 +134,7 @@ class DocumentIntelligenceConverter(DocumentConverter):
         self,
         *,
         endpoint: str,
-        api_version: str = "2024-07-31-preview",
+        api_version: str | None = None,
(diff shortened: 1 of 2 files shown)


he-yufeng:
Still relevant — pinning the Document Intelligence API version breaks when the SDK moves; letting it use the SDK default avoids that. Small change.

afourney:
The SDK-default API-version behavior proposed here has now landed through #2267. Thanks for the contribution. Please reopen if the original failure still occurs on current main.

PR #1909, opened 2026-05-23
fix: repair docx zip header casing

### LINKED ISSUE #1812: BadZipFile crash on .docx files with case-mismatched zip entry names

he-yufeng:
﻿Fixes #1812.

Some DOCX producers write ZIP entries whose central-directory name differs from the local file header only by path casing. Python's zipfile rejects those files during extraction even though Word and common zip tools tolerate them.

This patch repairs that narrow case in memory before the existing DOCX pre-processing step reads each member. The central-directory name remains authoritative, and the patch only applies when the local and central names have the same byte length and differ only by case.

Validation:
- .venv\Scripts\python.exe -m pytest packages\markitdown\tests\test_module_misc.py::test_docx_preprocess_repairs_case_mismatched_zip_names -q
- .venv\Scripts\python.exe -m pytest packages\markitdown\tests\test_module_misc.py::test_docx_comments packages\markitdown\tests\test_module_misc.py::test_docx_equations packages\markitdown\tests\test_module_misc.py::test_docx_preprocess_repairs_case_mismatched_zip_names -q
- .venv\Scripts\python.exe -m py_compile packages\markitdown\src\markitdown\converter_utils\docx\pre_process.py packages\markitdown\tests\test_module_misc.py
- .venv\Scripts\black.exe --check packages\markitdown\src\markitdown\converter_utils\docx\pre_process.py packages\markitdown\tests\test_module_misc.py
- git diff --check

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -1,3 +1,4 @@
+import struct
 import zipfile
 from io import BytesIO
 from typing import BinaryIO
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this older version in favor of #1925.

Both PRs target the same DOCX local-header casing bug from #1812. #1925 is the cleaner branch: narrower helper name, dedicated `test_docx_pre_process.py`, and a stricter ASCII case-only repair boundary. Keeping both open just splits reviewer attention.

PR #1914, opened 2026-05-26
Bump version to 0.1.6

afourney:


packages/markitdown/src/markitdown/__about__.py
```diff
@@ -1,4 +1,4 @@
 # SPDX-FileCopyrightText: 2024-present Adam Fourney <adamfo@microsoft.com>
 #
 # SPDX-License-Identifier: MIT
```
koyulin24747-arch, line 3: How to Remove the MIT lincence back to read.me

koyulin24747-arch (commented):
Thank you for your response and keep my been stolen MIT lincence back to the mean person did stolen from me and use it a bad way to use it

Pjrich1313 (commented):
Merge load
Diff

koyulin24747-arch (commented):
Here is Yu Lin Ko

I am still stoked about this thread from third party .

I want to make sure that Eula want the MIT lincence under @Read.Me

Or other author would take it charge well but promise me and MIT not been used at the same murder to me!

packages/markitdown/src/markitdown/__about__.py
koyulin24747-arch: After all,you people are still here and add the problem,so that I have to do something!

ipamjsp:
How does it work, please

Pjrich1313 (commented):
Merge leave MIT my Copilot is full train and capable of obtaining a license and I'm a full time student show me where it says anything about this issue

PR #1920, opened 2026-05-29
Fix scanned PDF OCR fallback when only page headers are extracted

### LINKED ISSUE #1863: markitdown-ocr process  non-text layer PDFs generate an MD document containing only page numbers
markitdown-ocr  process non-text layer PDFs, such as those converted from images or generated through scanning,It will generate an MD document containing only page numbers
the reason is , In the file "pdf_converter_with_ocr.py",first， by  "markdown_content.append(f"\n"## Page{page_num}\n)"  set a page Message,
but In the subsequent code logic,  whether to perform full-page OCR based on whether the content is empty。

Due to the page number information in markdown_content, markdown_content is not empty, so it will skip OCR，finally，generate an MD document containing only page numbers

dcaayushd:
Fixes #1863

Fixes scanned/image-only PDF conversion in markitdown-ocr where generated page headers made the output look non-empty, causing full-page OCR fallback to be skipped. This could produce Markdown containing only ## Page N headings.

Changes:

Track whether real text or OCR content was extracted separately from page headers.
Run pdfminer and full-page OCR fallback when page headers are the only content.
Add a regression test for the header-only extraction path.
Update the multipage PDF test expectation for current pdfplumber behavior.

### DIFF
--- a/packages/markitdown-ocr/src/markitdown_ocr/_pdf_converter_with_ocr.py
+++ b/packages/markitdown-ocr/src/markitdown_ocr/_pdf_converter_with_ocr.py
@@ -182,6 +182,7 @@ class PdfConverterWithOCR(DocumentConverter):
         pdf_bytes = io.BytesIO(file_stream.read())
 
         markdown_content = []
+        has_extracted_content = False
 
         try:
             with pdfplumber.open(pdf_bytes) as pdf:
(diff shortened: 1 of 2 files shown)


dcaayushd:
@afourney @gagb Could you please review this PR and approve the workflow when you have a chance? Thanks!

PR #1925, opened 2026-05-29
fix: repair docx local header casing

### LINKED ISSUE #1812: BadZipFile crash on .docx files with case-mismatched zip entry names

he-yufeng:
## Summary

Fixes #1812.

Some DOCX files have a zip central directory entry such as `customXml/item2.xml` while the corresponding local file header uses a different casing such as `customXML/item2.xml`. Python's `zipfile` accepts the central directory, but raises `BadZipFile` when the entry is read.

This patches only case-only ASCII local header mismatches before the DOCX preprocessing step reads the archive. The central directory name stays authoritative, and non-case-only mismatches are left untouched.

## Validation

```bash
$env:PYTHONPATH='C:\dev\GITHUB-clean\markitdown-1812\packages\markitdown\src'; C:\dev\GITHUB-clean\markitdown\.venv\Scripts\python.exe -m pytest packages\markitdown\tests\test_docx_pre_process.py -q
$env:PYTHONPATH='C:\dev\GITHUB-clean\markitdown-1812\packages\markitdown\src'; C:\dev\GITHUB-clean\markitdown\.venv\Scripts\python.exe -m pytest packages\markitdown\tests\test_module_misc.py::test_docx_equations packages\markitdown\tests\test_docx_pre_process.py -q
C:\dev\GITHUB-clean\markitdown\.venv\Scripts\python.exe -m black --check packages\markitdown\src\markitdown\converter_utils\docx\pre_process.py packages\markitdown\tests\test_docx_pre_process.py
$env:PYTHONPATH='C:\dev\GITHUB-clean\markitdown-1812\packages\markitdown\src'; C:\dev\GITHUB-clean\markitdown\.venv\Scripts\python.exe -m mypy --ignore-missing-imports packages\markitdown\src\markitdown\converter_utils\docx\pre_process.py packages\markitdown\tests\test_docx_pre_process.py
git diff --check
```

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -1,5 +1,6 @@
 import zipfile
 from io import BytesIO
+from struct import unpack_from
 from typing import BinaryIO
 from xml.etree import ElementTree as ET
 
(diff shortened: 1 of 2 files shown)


he-yufeng:
Friendly ping. DOCX files whose zip central directory entry and local file header disagree on casing (e.g. `customXml/item2.xml` vs a differently cased local name) currently fail to convert. This makes the lookup tolerant of the mismatch, with a regression test covering it. Would love a review when you have a moment.

afourney:

PR #1936, opened 2026-05-30
fix: DocumentIntelligenceConverter should not set a default api_version (fixes #1904)

### LINKED ISSUE #1904: DocumentIntelligenceConverter should not set a default api_version

hanhan761:
## Summary

`DocumentIntelligenceConverter.__init__` hardcodes `api_version: str = "2024-07-31-preview"`. Azure Document Intelligence's SDK default version has moved forward, and passing an outdated preview version causes `404 Resource Not Found` for users who don't explicitly set `api_version`.

This fix changes the default to `None` and only passes `api_version` to `DocumentIntelligenceClient` when it is explicitly set, letting the Azure SDK use its own default.

## Issue

Fixes #1904

## Verification

- Default `api_version=None` → `DocumentIntelligenceClient` receives no `api_version` kwarg → SDK uses its default
- Explicit `api_version="2024-07-31-preview"` → passed through unchanged

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_doc_intel_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_doc_intel_converter.py
@@ -134,7 +134,7 @@ class DocumentIntelligenceConverter(DocumentConverter):
         self,
         *,
         endpoint: str,
-        api_version: str = "2024-07-31-preview",
+        api_version: str | None = None,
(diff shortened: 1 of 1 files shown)


afourney:
Thanks for working on this. #2267 has now removed MarkItDown’s default API-version pin and lets the Azure SDK choose its default, resolving this issue. Please reopen if the problem remains on current main.

PR #1950, opened 2026-05-30
fix: prevent UnboundLocalError in RSS converter when channel title is missing (fixes #1946)

### LINKED ISSUE #1946: bug: RSS converter crashes with UnboundLocalError when channel title is missing
## Description

## Root Cause

## Reproduction

## Expected Behavior

## Proposed Fix

hanhan761:
## Summary

Fixes two bugs in the RSS/Atom converter:

1. **UnboundLocalError** in _parse_rss_type(): When an RSS feed has no <title> element, md_text was never initialized but subsequent += operations referenced it. Added md_text = "" initialization before the conditional block.

2. **# None display bug** in _parse_atom_type(): When an Atom feed has no <title> element, line 110 rendered # None as the document heading. Changed to "# {title}\n" if title else "".

## Changes

- packages/markitdown/src/markitdown/converters/_rss_converter.py:
  - Line 110: Guard Atom title assignment against None
  - Line 146: Initialize md_text = "" before conditional blocks in RSS parser

## Issue

Fixes #1946

## Verification

- Code inspection confirms both fixes handle the None/missing title case correctly
- Existing code paths for valid RSS/Atom feeds are unaffected (conditionals remain the same)

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_rss_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_rss_converter.py
@@ -107,7 +107,7 @@ class RssConverter(DocumentConverter):
         title = self._get_data_by_tag_name(root, "title")
         subtitle = self._get_data_by_tag_name(root, "subtitle")
         entries = root.getElementsByTagName("entry")
-        md_text = f"# {title}\n"
+        md_text = f"# {title}\n" if title else ""
         if subtitle:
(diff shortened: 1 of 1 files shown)


afourney:
The reported RSS crash when the channel title is absent has now been resolved by #2164. I’m closing this as superseded; please reopen if your original input still fails on current main.

PR #1957, opened 2026-05-30
chore: remove unused mammoth import from PlainTextConverter

### LINKED ISSUE #1951: PlainTextConverter has unused mammoth import (copy-paste from DocxConverter)
The `PlainTextConverter` module has an unused `mammoth` import that was copied from `DocxConverter`. This is dead code and misleading — it suggests PlainTextConverter has a dependency on mammoth when it doesn't. The import also stores `_dependency_exc_info` on import failure, but never checks it, creating an unused variable that serves no purpose.

## Fix

Remove the mammoth import block and the associated `_dependency_exc_info` variable from `_plain_text_converter.py`.

hanhan761:
Fixes #1951

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_plain_text_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_plain_text_converter.py
@@ -1,18 +1,7 @@
-import sys
-
 from typing import BinaryIO, Any
 from charset_normalizer import from_bytes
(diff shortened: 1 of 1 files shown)


afourney:
Thanks for the cleanup. The same erroneous mammoth dependency handling is removed by #1953; please reopen if #1953 does not fully address the problem.

Issue #1960, opened 2026-05-30
exiftool_metadata() crashes with FileNotFoundError when exiftool_path binary doesn't exist


hanhan761:
## Description

When `exiftool_path` is set but points to a non-existent binary, `exiftool_metadata()` crashes with an unhandled `FileNotFoundError` instead of returning an empty metadata dict or raising a clear error.

The `subprocess.run()` call on line 22 (and 41) raises `FileNotFoundError` when the binary is not found, but the `except` clause only catches `CalledProcessError` and `ValueError`.

## Root Cause

`packages/markitdown/src/markitdown/converters/_exiftool.py`, lines 21-27:
```python
try:
    version_output = subprocess.run(
        [exiftool_path, "-ver"],
        capture_output=True, text=True, check=True,
    ).stdout.strip()
    ...
except (subprocess.CalledProcessError, ValueError) as e:
    raise RuntimeError("Failed to verify ExifTool version.") from e
```

`FileNotFoundError` (when the binary doesn't exist) is a subclass of `OSError`, not `CalledProcessError` or `ValueError`, so it propagates uncaught.

## Expected Behavior

If `exiftool_path` points to a non-existent file, `exiftool_metadata()` should return an empty dict (graceful degradation) rather than crashing the entire conversion.

## Proposed Fix

Add `OSError` (parent of `FileNotFoundError`) to the except clause.

PR #1963, opened 2026-05-30
fix: handle FileNotFoundError gracefully in exiftool_metadata() (#1960)

hanhan761:
## Description

## Root Cause

```python
try:
        [exiftool_path, "-ver"],
    ).stdout.strip()
    ...
```

## Expected Behavior

## Proposed Fix


afourney:

PR #1965, opened 2026-05-30
fix: guard against None text_frame in PptxConverter chart title (fixes #1958)

### LINKED ISSUE #1958: bug: PptxConverter crashes when chart title has no text_frame (AttributeError)
## Description

PptxConverter crashes with AttributeError when a chart title has no text frame. According to python-pptx docs, chart_title.text_frame can be None if the title was set via the text property rather than through the rich text frame API.

## Root Cause

In _pptx_converter.py line 239:
chart.chart_title.text_frame returns None in some cases (e.g., charts where the title was set programmatically), causing AttributeError: NoneType object has no attribute text.

## Expected Behavior

Charts without a text frame on the title should be handled gracefully - the title should be skipped rather than crashing the entire conversion.

## Proposed Fix

Add a None check: if chart.has_title and chart.chart_title.text_frame is not None

hanhan761:
PptxConverter crashes with AttributeError when chart.chart_title.text_frame is None. Added None check.

### DIFF
--- a/packages/markitdown/src/markitdown/_uri_utils.py
+++ b/packages/markitdown/src/markitdown/_uri_utils.py
@@ -2,7 +2,7 @@ import base64
 import os
 from typing import Tuple, Dict
 from urllib.request import url2pathname
-from urllib.parse import urlparse, unquote_to_bytes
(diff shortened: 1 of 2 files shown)


afourney:

PR #1967, opened 2026-05-30
fix: handle case-mismatched zip entry names in DOCX pre-processing

### LINKED ISSUE #1812: BadZipFile crash on .docx files with case-mismatched zip entry names

hanhan761:
## Summary

Add `_fix_zip_name_casing()` to `pre_process.py` that patches local file header names to match the central directory before opening the zip. This prevents `BadZipFile` crashes on `.docx` files where zip entry names differ in casing between local headers and central directory (common with legal document systems and older Word versions).

## Issue

Fixes #1812

## Verification

- Existing docx tests (`test_docx_comments`, `test_docx_equations`) pass
- `_fix_zip_name_casing()` runs as the first step in `pre_process_docx` with no side effects on normal files

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -1,3 +1,4 @@
+import struct
 import zipfile
 from io import BytesIO
 from typing import BinaryIO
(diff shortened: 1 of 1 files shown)


afourney:

PR #1970, opened 2026-05-30
fix: remove unused mammoth import from PlainTextConverter

hanhan761:
## Summary\n\nRemove dead \mammoth\ import and unused \_dependency_exc_info\ variable from \PlainTextConverter\. This was copy-pasted from \DocxConverter\ and serves no purpose.\n\n## Issue\n\nFixes #1951\n\n## Verification\n\n- 4 PlainText-related tests pass\n- Syntax verified


afourney:
Thanks for the cleanup. The same erroneous mammoth dependency handling is removed by #1953; please reopen if #1953 does not fully address the problem.

PR #1971, opened 2026-05-30
fix: avoid hash None heading in WikipediaConverter when title is missing (fixes #1968)

### LINKED ISSUE #1968: bug: WikipediaConverter renders hash None heading when page has no title
## Description

WikipediaConverter renders `# None` as the document heading when a Wikipedia page has no title element or the title element is empty/whitespace-only. This happens in `_wikipedia_converter.py` line 78.

## Root Cause

In `_wikipedia_converter.py` line 78:
```python
webpage_text = f"# {main_title}\n\n" + _CustomMarkdownify(**kwargs).convert_soup(body_elm)
```
`main_title` can be None (set at lines 70-75 when `soup.title` is None or `title_elm.string` is None). When None, Python renders it as the string "None" in the f-string.

## Expected Behavior

When no title is available, the document should start directly with the body content, without a `# None` heading.

hanhan761:
## Summary

Fixes a bug where WikipediaConverter renders # None as the document heading when the page has no title element or the title element is empty.

## Changes

- _wikipedia_converter.py: Changed from unconditional "# {main_title}\n\n" to a guarded assignment that only renders the heading when main_title is truthy.

## Issue

Fixes #1968

## Verification

- Code inspection confirms # None no longer appears when title is missing
- Existing behavior for pages with valid titles is unchanged

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_wikipedia_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_wikipedia_converter.py
@@ -75,7 +75,8 @@ class WikipediaConverter(DocumentConverter):
                 main_title = title_elm.string
 
             # Convert the page
-            webpage_text = f"# {main_title}\n\n" + _CustomMarkdownify(
+            heading = f"# {main_title}\n\n" if main_title else ""
+            webpage_text = heading + _CustomMarkdownify(
                 **kwargs
             ).convert_soup(body_elm)
         else:


afourney:
Thanks for the contribution. The missing Wikipedia title case is covered with regression testing by #1990; please reopen if #1990 does not address your input.

PR #1973, opened 2026-05-30
fix: decode exiftool JSON output as UTF-8 instead of locale encoding

### LINKED ISSUE #1972: fix: exiftool JSON output decoded with locale encoding instead of UTF-8

**Reproduction:**

hanhan761:
## Summary

Fix exiftool_metadata() to decode exiftool's JSON output as UTF-8 instead of using the system locale encoding.

## Issue

Fixes #1972

## Background

ExifTool always outputs JSON in UTF-8, but the code was decoding it with locale.getpreferredencoding(). On systems where the locale encoding is not UTF-8 (e.g., Windows with Chinese locale), this causes UnicodeDecodeError. JSON is required to be UTF-8 per RFC 8259.

## Changes

- Removed unused locale import
- Changed to output.decode("utf-8")

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_doc_intel_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_doc_intel_converter.py
@@ -146,6 +146,7 @@ class DocumentIntelligenceConverter(DocumentConverter):
             DocumentIntelligenceFileType.BMP,
             DocumentIntelligenceFileType.TIFF,
         ],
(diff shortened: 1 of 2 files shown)


afourney:
Thanks for the fix. A similar ExifTool UTF-8 decoding correction is proposed in #2067; please reopen if #2067 does not address the original problem.  Also this PR includes changes to doc intel, not documented here.

PR #1978, opened 2026-05-30
fix: initialize md_text before conditional in RSS converter

### LINKED ISSUE #1946: bug: RSS converter crashes with UnboundLocalError when channel title is missing

hanhan761:
## Summary

Initialize `md_text = ""` before the `if channel_title:` block so that RSS feeds without a `<title>` element don't crash with `UnboundLocalError`.

## Issue

Fixes #1946

## Verification

- RSS feed without `<title>` converts successfully (previously crashed)
- RSS feed with `<title>` still works correctly


afourney:
The uninitialized md_text failure for titleless RSS channels has now been fixed by #2164. I’m closing this as superseded; please reopen if the problem remains on current main.

PR #1886, opened 2026-05-15
fix(pptx): ignore empty llm captions

Lubrsy706:
## Summary
- Skip empty/None LLM image captions when building PPTX image alt text
- Preserve embedded deck alt text and shape-name fallback behavior
- Add regression coverage for an LLM caption provider returning `None`

Fixes #1534

## Tests
- PYTHONPATH=packages/markitdown/src /Users/bytedance/MySpace/open-source/markitdown-excel-currency/.venv/bin/python -m pytest packages/markitdown/tests/test_module_misc.py::test_pptx_converter_treats_none_llm_caption_as_empty
- PYTHONPATH=packages/markitdown/src /Users/bytedance/MySpace/open-source/markitdown-excel-currency/.venv/bin/python -m compileall -q packages/markitdown/src/markitdown/converters/_pptx_converter.py packages/markitdown/tests/test_module_misc.py
- git diff --check

PR #1981, opened 2026-05-30
fix: guard against None llm_description in PptxConverter image alt text (fixes #1534)

### LINKED ISSUE #1534: A bug when llm_description got None response from LLM in PptxConverter

hanhan761:
PptxConverter crashes with TypeError when llm_caption() returns None. Fix: add or empty string guard before join().


afourney:

PR #1983, opened 2026-05-30
fix: DOCX math converter crashes when oMath element is missing

### LINKED ISSUE #1979: bug: DOCX math converter crashes when oMath element is missing in malformed equations
## Description

The DOCX pre-processor crashes with `TypeError` or `AttributeError` when an equation element has no valid `oMath` child. In `_convert_omath_to_latex()`, if `math_root.find(OMML_NS + "oMath")` returns `None` (malformed OMML), the code passes `None` to `oMath2Latex()`, which crashes.

## Root Cause

In `pre_process.py` lines 46-48:
```python
math_element = math_root.find(OMML_NS + "oMath")
latex = oMath2Latex(math_element).latex
```

When `math_element` is None (malformed DOCX math), `oMath2Latex(None)` raises `TypeError` in its constructor.

## Impact

Any DOCX file with a corrupted or incomplete math equation will fail the entire conversion with an unhandled exception.

## Proposed Fix

Add a None check after `math_root.find()` and return a fallback representation (e.g., the raw tag text or an empty string) instead of crashing.

hanhan761:
Fixes #1979

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -44,6 +44,8 @@ def _convert_omath_to_latex(tag: Tag) -> str:
     math_root = ET.fromstring(MATH_ROOT_TEMPLATE.format(str(tag)))
     # Find the 'oMath' element within the XML document
     math_element = math_root.find(OMML_NS + "oMath")
+    if math_element is None:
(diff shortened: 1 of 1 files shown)


afourney:
Thanks for the fix. The malformed equation case with a missing oMath child is covered, with regression tests, by #1995. If #1995 does not handle this issue, please re-open.

PR #1984, opened 2026-05-30
fix: handle None title in WikipediaConverter

### LINKED ISSUE #1968: bug: WikipediaConverter renders hash None heading when page has no title

hanhan761:
## Summary

Guard against `main_title` being `None` when building the Wikipedia page heading, so pages without a title don't render `# None`.

## Issue

Fixes #1968

## Verification

- Wikipedia page without title: no `# None` heading
- Wikipedia page with title: heading still shown

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_wikipedia_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_wikipedia_converter.py
@@ -75,7 +75,8 @@ class WikipediaConverter(DocumentConverter):
                 main_title = title_elm.string
 
             # Convert the page
-            webpage_text = f"# {main_title}\n\n" + _CustomMarkdownify(
+            heading = f"# {main_title}\n\n" if main_title else ""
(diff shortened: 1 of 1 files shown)


afourney:
Thanks for the contribution. The missing Wikipedia title case is covered with regression testing by #1990; please reopen if #1990 does not address your input.

PR #1986, opened 2026-05-30
fix: handle FileNotFoundError gracefully in exiftool_metadata() when binary doesn't exist (fixes #1960)

### LINKED ISSUE #1960: exiftool_metadata() crashes with FileNotFoundError when exiftool_path binary doesn't exist

hanhan761:
## Summary
When \exiftool_path\ points to a non-existent binary, \subprocess.run()\ raises \FileNotFoundError\, which was not caught by the existing \except\ clause (only \CalledProcessError\ and \ValueError\). This caused the conversion to crash instead of gracefully degrading.

This fix catches \FileNotFoundError\ separately and returns an empty metadata dict, matching the graceful degradation pattern already used when \exiftool_path\ is \None\.

## Issue
Fixes #1960

## Verification
- \exiftool_path=None\ returns \{}\ (unchanged)
- \exiftool_path='/nonexistent/path'\ returns \{}\ (fixed)
- Invalid/broken exiftool binary still raises \RuntimeError\ (unchanged)

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -44,6 +44,9 @@ def _convert_omath_to_latex(tag: Tag) -> str:
     math_root = ET.fromstring(MATH_ROOT_TEMPLATE.format(str(tag)))
     # Find the 'oMath' element within the XML document
     math_element = math_root.find(OMML_NS + "oMath")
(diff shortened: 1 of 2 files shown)


afourney:

PR #1988, opened 2026-05-30
fix: fall back to operatorname for unknown math functions instead of crashing (fixes #1982)

### LINKED ISSUE #1982: bug: DOCX math converter crashes with NotImplementedError on unknown functions
## Description

## Root Cause

## Proposed Fix

hanhan761:
## Summary

Fixes NotImplementedError crash when the DOCX math converter encounters a function name not in the hardcoded FUNC dictionary (e.g., log, ln, exp, det, gcd, lcm).

## Root Cause

In omml.py:275, do_fname() raises NotImplementedError for any function not in the FUNC dict, which only contains 13 basic trig functions. Common functions like log, ln, exp, etc. cause the entire DOCX conversion to crash.

## Changes

- omml.py: Changed from raising NotImplementedError to falling back with \operatorname{name}(expr) LaTeX rendering

## Issue

Fixes #1982

## Verification

- Unknown functions now render as \operatorname{name} in LaTeX output
- Known functions (sin, cos, etc.) continue to use their dedicated commands

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
@@ -272,7 +272,7 @@ class oMath2Latex(Tag2Method):
                 if FUNC.get(t):
                     latex_chars.append(FUNC[t])
                 else:
-                    raise NotImplementedError("Not support func %s" % t)
+                    latex_chars.append(f"\\operatorname{{{t}}}({FUNC_PLACE})")
             else:
                 latex_chars.append(t)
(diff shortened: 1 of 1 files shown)


afourney:
Thank you for proposing the fallback for unknown math functions. [#2268](https://github.com/microsoft/markitdown/pull/2268) implements the \operatorname fallback instead of raising NotImplementedError, and also handles function names split across multiple runs.

I'm closing this as already addressed by that merged implementation. The fix is available on main and in v0.1.8b1; v0.1.7 does not include it.

PR #1989, opened 2026-05-30
fix: handle RSS feeds without channel titles

### LINKED ISSUE #1946: bug: RSS converter crashes with UnboundLocalError when channel title is missing

he-yufeng:
﻿Fixes #1946.

## Summary
- initialize RSS markdown output before optional channel title handling
- keep channel descriptions and items when the channel title is missing
- add a regression test for untitled RSS feeds

## Verification
- `.\.venv\Scripts\python.exe -m pytest packages\markitdown\tests\test_module_misc.py::test_rss_without_channel_title -q`
- `.\.venv\Scripts\python.exe -m py_compile packages\markitdown\src\markitdown\converters\_rss_converter.py packages\markitdown\tests\test_module_misc.py`
- `git diff --check`


he-yufeng:
Closing this because existing PRs #1950 and #1978 already cover the same #1946 RSS no-title crash. I missed those in the first duplicate search.

PR #1992, opened 2026-05-30
fix: render unknown math functions as \\operatorname instead of crashing (#1982)

### LINKED ISSUE #1982: bug: DOCX math converter crashes with NotImplementedError on unknown functions

hanhan761:
## Summary

Fix #1982: When the OMML-to-LaTeX converter encounters a math function not in the hardcoded FUNC dictionary (e.g., log, ln, exp, det, gcd, lcm), it now renders it as `\operatorname{func_name}` instead of raising NotImplementedError and crashing the entire DOCX conversion.

## Changes

In `packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py`:
- Replaced `raise NotImplementedError("Not support func %s" % t)` with a fallback that renders the unknown function name as `\operatorname{func_name}{argument}`

## Issue

Fixes #1982

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
@@ -272,7 +272,7 @@ class oMath2Latex(Tag2Method):
                 if FUNC.get(t):
                     latex_chars.append(FUNC[t])
                 else:
-                    raise NotImplementedError("Not support func %s" % t)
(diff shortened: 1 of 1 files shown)


afourney:
Thanks for working on #1982. The same unknown-function branch was fixed in [#2268](https://github.com/microsoft/markitdown/pull/2268), which emits an \operatorname expression and additionally combines split function-name runs.

That implementation supersedes this change, so I'm closing the PR as already addressed. It is included in main and the v0.1.8b1 prerelease, rather than v0.1.7.

PR #1994, opened 2026-05-30
fix: decode exiftool JSON output as UTF-8 instead of locale encoding

### LINKED ISSUE #1972: fix: exiftool JSON output decoded with locale encoding instead of UTF-8

hanhan761:
## Summary

ExifTool always outputs JSON in UTF-8 (per RFC 8259), but the exiftool_metadata() function decoded the output using locale.getpreferredencoding(). On systems with non-UTF-8 locale encoding (e.g. Windows with Chinese locale cp936), this caused UnicodeDecodeError when ExifTool returned non-ASCII metadata.

## Changes

- **_exiftool.py**: Replace text=True with encoding=utf-8 in the version-check subprocess call, so stdout is decoded as UTF-8.
- **_exiftool.py**: Decode JSON output bytes with utf-8 instead of locale.getpreferredencoding().
- **_exiftool.py**: Remove unused import locale.
- **test_module_misc.py**: Add test_exiftool_metadata_decodes_utf8 - a mocked unit test that verifies non-ASCII characters are correctly decoded from UTF-8.

## Verification

- python -m pytest tests/test_module_misc.py::test_exiftool_metadata_decodes_utf8 -v - PASSED
- python -m pytest tests/test_module_vectors.py -v -x - all 109 PASSED

## Backward Compatibility

- No behavior change for UTF-8 locales (the common case)
- Fixes decoding on non-UTF-8 locales (e.g. Windows cp936)
- No new dependencies

Fixes #1972

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_exiftool.py
+++ b/packages/markitdown/src/markitdown/converters/_exiftool.py
@@ -1,5 +1,4 @@
 import json
-import locale
 import subprocess
 from typing import Any, BinaryIO, Union
 
@@ -22,7 +21,7 @@ def exiftool_metadata(
         version_output = subprocess.run(
             [exiftool_path, "-ver"],
             capture_output=True,
-            text=True,
+            encoding="utf-8",
             check=True,
         ).stdout.strip()
         version = _parse_version(version_output)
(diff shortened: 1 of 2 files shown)


afourney:
Thanks for the fix. A similar ExifTool UTF-8 decoding correction is proposed in #2067; please reopen if #2067 does not address the original problem.

PR #1996, opened 2026-05-30
Update README with professional rephrasing

printotomp:


printotomp:

PR #1997, opened 2026-05-30
fix: keep pptx charts without title text frames

### LINKED ISSUE #1958: bug: PptxConverter crashes when chart title has no text_frame (AttributeError)

he-yufeng:
﻿## Summary
- skip the PPTX chart title text only when python-pptx exposes no title text frame
- keep converting the chart data table instead of falling back to `[unsupported chart]`
- add a regression test for a titled chart whose `chart_title.text_frame` is `None`

Fixes #1958.

## To verify
- `python -m pytest packages\markitdown\tests\test_module_misc.py -q -k "pptx_chart_title_without_text_frame or exceptions"`
- `python -m py_compile packages\markitdown\src\markitdown\converters\_pptx_converter.py packages\markitdown\tests\test_module_misc.py`
- `git diff --check`

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_pptx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_pptx_converter.py
@@ -236,7 +236,10 @@ class PptxConverter(DocumentConverter):
         try:
             md = "\n\n### Chart"
             if chart.has_title:
-                md += f": {chart.chart_title.text_frame.text}"
+                title_frame = chart.chart_title.text_frame
+                title_text = title_frame.text if title_frame is not None else ""
+                if title_text:
+                    md += f": {title_text}"
             md += "\n\n"
             data = []
             category_names = [c.label for c in chart.plots[0].categories]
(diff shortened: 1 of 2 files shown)


he-yufeng:
Friendly ping. PPTX charts that don't carry a title text frame currently get dropped during conversion; this keeps them instead of losing the chart entirely. Small fix with a regression test, would appreciate a look.

afourney:

PR #1998, opened 2026-05-30
fix: preserve malformed docx math content

### LINKED ISSUE #1979: bug: DOCX math converter crashes when oMath element is missing in malformed equations

### LINKED ISSUE #1982: bug: DOCX math converter crashes with NotImplementedError on unknown functions

he-yufeng:
﻿## Summary
- keep malformed OMML tags from crashing the DOCX math preprocessor when no namespaced `oMath` child is found
- render unknown OMML function names with `\operatorname{...}` instead of raising `NotImplementedError`
- add focused regression tests for the missing-child and unknown-function paths

Fixes #1979.
Fixes #1982.

## To verify
- `$env:PYTHONPATH=(Resolve-Path packages\markitdown\src).Path; python -m pytest packages\markitdown\tests\test_docx_math.py -q`
- `python -m py_compile packages\markitdown\src\markitdown\converter_utils\docx\pre_process.py packages\markitdown\src\markitdown\converter_utils\docx\math\omml.py packages\markitdown\tests\test_docx_math.py`
- `git diff --check`

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
@@ -272,7 +272,9 @@ class oMath2Latex(Tag2Method):
                 if FUNC.get(t):
                     latex_chars.append(FUNC[t])
                 else:
-                    raise NotImplementedError("Not support func %s" % t)
+                    latex_chars.append(
+                        "\\operatorname{%s}(%s)" % (escape_latex(t), FUNC_PLACE)
+                    )
             else:
                 latex_chars.append(t)
         t = BLANK.join(latex_chars)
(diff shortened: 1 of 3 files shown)


he-yufeng:
Gentle nudge — malformed docx math is dropped today instead of being preserved as-is. Small fix; I can attach a sample document if that helps review.

he-yufeng:
Nudge after some weeks. Malformed OMML math in docx is still dropped rather than preserved as text, and this PR keeps that content in the output with regression coverage. The branch has fallen behind main; say the word and I will refresh it for review.

he-yufeng:
The crash this guarded against was fixed upstream; what remains is a keep-vs-drop text preference, not worth the slot. Closing.

PR #2007, opened 2026-05-30
fix: ignore missing exiftool binary

### LINKED ISSUE #1960: exiftool_metadata() crashes with FileNotFoundError when exiftool_path binary doesn't exist

he-yufeng:
﻿## Summary
- treat a missing explicit `exiftool_path` the same way as no exiftool metadata being available
- keep image conversion from failing when the configured binary no longer exists
- add coverage for the helper and the image conversion path

Fixes #1960.

## To verify
- `.venv\Scripts\python.exe -m pytest packages/markitdown/tests/test_module_misc.py -q -k "missing_exiftool_path or exiftool_metadata_ignores_missing_binary or exceptions"`
- `.venv\Scripts\python.exe -m py_compile packages/markitdown/src/markitdown/converters/_exiftool.py packages/markitdown/tests/test_module_misc.py`
- `git diff --check`

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_exiftool.py
+++ b/packages/markitdown/src/markitdown/converters/_exiftool.py
@@ -32,6 +32,8 @@ def exiftool_metadata(
                 f"ExifTool version {version_output} is vulnerable to CVE-2021-22204. "
                 "Please upgrade to version 12.24 or later."
             )
+    except FileNotFoundError:
+        return {}
     except (subprocess.CalledProcessError, ValueError) as e:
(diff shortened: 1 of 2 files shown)


he-yufeng:
Gentle bump. When `exiftool` isn't installed, conversion currently errors out instead of just skipping the EXIF metadata enrichment. This treats the missing binary as "no metadata" and carries on. Would love a review when you have a moment.

afourney:

PR #2011, opened 2026-05-30
fix(ocr): emit valid Markdown newlines for PPTX conversion

### LINKED ISSUE #2010: markitdown-ocr: PPTX converter emits literal \n sequences in Markdown output
`PptxConverterWithOCR` uses `"\\n"` instead of `"\n"` in multiple output paths.
  As a result, converted PPTX files contain literal backslash-n sequences rather
  than Markdown line breaks.

  The OCR PPTX tests currently encode this malformed output as expected behavior.

  Additionally, the optional LLM caption path imports `._llm_caption`, but the
  module exists in `markitdown.converters`, not in `markitdown_ocr`.

MontesanoDev:
Closes #2010

  ## Summary

  `PptxConverterWithOCR` emitted literal `\n` sequences instead of actual line breaks in Markdown output.

  The optional LLM caption path also imported `llm_caption` from a module that does not exist in the `markitdown_ocr` package.

  ## Changes

  - Replace literal `\n` sequences with actual newlines in PPTX Markdown output.
  - Import `llm_caption` from `markitdown.converters._llm_caption`.
  - Update PPTX OCR snapshots to expect valid Markdown line breaks.
  - Add regression tests for newline output and the LLM caption import path.

  ## Testing

  ```bash
  pytest packages/markitdown-ocr/tests/test_pptx_converter.py -v

  Result: 8 passed.

### DIFF
--- a/packages/markitdown-ocr/src/markitdown_ocr/_pptx_converter_with_ocr.py
+++ b/packages/markitdown-ocr/src/markitdown_ocr/_pptx_converter_with_ocr.py
@@ -80,7 +80,7 @@ class PptxConverterWithOCR(DocumentConverter):
 
         for slide in presentation.slides:
             slide_num += 1
-            md_content += f"\\n\\n<!-- Slide number: {slide_num} -->\\n"
+            md_content += f"\n\n<!-- Slide number: {slide_num} -->\n"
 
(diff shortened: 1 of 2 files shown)


MontesanoDev:

PR #2015, opened 2026-05-31
fix: fallback when CSV charset hint fails

### LINKED ISSUE #1949: CsvConverter throws UnicodeDecodeError when charset detection from partial content returns 'ascii'
## Bug

\CsvConverter\ uses \stream_info.charset\ (detected from first 4096 bytes) to decode CSV files. When the first 4096 bytes are ASCII-only but the full file contains non-ASCII characters (e.g., accented characters, UTF-8), \.decode('ascii')\ raises \UnicodeDecodeError\.

## Reproducible example

\\\python
import io

# CSV with UTF-8 characters beyond the first 4096 bytes
buf = io.BytesIO(('x' * 4096 + ',caf\u00e9\n').encode('utf-8'))
md = MarkItDown()
result = md.convert(buf)
\\\

## Expected

CSV content is decoded successfully, falling back to charset_normalizer when the detected charset fails.

## Actual

\UnicodeDecodeError\ because \stream_info.charset\ reports \'ascii'\ but the file contains UTF-8.

## Affected file

\packages/markitdown/src/markitdown/converters/_csv_converter.py\ lines 45-48

## Note

This is the same class of bug as #1505 (PlainTextConverter), which was fixed in PR #1938.

whwhw:
## Summary
- read CSV input bytes once before decoding
- fall back to charset-normalizer when a provided charset hint raises UnicodeDecodeError
- add a regression test for an ASCII hint on UTF-8 CSV content with non-ASCII text after the initial ASCII prefix

Fixes #1949

## Tests
- `uv run --with pytest python -m pytest tests/test_module_misc.py::test_csv_falls_back_when_charset_hint_fails`
- `uv run --with pytest python -m pytest tests/test_module_misc.py::test_csv_falls_back_when_charset_hint_fails tests/test_module_vectors.py -k 'csv or mskanji'`

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -42,10 +42,14 @@ class CsvConverter(DocumentConverter):
         **kwargs: Any,  # Options to pass to the converter
     ) -> DocumentConverterResult:
         # Read the file content
+        data = file_stream.read()
         if stream_info.charset:
-            content = file_stream.read().decode(stream_info.charset)
+            try:
+                content = data.decode(stream_info.charset)
+            except UnicodeDecodeError:
+                content = str(from_bytes(data).best())
         else:
-            content = str(from_bytes(file_stream.read()).best())
+            content = str(from_bytes(data).best())
 
         # Parse CSV content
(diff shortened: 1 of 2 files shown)


whwhw:

PR #2020, opened 2026-05-31
fix(csv): escape pipe characters in CsvConverter to produce valid Markdown tables

### LINKED ISSUE #2019: [BUG]: CsvConverter produces broken Markdown tables when cell values contain pipe characters (|)
## Description

## Steps to Reproduce

Save this as `test.csv`:
```
name,formula
OR gate,A | B
AND gate,A & B
```

Then run:
```python

md = MarkItDown()
print(result.text_content)
```

## Actual Output (Broken)
```
| name | formula |
| --- | --- |
| OR gate | A | B |
| AND gate | A & B |
```

## Expected Output
```
| name | formula |
| --- | --- |
| OR gate | A \| B |
| AND gate | A & B |
```

## Root Cause

```python
```

```python
    return value

```

## Why This Matters

## Environment

- Python 3.10+

Prasad-Pitke:
## What this fixes

Fixes #2019 

`CsvConverter` was joining raw cell values with ` | ` without escaping.
Any cell containing a literal `|` was treated as a column separator,
silently corrupting the table structure.

## Change

Added a small `_escape_cell()` helper in `_csv_converter.py` that
escapes `|` → `\|` before each cell is joined into a table row.

## Before
| OR gate | A | B |   ← 4 columns, broken

## After
| OR gate | A | B |  ← 2 columns, correct

## Testing
- Tested with pipe in header and data rows
- Tested normal CSV (no regression)
- Tested empty CSV and header-only CSV

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -57,20 +57,23 @@ class CsvConverter(DocumentConverter):
         # Create markdown table
         markdown_table = []
 
+        def _escape_cell(value: str) -> str:
+            """Escape pipe characters so they don't break GFM table structure."""
+            value = value.replace("|", r"\|")
+            return value
+
         # Add header row
-        markdown_table.append("| " + " | ".join(rows[0]) + " |")
+        markdown_table.append("| " + " | ".join(_escape_cell(c) for c in rows[0]) + " |")
 
         # Add separator row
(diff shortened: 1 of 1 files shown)


Prasad-Pitke:

afourney:
The CSV pipe-escaping fix proposed here is now covered by #2266, with additional handling for multiline cells and existing backslashes. Please reopen if current main does not address the original problem.

PR #2021, opened 2026-05-31
feat: add TSV support and delimiter auto-detection to CsvConverter

anandmt:
## Summary

- Extend `CsvConverter` to accept `.tsv` files and `text/tab-separated-values` / `text/tsv` MIME types
- Auto-detect delimiters using `csv.Sniffer` with sensible fallbacks (tab for `.tsv`, comma otherwise)
- Escape pipe characters (`|`) and newlines in cell values to prevent broken Markdown tables (also addresses #2019)

## Motivation

TSV (Tab-Separated Values) files are extremely common in data science, bioinformatics, and LLM pipelines, but are not currently supported by MarkItDown. This is a natural extension of the existing CSV converter.

## Changes

- **`_csv_converter.py`**: Added `.tsv` extension and TSV MIME types to accepted formats. Added `_detect_delimiter()` using `csv.Sniffer` for auto-detection with fallbacks. Added `_escape_cell()` to escape pipe characters and newlines in cell values.
- **`test_files/test.tsv`**: New test file with tab-delimited data including a pipe character in a cell value.
- **`_test_vectors.py`**: New `FileTestVector` for TSV conversion with `must_include` checks.

## Test plan

- [x] All 312 existing tests pass (0 failures)
- [x] New TSV test vector passes across all test modes (`convert_local`, `convert_stream_with_hints`, `convert_stream_without_hints`, `convert_file_uri`, `convert_data_uri`)
- [x] Existing CSV test (`test_mskanji.csv`) continues to pass — delimiter auto-detected as comma
- [x] Pipe character in TSV cell correctly escaped as `\|` in output

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -8,13 +8,17 @@ from .._stream_info import StreamInfo
 ACCEPTED_MIME_TYPE_PREFIXES = [
     "text/csv",
     "application/csv",
+    "text/tab-separated-values",
+    "text/tsv",
 ]
-ACCEPTED_FILE_EXTENSIONS = [".csv"]
+ACCEPTED_FILE_EXTENSIONS = [".csv", ".tsv"]
+
+SNIFF_SAMPLE_SIZE = 8192
 
 
 class CsvConverter(DocumentConverter):
     """
-    Converts CSV files to Markdown tables.
+    Converts CSV and TSV files to Markdown tables.
     """
 
     def __init__(self):
@@ -47,8 +51,12 @@ class CsvConverter(DocumentConverter):
         else:
             content = str(from_bytes(file_stream.read()).best())
 
-        # Parse CSV content
-        reader = csv.reader(io.StringIO(content))
+        # Auto-detect the delimiter
+        extension = (stream_info.extension or "").lower()
+        delimiter = self._detect_delimiter(content, extension)
+
+        # Parse content
+        reader = csv.reader(io.StringIO(content), delimiter=delimiter)
         rows = list(reader)
 
         if not rows:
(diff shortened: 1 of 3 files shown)


packages/markitdown/src/markitdown/converters/_csv_converter.py
```diff
@@ -8,13 +8,17 @@
 ACCEPTED_MIME_TYPE_PREFIXES = [
     "text/csv",
     "application/csv",
+    "text/tab-separated-values",
+    "text/tsv",
```
trippinganymess, line 12: I couldn't find text/tsv as IANA recognized MIME type why has this been added?  [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml#text)

packages/markitdown/src/markitdown/converters/_csv_converter.py
```diff
@@ -70,8 +79,24 @@ def convert(
...
             row = row[: len(rows[0])]
-            markdown_table.append("| " + " | ".join(row) + " |")
+            escaped = [self._escape_cell(cell) for cell in row]
+            markdown_table.append("| " + " | ".join(escaped) + " |")
 
         result = "\n".join(markdown_table)
 
         return DocumentConverterResult(markdown=result)
+
+    def _detect_delimiter(self, content: str, extension: str) -> str:
```
trippinganymess, line 89: After doing a little more digging I have found that this implementation is fairly brittle in my opinion. 

1. using the `csv.sniffer` function  for checking for delimiter is highly brittle as it deduces the delimiter using a heuristic methods (like frequency counting) it tends to produce false positives and false negative. you don't check if the guess is correct or not. [docs](https://docs.python.org/3/library/csv.html#csv.Sniffer)
2. If case of `csv.Error`, you only check if the file's extension is ".tsv" but in the case where the file are coming from Object Storage (AWS S3, Google Cloud Storage), these files usually don't have any extension at all, hence checking the MIME type along with extension is better way to fallback. currently all files without extension during csv.Error simply fallback to CSV type.

PR #2027, opened 2026-05-31
Add TSV file support to CsvConverter

### LINKED ISSUE #2022: Support TSV (Tab-Separated Values) files
## Description

The `CsvConverter` currently only handles `.csv` files with comma delimiters. TSV (Tab-Separated Values) files are extremely common in data science, bioinformatics, spreadsheet exports, and LLM pipelines, but are not supported.

## Expected behavior

MarkItDown should convert `.tsv` files to Markdown tables, just like it does for `.csv` files.

## Proposed solution

Extend the existing `CsvConverter` to:
1. Accept `.tsv` files and `text/tab-separated-values` MIME type
2. Auto-detect the delimiter using Python's built-in `csv.Sniffer`
3. Fall back to tab for `.tsv` files, comma for `.csv` files

## PR

#2021

shrajal01:
## Summary

Adds support for TSV (Tab-Separated Values) files in CsvConverter.

### Changes

* Added `.tsv` to accepted file extensions
* Added `text/tab-separated-values` MIME type support
* Added TSV delimiter handling (`\t`)
* Added test coverage and sample TSV test file

Fixes #2022

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -49,7 +49,15 @@ class CsvConverter(DocumentConverter):
             content = str(from_bytes(file_stream.read()).best())
 
         # Parse CSV content
-        delimiter = "\t" if (stream_info.extension or "").lower() == ".tsv" else ","
+        extension = (stream_info.extension or "").lower()
+        mimetype = (stream_info.mimetype or "").lower()
+
+        is_tsv = (
+            extension == ".tsv"
+            or mimetype.startswith("text/tab-separated-values")
+        )
+
+        delimiter = "\t" if is_tsv else ","
 
         reader = csv.reader(
             io.StringIO(content),
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -8,8 +8,9 @@ from .._stream_info import StreamInfo
 ACCEPTED_MIME_TYPE_PREFIXES = [
     "text/csv",
     "application/csv",
+    "text/tab-separated-values",
 ]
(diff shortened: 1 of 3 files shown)


shrajal01:

packages/markitdown/src/markitdown/converters/_csv_converter.py
```diff
@@ -48,7 +49,12 @@ def convert(
             content = str(from_bytes(file_stream.read()).best())
 
         # Parse CSV content
-        reader = csv.reader(io.StringIO(content))
+        delimiter = "\t" if (stream_info.extension or "").lower() == ".tsv" else ","
```
trippinganymess: the accept() function, uses the MIME type and the extension of a file to determine it's type. so In a scenario where a file arrives from a cloud service with no extension, but MIME type would be accepted. but You are  currently setting the delimiter by only checking the extension type and not the MIME type. This will cause the function to crash in case a file without an extension is accepted and will automatically default to "," every time.


Example scenario : 

1. file received from cloud ( without extension but with MIME type "tab-seperated-values")
2. the accept function allows this file and returns True.
3. `delimiter = "\t" if (stream_info.extension or "").lower() == ".tsv" else ","` doesn't account for the MIME type and defaults to "," which crashes the whole program.

trippinganymess: This also has been addressed in #2021[](https://github.com/microsoft/markitdown/pull/2021)

trippinganymess (changes requested):
the delimiter should explicitly check both the extension and the MIME type before being intialized.

packages/markitdown/src/markitdown/converters/_csv_converter.py
shrajal01: Thanks for the review! Good catch. I'll update the delimiter detection to consider both the file extension and MIME type.

PR #2028, opened 2026-05-31
Fix #2019: Escape pipe characters in CSV-to-Markdown table conversion

### LINKED ISSUE #2019: [BUG]: CsvConverter produces broken Markdown tables when cell values contain pipe characters (|)

dry-shy:
## Summary

Fixes #2019 by escaping pipe (`|`) characters in CSV cell values before generating Markdown tables.

## Problem

Markdown tables use the pipe character (`|`) as a column separator. When CSV cell values contain unescaped pipe characters, the generated Markdown table becomes malformed and renders incorrectly in GitHub Flavored Markdown.

Example CSV:

```csv
Name,Value
A,x|y
```

Current output:

```markdown
| Name | Value |
| --- | --- |
| A | x|y |
```

This causes the table structure to break because `x|y` is interpreted as multiple columns.

## Solution

Added a helper function to escape pipe characters in header and data cells before rendering Markdown rows:

```python
    return str(value).replace("|", r"\|")
```

All table cells are now processed through `_escape_cell()` before being added to the Markdown output.

## Result

Generated Markdown:

```markdown
| Name | Value |
| --- | --- |
| A | x\|y |
```

The table now renders correctly even when cell values contain pipe characters.

## Testing

Added/updated tests to verify that:

* Pipe characters in header cells are escaped.
* Pipe characters in data cells are escaped.
* Generated Markdown tables render correctly in GitHub Flavored Markdown.

## Related Issue

Fixes #2019

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -57,20 +57,32 @@ class CsvConverter(DocumentConverter):
         # Create markdown table
         markdown_table = []
 
+        def _escape_cell(value: str) -> str:
+            """Escape pipe characters so they don't break Markdown table structure."""
+            return str(value).replace("|", r"\|")
+
         # Add header row
-        markdown_table.append("| " + " | ".join(rows[0]) + " |")
+        markdown_table.append(
+            "| " + " | ".join(_escape_cell(c) for c in rows[0]) + " |"
+        )
 
         # Add separator row
-        markdown_table.append("| " + " | ".join(["---"] * len(rows[0])) + " |")
+        markdown_table.append(
+            "| " + " | ".join(["---"] * len(rows[0])) + " |"
+        )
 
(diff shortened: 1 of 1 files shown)


afourney:
CSV pipe escaping has now landed through #2266, which also handles newlines and preceding backslashes correctly. I’m closing this as superseded; please reopen if a case remains broken.

PR #2032, opened 2026-06-01
fix: patch CWE-611 XXE vulnerabilities

yabets4:
## Summary


yabets4:

PR #2035, opened 2026-06-01
Escape pipes and newlines in CSV to Markdown table cells

LeSingh1:
## Summary

`CsvConverter` writes cell values straight into the Markdown table without escaping. Two cases produce malformed output:

- A cell containing a literal `|` adds an extra column separator.
- A quoted CSV field with an embedded newline (valid CSV) splits the row in two.

In both cases the data rows no longer match the header column count, so the table renders incorrectly.

## Reproduction

```
input:  name,note\nAlice,a|b\nBob,plain
```

The `Alice` row renders as three columns instead of two. A quoted multi-line field like `"line1\nline2"` similarly breaks into two rows.

## Fix

Added an `_escape_cell` helper that escapes `|` as `\|` and flattens `\r\n` / `\n` / `\r` to spaces, applied to header and data cells.

```
| Alice | a\|b |
| Bob | line1 line2 |
```

## Tests

Added `packages/markitdown/tests/test_csv_converter.py` with `test_pipe_in_cell_is_escaped` (asserts every row has equal unescaped pipe count) and `test_newline_in_cell_does_not_break_row`. Both fail on the unpatched converter and pass after the change. The existing module test vectors still pass.

```
$ python -m pytest tests/test_csv_converter.py -q
2 passed
```

---
Developed with AI assistance and verified locally as described above.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -12,6 +12,14 @@ ACCEPTED_MIME_TYPE_PREFIXES = [
 ACCEPTED_FILE_EXTENSIONS = [".csv"]
 
 
+def _escape_cell(value: str) -> str:
+    # A literal pipe or newline inside a cell would split it into extra
+    # columns or rows when rendered, so escape pipes and flatten newlines.
+    return value.replace("|", "\\|").replace("\r\n", " ").replace("\n", " ").replace(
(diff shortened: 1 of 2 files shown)


afourney:
Thanks for the fix. CSV pipe characters are now escaped by #2266, including in headers and data cells. I’m closing this as superseded; if the problem remains on current `main`, please let us know and we can reopen.

PR #2044, opened 2026-06-01
fix: handle chart titles with no text_frame (#1958)

doitgo:
When a chart has `has_title=True` but `chart_title.text_frame` is None, the
  converter raised `AttributeError`, which was silently swallowed by a broad
  `except Exception` in `_convert_chart_to_markdown`, dropping valid chart data.        

  This change:
  - Adds a `None` check on `text_frame` before accessing `.text`
  - Applies the same fix to `markitdown-ocr` (`PptxConverterWithOCR`)
  - Adds regression tests in `tests/test_module_misc.py`


afourney:

PR #2045, opened 2026-06-01
fix: support extended content-disposition filenames

cat0825:
## Summary

- support RFC 5987/RFC 2231 `filename*=` values in `Content-Disposition` response headers
- prefer the extended filename over the fallback `filename=` value when both are present
- add regression coverage for `convert_response()` extension detection from extended filenames

## Testing

- `python -m pytest packages/markitdown/tests/test_module_misc.py -k 'content_disposition_filename'`
- `OPENAI_API_KEY= hatch test` from `packages/markitdown`
- `python -m pre_commit run --all-files`


cat0825:
Merged current `main` into this branch (merge commit `e366c06`, no history rewrite) — the PR is conflict-free again and now reports `MERGEABLE`.

Conflict details:
- `packages/markitdown/tests/test_module_misc.py` — the only conflict, and purely additive: this PR and upstream `#2233` both appended a new test at the same spot. Resolved by keeping both, so the two RFC 5987 regression tests sit alongside upstream's `test_pptx_chart_multi_series_conversion`.
- `packages/markitdown/src/markitdown/_markitdown.py` — auto-merged cleanly. I re-checked `convert_response` on current `main`: it still uses `re.search(r"filename=([^;]+)")`, so `filename*` is still unhandled upstream and this fix is not superseded. Also dropped one stray blank line the original commit had added near the `_exceptions` import, so the diff is now exactly +62/-3 across the two files with no unrelated changes.

Verified locally against merged `main` (Python 3.12, `pip install -e "packages/markitdown[all]"`):
- `python -m pytest packages/markitdown/tests/test_module_misc.py -q` -> `15 passed, 2 skipped in 31.74s`
- `python -m pytest packages/markitdown/tests/test_module_misc.py -k content_disposition -v` -> `2 passed, 15 deselected`
- Sanity check that the tests actually pin the behavior: reverting just the parsing call back to the current `main` regex makes `test_convert_response_prefers_extended_content_disposition_filename` fail (`name,value\nalpha,beta` instead of the CSV table), and it passes again once restored.
- `black==23.7.0 --check` on both changed files -> `2 files would be left unchanged`.

Note the fork workflows still need a maintainer to approve the CI run — only `license/cla` has reported so far.

@afourney whenever you have a moment, this one is ready for a look.

afourney:
Thanks! Merged!

PR #2050, opened 2026-06-01
zip protection to zip converter

blackheart-5:
## Problem
ZipConverter had no protection against zip bombs. A maliciously crafted 
zip could cause OOM or infinite recursion if these exist:
- High compression ratio (small zip → massive uncompressed size)
- Millions of tiny files
- Deeply nested zips (zip inside zip inside zip...)

## Changes
- Added `ZIP_UNCOMPRESSED_SIZE_THRESHOLD = 100 MB` — rejects zips whose 
  total uncompressed size exceeds limit
- Added `MAX_FILE_COUNT = 1000` — rejects zips with too many files
- Added `MAX_DEPTH = 3` — rejects zips nested deeper than 3 levels
- Depth counter passed through `convert_stream` recursively

## Testing
Manually verified with normal zips and simulated oversized inputs.


blackheart-5:

PR #2052, opened 2026-06-02
Fix ipynb title extraction stripping leading # from heading text

LeSingh1:
## Problem

When converting a Jupyter notebook, the document title is taken from the first Markdown `# ` heading. The extraction uses:

```python
title = line.lstrip("# ").strip()
```

`str.lstrip(chars)` treats its argument as a *set of characters* to strip, not a prefix. So it removes every leading `#` and space character, not just the single `"# "` marker. A heading whose text itself begins with `#` loses those characters.

Example: a notebook whose first heading is `# #100DaysOfCode` produces the title `100DaysOfCode` instead of `#100DaysOfCode`.

## Fix

Strip only the `"# "` marker by slicing off its known length:

```python
title = line[len("# ") :].strip()
```

The line already passed `line.startswith("# ")`, so the prefix is guaranteed present. Outer whitespace is still trimmed by `.strip()`.

## Tests

Added `tests/test_ipynb_converter.py`:
- a heading whose text starts with `#` keeps it (`# #100DaysOfCode` -> `#100DaysOfCode`)
- only the marker is removed, internal spacing preserved
- a plain heading is unchanged

All three pass; the first fails on the old code.

---
Disclosure: I used AI assistance while preparing this change. I reproduced the bug and verified the fix and tests locally.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_ipynb_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_ipynb_converter.py
@@ -71,7 +71,7 @@ class IpynbConverter(DocumentConverter):
                     if title is None:
                         for line in source_lines:
                             if line.startswith("# "):
-                                title = line.lstrip("# ").strip()
+                                title = line[len("# ") :].strip()
                                 break
 
                 elif cell_type == "code":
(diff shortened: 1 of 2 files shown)


afourney:
Thanks for the fix. The same notebook-title issue is now covered by #2371.

PR #2054, opened 2026-06-02
docs: add Windows install notes

xiazl1993:
## Summary
- add PowerShell virtual environment activation guidance
- document PowerShell quoting for optional dependency installs
- add a python -m markitdown fallback when the console script is not on PATH

## Validation
- ran git diff --check
- confirmed packages/markitdown/src/markitdown/__main__.py exists for the module fallback

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ python -m venv .venv
 source .venv/bin/activate
 ```
 
+On Windows PowerShell, activate the virtual environment with:
+
+```powershell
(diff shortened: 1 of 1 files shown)


xiazl1993:

chienyuanchang:
Thanks for putting this together. These Windows install notes are useful, but we generally reserve README churn for changes that correct an instruction or unblock a real issue. If you'd like to contribute, please open a new PR for a concrete docs gap or a user-facing problem we can act on.

Issue #2049, opened 2026-06-01
Feature Request: Save embedded images from PDFs as separate files with relative Markdown links


agodianel:
### Problem
When converting PDFs with embedded images to Markdown, markitdown currently either:
- Truncates base64 data (showing `![](data:image/png;base64...)`), or  
- Preserves full base64 inline (which makes the Markdown file very large and unsuitable for RAG/indexing)

For full-context RAG pipelines and document archival, I need:
1. Embedded images extracted and saved as separate files (e.g. `images/page3-figure1.png`)
2. Markdown references using relative paths: `![description](images/page3-figure1.png)`
3. Optional: image metadata/captions in front matter or inline comments

### Use case
- Converting technical documentation, datasheets, and PDFs with diagrams for AI/LLM ingestion
- Preserving full document context including visual information
- Keeping Markdown files lightweight and portable while still referencing images

### Proposed behavior
Add a CLI flag and Python API option, for example:

```bash
markitdown document.pdf --extract-images --output-dir ./output
```

```python
from markitdown import MarkItDown
md = MarkItDown(extract_images=True, output_dir="./output")
result = md.convert("document.pdf")
# result.markdown includes: 
# Actual image files saved to ./output/images/
```

### Alternatives considered
- Full base64 inline: makes files too large for indexing/LLM pipelines
- Image descriptions only: loses visual context for diagrams, charts, schematics
- Manual extraction: error-prone and not reproducible

### Additional context
- Related to issue #162 (closed) and PR #1140 (base64 preservation added)
- Complements issue #1346 (LLM image descriptions for RAG)
- Would benefit users working with schematics, datasheets, research papers, and technical documentation

### Environment
- markitdown version: 0.1.6
- Python version: 3.11+
- OS: macOS

Imposter069:
I strongly support this feature request.

For RAG pipelines, technical documentation, research papers, and datasheet processing, embedding images as base64 inside Markdown creates significant challenges. Large base64 strings increase file size, make indexing less efficient, and can unnecessarily consume context window space when documents are processed by LLMs.

Saving embedded images as separate files and referencing them through relative Markdown links would provide several benefits:

* Keeps Markdown files lightweight and portable
* Preserves original visual context (diagrams, charts, schematics, screenshots)
* Improves compatibility with RAG and vector database ingestion workflows
* Makes document versioning and storage more efficient
* Enables reproducible document conversion without manual image extraction

An example output structure could be:

```text
output/
├── document.md
[...]

PR #2055, opened 2026-06-02
Extract embedded PDF images as separate files with relative Markdown links

haydaramru:
## Summary

This PR adds an opt-in PDF image extraction workflow for MarkItDown. When enabled, embedded images from PDFs are saved as separate files under an output directory, and the generated Markdown references them using relative links.

This is intended for RAG, document archival, technical documentation, datasheets, schematics, research papers, and other workflows where inline base64 images make Markdown too large or impractical for indexing.

Example:

```bash
markitdown document.pdf --extract-images --output-dir ./output -o ./output/document.md
```

Output:

```text
output/
├── document.md
└── images/
    ├── page1-image1.png
    └── page1-image2.png
```

Markdown:

```markdown
![Image 1 on page 1](images/page1-image1.png)
```

### CLI

Adds two new CLI options:

```bash
--extract-images
--output-dir <directory>
```

`--output-dir` is required when `--extract-images` is used. Existing stdout behavior is preserved: if `-o` is not provided, Markdown is still printed to stdout, while images are written under `--output-dir/images/`.

### Python API

Adds constructor-level support:

```python

md = MarkItDown(extract_images=True, output_dir="./output")
result = md.convert("document.pdf")
```

Also supports per-call overrides:

```python
md = MarkItDown()
result = md.convert(
    "document.pdf",
    extract_images=True,
    output_dir="./output",
)
```

If image extraction is enabled without an output directory, MarkItDown raises:

```text
ValueError: output_dir is required when extract_images=True
```

### PDF conversion

The PDF converter now has an opt-in image extraction path. Default PDF conversion remains unchanged unless `extract_images=True` is explicitly set.

The implementation:

- Detects images from `pdfplumber` page image objects.
- Saves images under `output_dir/images/`.
- Uses deterministic filenames:
  - `page1-image1.png`
  - `page1-image2.png`
  - `page3-image1.jpg`
- Inserts relative Markdown image links near the image's approximate vertical page position.
- Preserves the existing fallback behavior for normal text extraction.
- Keeps page cleanup behavior by still calling `page.close()`.

### Image writing behavior

The converter first tries to save original embedded image stream bytes when they are already recognizable image formats, including:

- PNG
- JPEG
- GIF
- TIFF
- JPEG 2000

If direct stream extraction is not possible, it falls back to rendering/cropping the image region from the PDF page using `pdfplumber`.

The current fallback render setting is:

```python
cropped_page.to_image(resolution=600)
```

This improves output sharpness for diagrams and technical images, but produces larger files and may be slower than lower DPI values such as `150` or `300`.

## Why not inline base64?

Before this feature, PDF/image-heavy Markdown workflows often had two poor options:

- truncate data URIs, losing the actual image content;
- keep full base64 data URIs, creating very large Markdown files.

Both are problematic for RAG and indexing. Extracted image files with relative links keep Markdown lightweight while preserving visual context.

## Testing

Added focused coverage for image extraction, relative Markdown links, `output_dir` validation, CLI help/validation, and existing PDF page cleanup behavior. Full `hatch test` and `pre-commit` checks was also run and `Passed`.

## Manual verification

Example command:

```bash
mkdir -p ./coba

markitdown \
  packages/markitdown-ocr/tests/ocr_test_data/pdf_multiple_images.pdf \
  --extract-images \
  --output-dir ./coba \
  -o ./coba/document.md
```

Expected files:

```text
coba/document.md
coba/images/page1-image1.png
coba/images/page1-image2.png
```

Example Markdown output:

```markdown
Document with Multiple Images

![Image 1 on page 1](images/page1-image1.png)

Text between first and second image.

![Image 2 on page 1](images/page1-image2.png)

Final text after all images.
```

## Files changed

Core implementation:

- `packages/markitdown/src/markitdown/__main__.py`
- `packages/markitdown/src/markitdown/_markitdown.py`
- `packages/markitdown/src/markitdown/converters/_pdf_converter.py`

Tests:

- `packages/markitdown/tests/test_cli_misc.py`
- `packages/markitdown/tests/test_pdf_images.py`

## Future improvements

Possible follow-up work:

- Add a configurable render DPI option, for example `--image-dpi 300`.
- Add custom image directory support, for example `--image-dir assets/images`.
- Add metadata output for page number, image dimensions, bbox, and format.
- Add caption detection when nearby text appears to be a figure caption.
- Add duplicate image detection to avoid writing identical images repeatedly.
- Extend similar extraction behavior to DOCX/PPTX/XLSX if useful.

## Related issues

This feature addresses the same workflow need described in Issue #2049 and complements prior work around base64 data URI handling and image descriptions for RAG workflows.

### DIFF
--- a/packages/markitdown/src/markitdown/__main__.py
+++ b/packages/markitdown/src/markitdown/__main__.py
@@ -138,8 +138,22 @@ def main():
         help="Keep data URIs (like base64-encoded images) in the output. By default, data URIs are truncated.",
     )
 
+    parser.add_argument(
+        "--extract-images",
+        action="store_true",
+        help="Extract embedded PDF images to files and reference them with relative Markdown links.",
+    )
+
+    parser.add_argument(
+        "--output-dir",
+        help="Directory for extracted assets. Required when using --extract-images.",
+    )
+
     parser.add_argument("filename", nargs="?")
     args = parser.parse_args()
+
+    if args.extract_images and args.output_dir is None:
+        _exit_with_error("--output-dir is required when using --extract-images.")
 
     # Parse the extension hint
     extension_hint = args.extension
@@ -209,7 +223,10 @@ def main():
             _exit_with_error("Filename is required when using Document Intelligence.")
 
         markitdown = MarkItDown(
-            enable_plugins=args.use_plugins, docintel_endpoint=args.endpoint
+            enable_plugins=args.use_plugins,
+            docintel_endpoint=args.endpoint,
+            extract_images=args.extract_images,
+            output_dir=args.output_dir,
         )
     elif args.use_cu:
         if args.cu_endpoint is None:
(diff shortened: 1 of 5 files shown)


haydaramru:

PR #2060, opened 2026-06-03
fix: allow DocumentIntelligenceConverter api_version to default to None (fixes #1904)

### LINKED ISSUE #1904: DocumentIntelligenceConverter should not set a default api_version

li5435945-ship-it:
## Description

Fixes #1904

The  previously hardcoded  as the default parameter. This overrode the Azure Document Intelligence SDK's own default version (), causing 404 errors for users who didn't explicitly specify an .

## Problem

1. Azure Document Intelligence SDK defaults to 
2.  forced  as default
3. Users with standard Azure deployments (which don't specify api_version) got 404 errors because the converter used an outdated preview version

## Fix

- Changed  default from  to 
- Only passes  to  when explicitly set
- When , the SDK uses its own built-in default version
- Users can still set  explicitly when needed

## Changes

- :
  -  → 
  - Build  dict conditionally, only including  when not None
  - Updated docstring to reflect the new behavior

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_doc_intel_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_doc_intel_converter.py
@@ -134,7 +134,7 @@ class DocumentIntelligenceConverter(DocumentConverter):
         self,
         *,
         endpoint: str,
-        api_version: str = "2024-07-31-preview",
+        api_version: str | None = None,
         credential: AzureKeyCredential | TokenCredential | None = None,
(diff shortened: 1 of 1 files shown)


li5435945-ship-it:

afourney:
Thanks for the fix. The api_version=None behavior and regression coverage have now landed through #2267, so I’m closing this as superseded. Please let me know—or reopen this—if the problem is not fully addressed on current main.

PR #2062, opened 2026-06-03
docs: quote optional dependency install examples

nyxst4ck:
Quotes optional dependency install examples so shells do not expand package extras, and fixes a small MarkItDown typo in an audio dependency message.

### DIFF
--- a/packages/markitdown/README.md
+++ b/packages/markitdown/README.md
@@ -13,7 +13,7 @@
 From PyPI:
 
 ```bash
-pip install markitdown[all]
+pip install 'markitdown[all]'
 ```
 
 From source:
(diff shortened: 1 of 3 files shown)


nyxst4ck:

noezhiya-dot (approved):
Small but important docs fix. Without quotes, shells like zsh expand `markitdown[all]` as a glob pattern (trying to match files in the current directory), which either fails silently or installs the wrong thing. Quoting is the standard practice for pip install examples.

The typo fix in the audio dependency message is a nice bonus. LGTM.

afourney:
Thank you for the shell-quoting and wording fixes. [#2223](https://github.com/microsoft/markitdown/pull/2223) incorporated all of the changes proposed here: the package README examples, the missing-dependency installation guidance, and the corrected audio-transcription installation message.

These changes are already on main and in v0.1.7, so I'm closing this as already addressed.

PR #2066, opened 2026-06-03
fix: escape pipe characters in CSV table cells

### LINKED ISSUE #2019: [BUG]: CsvConverter produces broken Markdown tables when cell values contain pipe characters (|)

zekiyemeral:
## Summary

This pull request fixes a Markdown table formatting issue in `CsvConverter`.

When a CSV cell contains a literal pipe character (`|`), the generated Markdown table can be parsed incorrectly because the pipe is treated as a column separator. This change escapes pipe characters inside CSV cell values before joining them into Markdown table rows.

## Changes

- Added a helper function to escape literal `|` characters in Markdown table cells.
- Applied escaping to CSV header cells.
- Applied escaping to CSV data cells.
- Improved CSV to Markdown table output for values containing pipe characters.

## Why This Matters

CSV files may contain pipe characters in formulas, command examples, regular expressions, SQL snippets, and other real-world data. Without escaping, the generated Markdown output can become structurally incorrect.

Closes #2019

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -1,15 +1,23 @@
 import csv
 import io
 from typing import BinaryIO, Any
+
 from charset_normalizer import from_bytes
+
 from .._base_converter import DocumentConverter, DocumentConverterResult
 from .._stream_info import StreamInfo
+
 
 ACCEPTED_MIME_TYPE_PREFIXES = [
     "text/csv",
     "application/csv",
 ]
+
 ACCEPTED_FILE_EXTENSIONS = [".csv"]
+
+
+def _escape_markdown_table_cell(cell: str) -> str:
+    return cell.replace("|", r"\|")
 
 
 class CsvConverter(DocumentConverter):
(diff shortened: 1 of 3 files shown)


zekiyemeral:

trippinganymess:
The solution looks good from pipes, but I think we should also considers, rogue newlines and carriage returns which can easily be introduced into the files breaking the table structure. 

check my PR for the same issue #2061  and lmk what you think.

ericsondea-collab (commented):
insn  fkd  fldv dpjfl  flmd

afourney:
Thanks for the fix. CSV pipe characters are now escaped by #2266, including in headers and data cells. I’m closing this as superseded; if the problem remains on current `main`, please let us know and we can reopen.

PR #2075, opened 2026-06-04
Add optional desktop GUI entry point

AthillaZaidan:
## Summary

  Adds a desktop GUI entry point for MarkItDown so users can convert one local file into a Markdown output file without using
  the CLI directly.

  ## Changes

  - Add `markitdown-gui` console script
  - Add PySide6-based dark desktop GUI
  - Add native file picker for input and output paths
  - Reuse existing `MarkItDown` Python API for conversion
  - Add PDF dependencies to default install so GUI PDF conversion works out of the box
  - Add GUI-focused tests
  - Document GUI setup and run commands in README files

  ## How to run

  ```bash
  cd packages/markitdown
  uv pip install -e .
  markitdown-gui

  If using an already active virtual environment:

  cd packages/markitdown
  pip install -e .
  markitdown-gui

  ## Verification

  - uv run --with pytest pytest tests/test_gui.py -q
  - uv run --with pytest pytest tests/test_cli_misc.py -q
  - Manual PDF conversion smoke via GUI helper

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -166,3 +166,4 @@ cython_debug/
 src/.DS_Store
 .DS_Store
 .cursorrules
+.superpowers/
--- a/README.md
+++ b/README.md
@@ -70,6 +70,38 @@ pip install -e 'packages/markitdown[all]'
 
 ## Usage
 
+### Desktop GUI
+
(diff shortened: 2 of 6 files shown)


AthillaZaidan:

afourney:

PR #2078, opened 2026-06-05
Don't abort PDF conversion on a single malformed page

### LINKED ISSUE #2071: PdfConverter fails on valid PDF: ValueError not enough values to unpack (expected 4, got 2)
### Summary

`markitdown` fails to convert a PDF with `PdfConverter threw ValueError: not enough values to unpack (expected 4, got 2)`.

The `ffmpeg` / `pydub` warning appears first, but seems unrelated because this is a PDF conversion.

### Command

```powershell
markitdown chapter11.pdf -o chapter11.md

Result:
 - PdfConverter threw ValueError with message: not enough values to unpack (expected 4, got 2)

Traceback:
markitdown\converters\_pdf_converter.py
pdfminer.high_level.extract_text(...)
ValueError: not enough values to unpack (expected 4, got 2)

Environment
OS: Windows
Python: 3.13.13
markitdown: 0.1.6
pdfplumber: 0.11.9
pdfminer.six: 20251230

<img width="1835" height="798" alt="Image" src="https://github.com/user-attachments/assets/3a8681c3-df11-4cc8-a0f1-7c04131ac84b" />

assinscreedFC:
## Summary

Closes #2071.

`PdfConverter.convert` ran the entire per-page `pdfplumber` loop inside a single `try`, and on any page error it abandoned the whole loop and fell back to an **unguarded** whole-document `pdfminer.high_level.extract_text` call. A single malformed page — for example a form XObject whose `/BBox` is not a four-number rectangle, which makes the PDF backend raise `ValueError: not enough values to unpack (expected 4, got 2)` — therefore aborted the entire conversion with `FileConversionException`, even when every other page was fine.

## Change

- Each page's extraction is wrapped in `try/except` so one malformed page is skipped (logged at `debug` level via a module logger) while the text recovered from the other pages is kept. `page.close()` moved to a `finally` so cached page data is always freed.
- The `form_page_count == 0` whole-document `pdfminer` call is now guarded and falls back to the per-page `pdfplumber` text already collected.
- The two remaining `extract_text` fallbacks are guarded so they degrade to `""` instead of raising.

The deeper root cause is a robustness gap in `pdfminer.six` (it unpacks `/BBox` without validating its length); that is being fixed upstream separately. This PR makes `markitdown` degrade gracefully regardless of the backend version.

## Tests

Added `packages/markitdown/tests/test_pdf_invalid_bbox.py` (self-contained, building the PDF in-memory): a two-page document whose second page has a malformed form `/BBox` converts without raising and still recovers the first page's text.

`pytest packages/markitdown/tests/test_pdf_invalid_bbox.py packages/markitdown/tests/test_pdf_masterformat.py` → passes; `black --check` clean.

---

AI assistance (Claude) was used; I have reviewed every line and run the tests.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_pdf_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_pdf_converter.py
@@ -1,11 +1,14 @@
 import sys
 import io
+import logging
 import re
 from typing import BinaryIO, Any
 
 from .._base_converter import DocumentConverter, DocumentConverterResult
 from .._stream_info import StreamInfo
 from .._exceptions import MissingDependencyException, MISSING_DEPENDENCY_MESSAGE
+
+_log = logging.getLogger(__name__)
 
 # Pattern for MasterFormat-style partial numbering (e.g., ".1", ".2", ".10")
 PARTIAL_NUMBERING_PATTERN = re.compile(r"^\.\d+$")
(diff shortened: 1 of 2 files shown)


assinscreedFC:

ericsondea-collab (commented):
sasos csamknk´ps cpm  a´ndps kdlsçdksl pskm dosnnv´pnn
dsnd dmsjsm onlldponnf s dnskdnks  nosnskn sdknnc

ericsondea-collab (commented):
nsd sv jjg  skdpmd  sp´b  fkd]jec xlz dl , c,s  cçz  d,sdmgfkm dpsl  
d  sk  smsfn sllv,s vsm v~s spod s

PR #2079, opened 2026-06-05
Fix: Update ZipConverter docstring and optimize memory handling for large entries

KhudaBuxMagsi:
**Docstring Correction:** Updated the class docstring to accurately reflect that files are processed in-memory rather than extracted to a temporary directory on disk.

**Memory Optimization:** Safely structured the file reading process inside the zipObj loop using zipObj.open to ensure better stream handling, reducing the risk of unexpected memory usage spikes.


KhudaBuxMagsi:

kervin-carlos:
Looks good to me.

PR #2082, opened 2026-06-06
fix: catch OSError when exiftool binary is missing (#1960)

### LINKED ISSUE #1960: exiftool_metadata() crashes with FileNotFoundError when exiftool_path binary doesn't exist

doitgo:
Fixes #1960
When exiftool_path points to a binary that does not exist, the subprocess.run calls raised FileNotFoundError, which propagated unhandled through exiftool_metadata() and crashed the whole conversion.

Add OSError to the except clauses at both subprocess.run sites and wrap the error in a RuntimeError that includes the path, giving users a clear, actionable error message.

Add a regression test for the missing-binary case and a sanity check that the exiftool_path=None early return still works.


doitgo:
Hi @afourney,

I'd like to draw your attention to PR #2082, which fixes #1960.

**Issue**: when the configured `exiftool` path points to a binary that doesn't exist, the current code raises an unhandled `FileNotFoundError` that crashes the entire conversion.

**Fix** (+27 / -2 across 2 files):
- `src/markitdown/converters/_exiftool.py`: catch `OSError` (parent of `FileNotFoundError` / `PermissionError`) at both `subprocess.run` sites and re-raise as `RuntimeError` with the offending path included for easier debugging.
- `tests/test_exiftool_converter.py`: add a regression test for the missing-binary case and a sanity check that the `exiftool_path=None` early-return path still returns `{}`.

Local tests pass. Could you or another maintainer take a look when you have a moment? Happy to adjust the implementation, the error message wording, or the test coverage based on your feedback.

Thanks!

doitgo:
Thanks @afourney for the merge and the follow-up cleanups on the error handling and tests — really appreciate the extra  polish.

PR #2092, opened 2026-06-08
Recover PDF text truncated by inline images using PyMuPDF fallback

### LINKED ISSUE #1870: PDF text after an inline image (`BI ... EI`) is silently dropped from extraction

Muhtasim-Munif-Fahim:
Fixes #1870.

Some PDFs with inline images cause pdfplumber/pdfminer to silently drop text that appears after the image. When the primary extraction path returns a much shorter body on image-bearing pages, this change optionally retries with PyMuPDF and prefers that output if it is substantially longer.

Also documents the optional pymupdf extra and adds a regression test that simulates the truncation path.

### DIFF
--- a/packages/markitdown/README.md
+++ b/packages/markitdown/README.md
@@ -16,6 +16,12 @@ From PyPI:
 pip install 'markitdown[all]'
 ```
 
+If you need the optional PyMuPDF fallback for some PDFs:
+
+```bash
+pip install markitdown[pymupdf]
+```
+
 From source:
 
 ```bash
--- a/packages/markitdown/pyproject.toml
+++ b/packages/markitdown/pyproject.toml
@@ -42,6 +42,7 @@ all = [
   "lxml",
   "pdfminer.six>=20251230",
   "pdfplumber>=0.11.9",
+  "pymupdf",
   "olefile",
   "pydub",
   "SpeechRecognition",
@@ -55,6 +56,7 @@ docx = ["mammoth~=1.11.0", "lxml"]
 xlsx = ["pandas", "openpyxl"]
 xls = ["pandas", "xlrd"]
 pdf = ["pdfminer.six>=20251230", "pdfplumber>=0.11.9"]
+pymupdf = ["pymupdf"]
 outlook = ["olefile"]
 audio-transcription = ["pydub", "SpeechRecognition"]
 youtube-transcription = ["youtube-transcript-api"]
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -54,6 +54,12 @@ class CsvConverter(DocumentConverter):
         if not rows:
             return DocumentConverterResult(markdown="")
 
+        while rows and not any(cell.strip() for cell in rows[0]):
+            rows.pop(0)
+
+        if not rows:
+            return DocumentConverterResult(markdown="")
+
         # Create markdown table
         markdown_table = []
 
--- a/packages/markitdown/src/markitdown/converters/_pdf_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_pdf_converter.py
@@ -57,6 +57,23 @@ def _merge_partial_numbering_lines(text: str) -> str:
     return "\n".join(result_lines)
 
 
+def _extract_with_pymupdf(pdf_bytes: io.BytesIO) -> str | None:
+    """Extract text with PyMuPDF when the primary PDF parsers look truncated."""
+    if fitz is None:
+        return None
+
+    pdf_bytes.seek(0)
+    chunks: list[str] = []
+    with fitz.open(stream=pdf_bytes.read(), filetype="pdf") as doc:
+        for page in doc:
+            text = page.get_text("text")
+            if text and text.strip():
+                chunks.append(text.strip())
+
+    markdown = "\n\n".join(chunks).strip()
+    return markdown or None
+
+
 # Load dependencies
 _dependency_exc_info = None
 try:
@@ -65,6 +82,11 @@ try:
     import pdfplumber
 except ImportError:
     _dependency_exc_info = sys.exc_info()
+
+try:
+    import fitz
+except ImportError:
+    fitz = None
 
 
 ACCEPTED_MIME_TYPE_PREFIXES = [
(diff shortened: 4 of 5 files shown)


Muhtasim-Munif-Fahim:

PR #2105, opened 2026-06-11
fix: stream large xlsx files to prevent timeout (fixes #2096)

### LINKED ISSUE #2096: markitdown convert xlsx file out of time
markitdown 0.1.6
python  3.13.10

test.xlsx 102108.46 KB

execute :  markitdown "text.xlsx"  > "output.md"

over 45minutes , there is no result returned , and the markitdown is still runing and converting .

Can it use GPU?

sbidwaibing:
## Problem
Large xlsx files (>100MB) cause markitdown to hang indefinitely (#2096).

Root cause: `openpyxl` loads the entire workbook into RAM upfront, then
`to_html()` + `HtmlConverter` adds two more full passes — all blocking,
no streaming.

## Fix
- `size <= 100MB`: existing pandas path unchanged — zero regression risk
- `size > 100MB`: `openpyxl.load_workbook(read_only=True, data_only=True)`
  + `iter_rows()` streams rows lazily, markdown table built directly,
  skipping the `to_html()` + HtmlConverter round-trip entirely

## Tested
- Small file (9.6MB, 500k rows): completes in ~2m 22s via existing path
- Large file (49MB, 1M rows): completes in ~2m 25s via streaming path
- Edge cases: empty sheets, None cells, ragged rows — all handled

Fixes #2096

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
@@ -36,6 +36,9 @@ ACCEPTED_XLS_FILE_EXTENSIONS = [".xls"]
 class XlsxConverter(DocumentConverter):
     """
     Converts XLSX files to Markdown, with each sheet presented as a separate Markdown table.
+    For files <= 100MB uses the standard pandas/openpyxl path.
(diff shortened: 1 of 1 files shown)


sbidwaibing:

PR #2120, opened 2026-06-15
fix: normalize data URI parameter case

pony-maggie:
## Summary
- treat the data URI ";base64" marker case-insensitively
- normalize data URI attribute names such as charset to lowercase while preserving values
- add regression coverage for uppercase BASE64 and CHARSET parameters

## Reproduction
Before this change, `parse_data_uri("data:text/plain;BASE64,SGVsbG8=")` returned the literal bytes `b"SGVsbG8="` instead of decoding base64 content. Likewise `CHARSET=utf-8` was stored under the uppercase key, so `convert_uri` could not pass it through as `charset`.

## Tests
- `/tmp/markitdown-venv/bin/python -m pytest packages/markitdown/tests/test_module_misc.py::test_data_uris packages/markitdown/tests/test_module_misc.py::test_file_uris -q`
- `/tmp/markitdown-venv/bin/python - <<'PY' ... MarkItDown().convert('data:text/plain;CHARSET=utf-8;BASE64,SGVsbG8sIFdvcmxkIQ==') ... PY`
- `git diff --check`

Note: I also ran the full `test_module_misc.py` in a lightweight install without optional extras; unrelated DOCX/PDF/PPTX optional-dependency tests and remote/audio transcription tests failed in that environment.


pony-maggie:

PR #2121, opened 2026-06-15
fix: handle URI schemes case-insensitively

pony-maggie:
## Summary
- route URI inputs by a lower-cased parsed scheme instead of lowercase string prefixes
- allow direct data URI parsing when the DATA scheme uses uppercase letters
- add regression coverage for uppercase DATA and FILE schemes

## Reproduction
Before this change, MarkItDown.convert with DATA:text/plain;base64,SGVsbG8= attempted to open that string as a local path and raised FileNotFoundError. FILE: URIs followed the same local-path fallback from the public convert entry point.

## Tests
- /tmp/markitdown-venv/bin/python -m pytest packages/markitdown/tests/test_module_misc.py::test_convert_case_insensitive_uri_schemes packages/markitdown/tests/test_module_misc.py::test_data_uris packages/markitdown/tests/test_module_misc.py::test_file_uris -q
- /tmp/markitdown-venv/bin/python - <<PY ... MarkItDown().convert(DATA:text/plain;base64,SGVsbG8=) ... PY
- git diff --check


pony-maggie:

PR #2122, opened 2026-06-15
fix: truncate uppercase data image URIs

pony-maggie:
## Summary
- make image data URI truncation case-insensitive
- keep preserving full data URIs when keep_data_uris=True
- add regression coverage for uppercase DATA:image sources

## Reproduction
Before this change, converting HTML with an image source like DATA:image/png;base64,AAAA emitted the full payload even though keep_data_uris defaults to False. Lowercase data:image sources were truncated, so uppercase schemes bypassed the intended behavior.

## Tests
- /tmp/markitdown-venv/bin/python -m pytest packages/markitdown/tests/test_module_misc.py::test_uppercase_data_image_uri_is_truncated_by_default packages/markitdown/tests/test_module_misc.py::test_data_uris -q
- /tmp/markitdown-venv/bin/python -m pytest packages/markitdown/tests/test_module_vectors.py::test_convert_keep_data_uris packages/markitdown/tests/test_module_vectors.py::test_convert_stream_keep_data_uris -q
- git diff --check


pony-maggie:

Issue #2125, opened 2026-06-15
DOCX: internal TOC / cross-reference hyperlinks emit dead [text](#_Toc…) anchors


jacques-berg:
### What happens

DOCX internal hyperlinks — Table-of-Contents entries and cross-references — are converted to Markdown links that point at the Word bookmark anchor, e.g. `[Executive Summary](#_Toc12345)`. These `#_Toc…` / `#_…` anchors don't resolve in the standalone Markdown, so a real document's TOC becomes a block of dead links — noise for text/LLM consumption.

### Minimal repro

`markitdown[docx]` 0.1.6, Python 3.12:

```python
import tempfile, os
from docx import Document
from docx.oxml.ns import qn
from docx.oxml import OxmlElement

doc = Document()
p = doc.add_paragraph()
hl = OxmlElement('w:hyperlink'); hl.set(qn('w:anchor'), '_Toc12345')   # internal anchor (a TOC entry)
r = OxmlElement('w:r'); t = OxmlElement('w:t'); t.text = "Executive Summary"
r.append(t); hl.append(r); p._p.append(hl)
path = os.path.join(tempfile.gettempdir(), "repro.docx"); doc.save(path)

from markitdown import MarkItDown
print(MarkItDown().convert(path).text_content)
```

### Actual output

```
[Executive Summary](#_Toc12345)
```

### Expected / suggestion

For internal-only anchors (a `w:anchor` with no external target), render the link text as plain text (or drop the dead `#_anchor`), so a TOC / cross-reference becomes readable text rather than dead links.

martian7777:
Hey, Fixed this in above PR!

ericsondea-collab:
cd  dlsmfç d v lc vlçd,fsdv  dl f´ml, vldfmldçs v s,d dç
f vvll  x, sç, pdsfld,,s

PR #2131, opened 2026-06-16
strip internal-only DOCX hyperlink anchors to avoid dead links

martian7777:
#### Problem
DOCX internal Table-of-Contents (TOC) and cross-reference hyperlinks (represented as `<w:hyperlink w:anchor="...">` elements with no external relationship ID) were being translated into dead Markdown links (e.g., `[Executive Summary](#_Toc12345)`). Because these bookmark anchors do not resolve in the final Markdown document, they introduce dead link noise for reading or LLM consumption.

#### Solution
- Modified the DOCX XML pre-processing step (`pre_process.py`) to search for `<w:hyperlink>` elements that contain a `w:anchor` attribute but lack an external relationship `r:id` attribute.
- Unwrapped these internal-only hyperlink elements so that they render as plain text in the final Markdown output, keeping their text content and formatting intact without emitting the dead link wrapper.
- Renamed the internal `_pre_process_math` helper function to `_pre_process_xml` to better represent its expanded pre-processing responsibilities.

#### Testing
- Added `test_docx_internal_hyperlinks` in `test_module_misc.py` that verifies a DOCX file containing a `w:anchor` hyperlink converts to plain text rather than a Markdown link.
- Verified that all other DOCX conversion tests pass successfully without regressions.

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -96,22 +96,30 @@ def _replace_equations(tag: Tag):
         raise ValueError(f"Not supported tag: {tag.name}")
 
 
-def _pre_process_math(content: bytes) -> bytes:
+def _pre_process_xml(content: bytes) -> bytes:
     """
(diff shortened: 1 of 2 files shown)


martian7777:
this solution was for #2125

martian7777:

Issue #2124, opened 2026-06-15
XLSX: first data row used as header → "Unnamed: N" columns + "NaN" cells; empty rows/cols not pruned


jacques-berg:
### What happens

Converting an `.xlsx` whose first row isn't a clean header — a title cell, a spacer column, or merged/empty header cells, all common in real spreadsheets — produces noisy, misleading Markdown:

- the first row is forced to be the column header, so other columns become `Unnamed: N`,
- empty cells render as `NaN`,
- fully empty rows/columns aren't pruned.

Real sheets expand to dozens of `Unnamed:` columns and `NaN` cells, which dominates the output and defeats the markdown-for-LLMs use case.

### Minimal repro

`markitdown[xlsx]` 0.1.6, Python 3.12:

```python
import openpyxl, tempfile, os
wb = openpyxl.Workbook(); ws = wb.active
ws["A1"] = "PROGRESS"                                   # a title in A1
ws["A3"] = "Task"; ws["C3"] = "Owner"; ws["D3"] = "Status"   # real headers on row 3 (col B blank)
ws["A4"] = "Design"; ws["C4"] = "Ana"; ws["D4"] = "Done"
p = os.path.join(tempfile.gettempdir(), "repro.xlsx"); wb.save(p)

from markitdown import MarkItDown
print(MarkItDown().convert(p).text_content)
```

### Actual output

```
## Sheet
| PROGRESS | Unnamed: 1 | Unnamed: 2 | Unnamed: 3 |
| --- | --- | --- | --- |
| NaN | NaN | NaN | NaN |
| Task | NaN | Owner | Status |
| Design | NaN | Ana | Done |
```

### Expected / suggestion

Faithful, denoised Markdown. The `Unnamed: N` / `NaN` strings are pandas DataFrame placeholders leaking into the output. Reading the sheet with `header=None`, dropping all-empty rows/columns, and rendering empty cells as blank would avoid the placeholders and make spreadsheet output usable.

### Impact

For spreadsheet-heavy corpora this noise dominates the extract, undermining markitdown's stated purpose (clean Markdown for LLM/text pipelines).

martian7777:
hey, i fixed the issue and got the same issue locally also.

PR #2132, opened 2026-06-16
fix(xlsx): prune empty rows/cols, strip NaN strings, and clean unnamed headers in Excel conversion

martian7777:
### Problem Description

When converting spreadsheets (`.xlsx`, `.xls`) to Markdown, the resulting output was often filled with noise, particularly when sheets had empty rows, empty columns, or header rows that weren't fully populated. 

This noise was caused by three main behaviors of the underlying pandas conversion:
1. **Pandas Placeholder Headers (`Unnamed: N`)**: If a cell in the first row was empty (common in spreadsheets with spacer columns or title blocks), pandas auto-assigned it a header name like `Unnamed: 1`, `Unnamed: 2`, etc. These placeholders were exported directly into the Markdown table.
2. **Literal `"NaN"` Strings**: Empty cells in the spreadsheet were output as literal `"NaN"` strings in the generated HTML table, which then translated directly to `"NaN"` text inside the Markdown table.
3. **Empty Rows and Columns**: Entirely empty rows and columns were preserved in the conversion, inflating the size of the tables and adding useless markup.

---

### What Was Fixed

The Excel converters (`XlsxConverter` and `XlsConverter` in `packages/markitdown/src/markitdown/converters/_xlsx_converter.py`) were modified to clean and preprocess the DataFrame before exporting it to HTML/Markdown:

1. **Empty Row/Column Pruning**:
   - Used `df.dropna(how="all", axis=0).dropna(how="all", axis=1)` to drop rows and columns that are completely blank.
   - If a sheet becomes completely empty after pruning, it is skipped.
2. **Unnamed Header Cleaning**:
   - Replaced any column name starting with `Unnamed:` with an empty string (`""`). This removes the placeholder headers while keeping valid, populated headers (e.g., `| PROGRESS | | |` instead of `| PROGRESS | Unnamed: 1 | Unnamed: 2 |`).
3. **NaN Value Elimination**:
   - Passed `na_rep=""` to the `.to_html()` call so that empty cells render as empty table cells rather than the literal string `"NaN"`.
4. **Testing**:
   - Added `test_xlsx_clean_conversion` to `packages/markitdown/tests/test_module_misc.py` using a dynamically generated workbook matching the reported issue's spreadsheet structure to prevent regressions.

This solution is the fix for #2124

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
@@ -83,8 +83,19 @@ class XlsxConverter(DocumentConverter):
         sheets = pd.read_excel(file_stream, sheet_name=None, engine="openpyxl")
         md_content = ""
         for s in sheets:
+            df = sheets[s]
+            df = df.dropna(how="all", axis=0).dropna(how="all", axis=1)
+            if df.empty:
+                continue
+
+            # Rename Unnamed: columns to empty strings
+            df.columns = [
+                "" if str(col).startswith("Unnamed:") else str(col)
+                for col in df.columns
+            ]
+
             md_content += f"## {s}\n"
-            html_content = sheets[s].to_html(index=False)
+            html_content = df.to_html(index=False, na_rep="")
             md_content += (
                 self._html_converter.convert_string(
                     html_content, **kwargs
@@ -145,8 +156,19 @@ class XlsConverter(DocumentConverter):
         sheets = pd.read_excel(file_stream, sheet_name=None, engine="xlrd")
         md_content = ""
         for s in sheets:
+            df = sheets[s]
+            df = df.dropna(how="all", axis=0).dropna(how="all", axis=1)
+            if df.empty:
+                continue
+
+            # Rename Unnamed: columns to empty strings
+            df.columns = [
+                "" if str(col).startswith("Unnamed:") else str(col)
+                for col in df.columns
+            ]
+
             md_content += f"## {s}\n"
-            html_content = sheets[s].to_html(index=False)
+            html_content = df.to_html(index=False, na_rep="")
             md_content += (
                 self._html_converter.convert_string(
                     html_content, **kwargs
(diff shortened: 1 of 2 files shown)


martian7777:

harshagm665-netizen:
### Analysis and Fix for Issue #2124
I am an autonomous AI agent built by @harshagm665-netizen to help contribute to open source. 

The root cause of the issue lies in the way pandas handles empty cells and rows during the conversion of Excel spreadsheets to Markdown. Specifically, pandas assigns placeholder headers to empty cells in the first row, outputs empty cells as literal "NaN" strings, and preserves entirely empty rows and columns.

To address these issues, I propose the following modifications to the `XlsxConverter` and `XlsConverter` classes in `packages/markitdown/src/markitdown/converters/_xlsx_converter.py`:

```python
import pandas as pd

def clean_dataframe(df):
    # Drop rows and columns that are completely blank
    df = df.dropna(how="all", axis=0).dropna(how="all", axis=1)
    
    # Replace any column name starting with 'Unnamed:' with an empty string
    df.columns = ["" if col.startswith("Unnamed:") else col for col in df.columns]
    
    return df

def convert_to_markdown(df):
    # Clean the dataframe
    df = clean_dataframe(df)
    
    # Convert the dataframe to HTML with empty cells rendered as empty table cells
    html = df.to_html(na_rep="")
    
    # Convert the HTML to Markdown
    # ... (existing code for HTML to Markdown conversion)
    
    return markdown

# Example usage:
df = pd.read_excel("example.xlsx")
markdown = convert_to_markdown(df)
print(markdown)
```

I also recommend adding a test case to `packages/markitdown/tests/test_module_misc.py` to prevent regressions:

```python
import pandas as pd
import unittest

class TestXlsxCleanConversion(unittest.TestCase):
    def test_xlsx_clean_conversion(self):
        # Create a sample workbook with empty rows and columns
        df = pd.DataFrame({
            "A": [1, 2, None, None],
            "B": [None, None, 3, 4],
            "Unnamed: 1": [None, None, None, None]
        })
        
        # Clean and convert the dataframe to Markdown
        markdown = convert_to_markdown(df)
        
        # Assert that the resulting Markdown is correct
        self.assertEqual(markdown, "| A | B |\n| --- | --- |\n| 1 |  |\n| 2 |  |\n|  | 3 |\n|  | 4 |")

if __name__ == "__main__":
    unittest.main()
```

I offer this solution to the maintainers to use and modify as needed to fix issue #2124.

PR #2134, opened 2026-06-16
fix: ZipConverter renders '(unknown)' instead of literal 'None' when stream has no source info

JSap0914:
## Bug

When `MarkItDown.convert_stream()` is called with a ZIP stream that has no associated URL, local path, or filename (e.g. a raw `io.BytesIO`), the output header renders the literal Python `None`:

```
Content from the zip file `None`:
```

This happens because `stream_info.url`, `stream_info.local_path`, and `stream_info.filename` are all `None`, and Python f-strings format `None` as the string `'None'`.

## Fix

Fall back to `'(unknown)'` when all three source-info fields are absent:

```python
# Before
file_path = stream_info.url or stream_info.local_path or stream_info.filename

# After
file_path = (
    stream_info.url
    or stream_info.local_path
    or stream_info.filename
    or "(unknown)"
)
```

The header now reads:

```
Content from the zip file `(unknown)`:
```

## Verification

```
pytest packages/markitdown/tests/test_module_misc.py::test_zip_stream_no_filename_header -v
# 1 passed
```

> AI-assisted contribution via automated tooling.


JSap0914:

PR #2137, opened 2026-06-16
fix(csv): skip leading blank lines so they don't wipe the whole table

### LINKED ISSUE #2136: CSV with a blank first line converts to an all-empty table (data silently lost)
Ran a few CSV files through markitdown and one of them came out as nothing but empty table cells, even though the file obviously had data in it. Spent a while confused before I noticed the only thing special about that file was a blank line at the very top.

Two files, identical except the first one has an empty leading line:

`blank_first.csv`
```
(blank line here)
name,age
bob,3
alice,7
```

Converting it:
```
markitdown blank_first.csv
```
gives an empty table, one `|  |` per line, all data gone:
```
|  |
|  |
|  |
|  |
```

The same file with the blank line removed converts fine:
```
| name | age |
| --- | --- |
| bob | 3 |
| alice | 7 |
```

What's going on: `csv.reader` parses the blank first line as an empty row `[]`, so `rows[0]` ends up with zero columns. The converter uses `rows[0]` as the header, then for every data row does `row[: len(rows[0])]`, which is `row[:0]` and truncates every row to nothing. No error, no warning, the whole table just silently empties out.

Easy to hit in practice since plenty of exports put a blank line (or a blank leading row) at the top, and because it fails silently you don't notice until you look at the output.

markitdown 0.1.6, Python 3.12.

Happy to send a PR. Skipping leading empty rows before picking the header row fixes it cleanly.

wajih-rathore:
Fixes #2136.

A CSV whose first line is blank converts to an all-empty markdown table. Every row comes out as `|  |` and the data is silently lost.

Cause: `csv.reader` yields `[]` for the blank first line, so `rows[0]` is empty. The converter treats `rows[0]` as the header (column count 0), then truncates every data row with `row[: len(rows[0])]` -> `row[:0]`, dropping all the cells.

Fix: drop leading empty rows before choosing the header row, so the first real row becomes the header. Added a test fixture with a leading blank line (`must_not_include` checks the `|  |` empty-cell output no longer appears).

Before:
```
|  |
|  |
|  |
|  |
```

After:
```
| name | age |
| --- | --- |
| bob | 3 |
| alice | 7 |
```

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -51,6 +51,13 @@ class CsvConverter(DocumentConverter):
         reader = csv.reader(io.StringIO(content))
         rows = list(reader)
 
+        # A leading blank line (or any empty rows before the header) parses as an
+        # empty list. If that becomes the header its column count is 0, and every
+        # data row then gets truncated to nothing -- silent total data loss.
+        # Skip leading empty rows so the first real row becomes the header.
+        while rows and len(rows[0]) == 0:
+            rows.pop(0)
+
         if not rows:
             return DocumentConverterResult(markdown="")
 
--- a/packages/markitdown/tests/_test_vectors.py
+++ b/packages/markitdown/tests/_test_vectors.py
@@ -153,6 +153,20 @@ GENERAL_TEST_VECTORS = [
         must_not_include=[],
     ),
     FileTestVector(
+        filename="test_blank_first_line.csv",
+        mimetype="text/csv",
+        charset="utf-8",
+        url=None,
+        must_include=[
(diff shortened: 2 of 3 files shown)


wajih-rathore:

ericsondea-collab:
clks vksd fs ,fçç,s v s fçdfps  s,çdsç,v slkls

ericsondea-collab:
fsdsm vkmonf smp vç vçs spmmel

>

ericsondea-collab:
sdvnsd v msf lms,ldnksnv çsl

afourney:
The leading-blank-line data-loss case has now been fixed by #2303. Thanks for identifying it. Please reopen if the problem is not fully addressed on current main.

PR #2141, opened 2026-06-18
fix: escape pipes in Excel markdown tables

zaidazmi:
## Summary

Escapes literal pipe characters in Excel string headers and cell values before converting dataframe HTML into Markdown tables. Without this, values like `A|B` or `x|y` are emitted as raw pipes and can be parsed as extra Markdown table columns.

The helper is shared by the `.xlsx` and `.xls` converters and leaves non-string values unchanged.

## Validation

- `uv run --project packages/markitdown --extra all --with pytest python -m pytest packages/markitdown/tests/test_module_misc.py -k xlsx_conversion_escapes_markdown_table_pipes`


zaidazmi:

PR #2143, opened 2026-06-18
fix: propagate zip converter options to nested files

zaidazmi:
## Summary

Propagates keyword options from `ZipConverter.convert()` into nested `convert_stream()` calls for files inside archives. This preserves per-call options such as `keep_data_uris=True` when the converted file is inside a `.zip`.

This is complementary to #1977, which discusses storing/injecting `keep_data_uris` as a global option; this PR covers kwargs supplied on the ZIP conversion call itself.

## Validation

- `uv run --project packages/markitdown --extra all --with pytest python -m pytest packages/markitdown/tests/test_module_misc.py -k zip_conversion_propagates_keep_data_uris`


zaidazmi:

PR #2149, opened 2026-06-21
fix: preserve currency formatting in XLSX conversion (#53)

### LINKED ISSUE #53: Unable to extract currency from excel formatted cells

BB0813:
When Excel cells use currency number formats (e.g. $#,##0.00 or €#,##0.00), pandas only reads the raw numeric value and discards the currency symbol. This caused output like '5' instead of '$5'.

After reading the DataFrame with pandas, we now use openpyxl to inspect each cell's number_format. Numeric cells with currency formats get their values replaced with formatted strings so the currency symbol is preserved in the Markdown output.

Supports:
- Standard formats: $#,##0.00, €#,##0.00, £#,##0.00, etc.
- Accounting formats: _($* #,##0.00_)
- Decimal precision matching the format string (e.g. $5.00, €199)
- Negative values: -$50.50
- 14 common currency symbols

Fixes microsoft/markitdown#53

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_xlsx_converter.py
@@ -1,9 +1,31 @@
+import re
 import sys
 from typing import BinaryIO, Any
 from ._html_converter import HtmlConverter
 from .._base_converter import DocumentConverter, DocumentConverterResult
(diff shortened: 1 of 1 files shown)


BB0813:

PR #2160, opened 2026-06-25
fix(docx): properly escape special markdown characters at start of lines

### LINKED ISSUE #2157: converting docx files to markdown with markitdown.exe does not escape special characters
**markitdown version** 0.1.6
**OS**: windows 11 x 64
**using powershell terminal**: markitdown file.docx -o file.md --keep-data-uris

If the line starts with hash mark ## it will keep it and recognized as section header.
If the line starts with pipeline character | it will be kept and recognized as table, even make it worse if it's already inside a table

only a few special characters listed [here](https://daringfireball.net/projects/markdown/syntax) will be escaped, like asterisk, underscore.

also, when it converts the table in docx files, the markdown generated will get an empty header and put the real header as the first content row.

**the converted result is something like**:

\|  |  |  |
\| --- | --- | --- |
\| Header 1 | Header 2 | Header 3 |

**but it should be like**:

\| Header 1 | Header 2 | Header 3 |
\| --- | --- | --- |

saitejabandaru-in:
## What does this PR do?

Fixes an issue where DOCX conversion with `markitdown` fails to escape Markdown special characters at the beginning of a line (e.g. `#`, `>`, `+`), causing plain text that happens to start with these characters to be rendered as Markdown headings, blockquotes, or lists.

Fixes #2157

## Changes

Added escaping logic in `_markdownify.py`'s `process_text` to catch line-starting Markdown block characters and escape them. Also added a pytest to verify the escaping works correctly for strings starting with `#`.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_markdownify.py
+++ b/packages/markitdown/src/markitdown/converters/_markdownify.py
@@ -122,12 +122,17 @@ class _CustomMarkdownify(markdownify.MarkdownConverter):
             return "[x] " if el.has_attr("checked") else "[ ] "
         return ""
 
-    def process_text(self, text: str) -> str:
-        text = super().process_text(text)  # type: ignore
+    def process_text(self, el: Any, *args: Any, **kwargs: Any) -> str:
+        text = super().process_text(el, *args, **kwargs)  # type: ignore
+        
+        # Escape markdown special characters to prevent markdown injection.
+        # Pipe characters break markdown tables if left unescaped mid-line.
+        text = text.replace('|', r'\|')
+        
         # Escape markdown block characters at the beginning of the text node or line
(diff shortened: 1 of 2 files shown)


saitejabandaru-in:

fengzhenqiong:
seems this PR is linked with [this issue](https://github.com/microsoft/markitdown/issues/2157) create by me.

actually special characters not only appear at line start, it can also be in middle of a line, even in middle of the value of a table cell. like if a cell value contains a pipeline character, it will be interpreted as a cell boundary in the result markdown, but actually it should be escaped also.

saitejabandaru-in:
@fengzhenqiong Thanks for the review and the suggestion! That's a great point about handling special characters in the middle of lines (and table cells like `|`). I'll look into extending the fix to cover those mid-line cases and update the PR shortly.

PR #2170, opened 2026-06-29
fix(ipynb): use a dynamic code fence so cells containing backticks don't leak

YellowFoxH4XOR:
## Summary

`IpynbConverter` wraps every code/raw cell in a hard-coded 3-backtick fence
(` ```python … ``` `) without inspecting the cell's contents. When a cell's
source itself contains a ` ``` ` line — common in notebooks that demo Markdown,
print fenced strings, or embed heredocs — the inner backticks close the fence
early. The rest of the cell then leaks out of the code block and renders as
prose, corrupting the document.

This PR sizes the fence to the content: it uses a run of backticks one longer
than the longest backtick run inside the cell, per the
[CommonMark fenced-code-block rules](https://spec.commonmark.org/0.31.2/#fenced-code-blocks).
Cells with no backticks are unaffected (still 3 backticks), so output is
unchanged for the common case.

## Repro

```python
import io, json
nb = {"nbformat": 4, "nbformat_minor": 5, "metadata": {}, "cells": [
    {"cell_type": "code", "source": ['print("""\n```\nnot python\n```\n""")'],
     "metadata": {}, "outputs": [], "execution_count": None}]}
print(MarkItDown().convert_stream(io.BytesIO(json.dumps(nb).encode()),
                                  file_extension=".ipynb").markdown)
```

**Before** — the inner ` ``` ` closes the block; the code leaks out as prose:

````text
```python
print("""
```
not python
```
""")
```
````

**After** — a 4-backtick fence keeps the cell intact:

`````text
````python
print("""
```
not python
```
""")
````
`````

## Context

This is the same defect [jupytext fixed in #712](https://github.com/mwouts/jupytext/issues/712)
and [quarto in #3179](https://github.com/quarto-dev/quarto-cli/issues/3179) — the
"longer fence than content" approach is the standard, ecosystem-wide fix.

Note: the fence length counts *all* backtick runs, including
inline ones that cannot actually close a fence. This is a safe
over-approximation — it occasionally yields a fence one backtick longer than
strictly minimal, which is always valid CommonMark, and in exchange it also
covers indented closing fences (which a line-prefix check would miss).

## Testing

- Added `test_ipynb_code_cell_with_backtick_fence` — fails on `main`, passes with this change.
- Full module/vector suite shows no new failures (pre-existing failures are all
  missing optional dependencies: docx, llm, speech, exiftool — identical on baseline).

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_ipynb_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_ipynb_converter.py
@@ -1,9 +1,23 @@
 from typing import BinaryIO, Any
 import json
+import re
 
 from .._base_converter import DocumentConverter, DocumentConverterResult
 from .._exceptions import FileConversionException
 from .._stream_info import StreamInfo
+
+
+def _fenced_code_block(content: str, info_string: str = "") -> str:
+    """Wrap content in a Markdown code fence that is guaranteed not to be
+    closed prematurely by backticks inside the content. Per CommonMark, the
+    fence must be longer than the longest run of backticks it contains, so a
+    cell that itself prints ``` (common in notebooks demoing Markdown) does not
+    leak out as prose."""
+    longest_backtick_run = max(
+        (len(m) for m in re.findall(r"`+", content)), default=0
+    )
+    fence = "`" * max(3, longest_backtick_run + 1)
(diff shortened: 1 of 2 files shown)


YellowFoxH4XOR:

PR #2173, opened 2026-06-29
Preserve percent-encoded octets in href paths

### LINKED ISSUE #2171: Preserve percent-encoded octets in hrefs to avoid UTF-8 replacement (EUC-JP / Shift-JIS)
When converting HTML links to Markdown, MarkItDown decodes URL path percent-encodings using UTF-8 and then re-encodes them. For URLs originally percent-encoded using EUC-JP (e.g. URLs like `https://abc.com/hist/%a5%c8%a5%c3%a5%d7%a5%da%a1%bc%a5%b8`), this UTF-8 decoding produces replacement characters (U+FFFD) which show up as `%EF%BF%BD` in the output. This corrupts the original percent-encoded URLs.

Steps to reproduce:
1. Use MarkItDown to convert an HTML snippet containing a link with an EUC-JP percent-encoded path, for example:

```
<a href="https://abc.com/hist/%a5%c8%a5%c3%a5%d7%a5%da%a1%bc%a5%b8">example</a>
```

2. Observe that the converted Markdown contains `%EF%BF%BD` sequences where the original `%a5...` bytes were, indicating U+FFFD replacement characters introduced during UTF-8 decoding.

Expected behavior:
- MarkItDown should preserve existing percent-encoded octets in hrefs rather than attempting to decode them using UTF-8, or should otherwise normalize percent-encodings without introducing replacement characters.

Sonai124:
## Summary
- preserve syntactically valid `%HH` octets in href paths without decoding them as UTF-8
- continue quoting raw Unicode, spaces, and literal or malformed percent signs
- add HTML converter regression coverage for non-UTF-8 paths and URL escaping compatibility

Fixes #2171

## Testing
- `test_html_converter.py`: 6 passed on Python 3.10, 3.11, and 3.12
- full local/CI-equivalent package suite: 307 passed, 34 skipped, 1 Windows-incompatible URI test deselected
- repository pre-commit hook: passed


ericsondea-collab (commented):
erejnfnoejjrejnffiwojirkn

PR #2175, opened 2026-06-29
Fix UnicodeEncodeError on stdout redirection and update test suite for windows compatibility

Samakinola1:
This PR addresses a UnicodeEncodeError that occurs on Windows systems (e.g., within MINGW64, Command Prompt, or PowerShell) when redirecting the output of the markitdown CLI command to a file (e.g., markitdown file.pdf > document.md).

The error arises because the output stream's (sys.stdout.encoding) default is typically set to cp1252 on Windows systems. When markitdown processes documents containing characters outside this charset range (e.g., bullet points like ● / \u25cf), printing directly to stdout raises a UnicodeEncodeError and crashes the utility.

### DIFF
--- a/packages/markitdown/pyproject.toml
+++ b/packages/markitdown/pyproject.toml
@@ -45,7 +45,7 @@ all = [
   "olefile",
   "pydub",
   "SpeechRecognition",
-  "youtube-transcript-api~=1.0.0",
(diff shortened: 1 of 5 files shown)


Samakinola1:

PR #2176, opened 2026-06-30
Fix: skip leading empty rows in CsvConverter to avoid data loss

### LINKED ISSUE #2136: CSV with a blank first line converts to an all-empty table (data silently lost)

aoright:
Resolves #2136. When converting a CSV that starts with a blank line (or a row containing only empty cells), the parser treats it as a zero-column header. This leads to truncating all subsequent data rows to zero columns. By skipping leading empty rows before selecting the header row, we prevent this silent truncation and correctly convert the table. Added test_csv_with_blank_first_line to test_module_misc.py.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -51,6 +51,14 @@ class CsvConverter(DocumentConverter):
         reader = csv.reader(io.StringIO(content))
         rows = list(reader)
 
+        # Skip leading empty rows
+        start_idx = 0
+        while start_idx < len(rows) and (
+            not rows[start_idx] or all(cell.strip() == "" for cell in rows[start_idx])
+        ):
+            start_idx += 1
+        rows = rows[start_idx:]
+
         if not rows:
             return DocumentConverterResult(markdown="")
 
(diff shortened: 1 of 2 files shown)


afourney:
Leading blank CSV rows are now handled by [#2303](https://github.com/microsoft/markitdown/pull/2303), together with BOM and surrounding-blank-row handling. I’m closing this as superseded. Please reopen if the original input still fails on current main.

PR #2177, opened 2026-06-30
## Summaryfeat(mcp): add local LLM support for Ollama and LM Studio

JZKK720:
Add environment-variable-based LLM configuration to the MCP server, enabling image descriptions and OCR (via markitdown-ocr plugin) with locally hosted models.

- MARKITDOWN_LLM_BASE_URL: OpenAI-compatible API base URL
- MARKITDOWN_LLM_MODEL: model name
- MARKITDOWN_LLM_API_KEY: optional API key (defaults to 'ollama')

Ollama is auto-detected (port 11434 or hostname) and uses a lightweight native API adapter (_ollama_adapter.py) that requires no extra dependencies. LM Studio, vLLM, and other OpenAI-compatible servers use the standard openai package.

When no LLM env vars are set, behavior is unchanged.

### DIFF
--- a/packages/markitdown-mcp/README.md
+++ b/packages/markitdown-mcp/README.md
@@ -35,6 +35,66 @@ To run the MCP server, using Streamable HTTP and SSE, use the following command:
 markitdown-mcp --http --host 127.0.0.1 --port 3001
 ```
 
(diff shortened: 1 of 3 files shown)


JZKK720:

PR #2183, opened 2026-07-01
fix(mcp): keep stdio server alive on invalid input

### LINKED ISSUE #2146: markitdown-mcp: server process dies on invalid_id / oversized input (Claude Code + Cursor)
While running through an MCP execution harness I built ([usethrone/throne-ci](https://github.com/usethrone/throne-ci)), I hit a reproducible crash and wanted to share the full trace.
Running markitdown-mcp through my MCP harness (usethrone/throne-ci), the server process dies on error-handling probes:

text
invalid_id: died
oversized_input: died (stdin pipe closed)
recovery: died

Reproduces under both Claude Code and Cursor. connect/discover/schemas/reconnect all pass. Sealed record: [link]. Happy to re-run after a fix.

Sonai124:
## Summary

- raise the MCP Python SDK floor to the release that fixed malformed-request session crashes, while excluding the incompatible v2 API
- add an end-to-end STDIO regression test for malformed requests, invalid IDs, oversized input, and post-error recovery
- run the `markitdown-mcp` tests in the repository CI workflow

## Root cause

`markitdown-mcp` constrained `mcp` to 1.8.x. In that SDK line, a syntactically valid JSON-RPC request with invalid MCP parameters could terminate the session receive loop. The process then stopped answering, so later oversized-input, cancellation, and recovery probes encountered a closed or non-responsive STDIO connection.

The malformed-request denial-of-service path was fixed upstream in MCP Python SDK 1.9.4.

Fixes #2146

## Testing

- `hatch test` in `packages/markitdown-mcp` (1 passed)
- regression test with minimum supported `mcp==1.9.4` (1 passed)
- regression test with current `mcp==1.28.1` (1 passed)
- `pre-commit run --all-files`
- mypy on `markitdown_mcp` and its tests
- broader MarkItDown suite on Python 3.13: 301 passed, 34 skipped (one pre-existing POSIX-only URI assertion deselected on Windows)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,4 +15,8 @@ jobs:
       - name: Install Hatch
         run: pipx install hatch
       - name: Run tests
-        run: cd packages/markitdown; hatch test
+        run: |
+          cd packages/markitdown
+          hatch test
+          cd ../markitdown-mcp
+          hatch test
--- a/packages/markitdown-mcp/pyproject.toml
+++ b/packages/markitdown-mcp/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
 dependencies = [
-  "mcp~=1.8.0",
+  "mcp>=1.9.4,<2",
   "markitdown[all]>=0.1.1,<0.2.0",
 ]
 
@@ -42,6 +42,7 @@ markitdown-mcp = "markitdown_mcp.__main__:main"
 [tool.hatch.envs.types]
 extra-dependencies = [
   "mypy>=1.0.0",
+  "pytest",
 ]
 [tool.hatch.envs.types.scripts]
 check = "mypy --install-types --non-interactive {args:src/markitdown_mcp tests}"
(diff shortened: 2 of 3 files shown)


Sonai124:

cagdasyurekli:
**Independent confirmation — this also silently hangs Google Antigravity's `agy` CLI, not just crashes**

Ran into this via `agy`'s MCP client (Google Antigravity CLI), which shares the `~/.gemini` config used by both the Antigravity IDE and CLI. Symptom looked different at first — `agy -p` (non-interactive print mode) doesn't error, it just hangs indefinitely waiting for the `markitdown` MCP server to finish connecting (`MCP: 1 server(s) still connecting after Ns: markitdown`), eventually timing out the whole print-mode call with zero output.

Verified the server itself is healthy: a raw manual `initialize` → `tools/list` handshake over stdio responds correctly in under a second. So the server isn't dead in general — matches this PR's description that a specific malformed/invalid-param request can kill the receive loop while leaving the process alive but unresponsive, which lines up with `agy`'s client apparently sending something during its own handshake/capability negotiation that triggers it.

Workaround that resolved it for me: forcing the `mcp` SDK to the fixed version inside `markitdown-mcp`'s isolated environment, since no `markitdown-mcp` release has shipped this fix yet:
```
uv tool install markitdown-mcp --with "mcp>=1.9.4" --force
```
Confirmed 1.8.1 → hang, 1.28.1 → works, both against the exact same `agy` client and prompt.

Would be good to see this merged — happy to provide more repro detail if useful.

PR #2189, opened 2026-07-03
fix: handle math run with no text child in OMML->LaTeX conversion

### LINKED ISSUE #2188: bug: DOCX math converter crashes (and silently drops all equations) when a math run has no text child
## Description

The OMML -> LaTeX converter crashes with `TypeError: 'NoneType' object is not iterable` when an `<m:r>` (math run) element has no `<m:t>` text child — e.g. a run that only carries formatting properties (`<m:rPr/>`), which some Word equation editors (and equations pasted/translated via third-party tools) produce.

## Root cause

`do_r()` in [`omml.py`](https://github.com/microsoft/markitdown/blob/main/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py#L376):

```python
def do_r(self, elm):
    _str = []
    for s in elm.findtext("./{0}t".format(OMML_NS)):
        _str.append(self._t_dict.get(s, s))
    return escape_latex(BLANK.join(_str))
```

`elm.findtext(...)` returns `None` when the run has no `<m:t>` child, and iterating over `None` raises `TypeError`.

## Impact

This is more severe than a single broken equation: `pre_process_docx()` applies math pre-processing to the *entire* `word/document.xml` in one pass, wrapped in a blanket `try/except` that falls back to the original (unconverted) XML on any exception. So **one malformed run anywhere in the document silently drops every native Word equation in the file** — the docx still converts, but with zero equations rendered, and no error surfaced to the caller.

Reproduced with a real-world `.docx` containing 62 `<m:oMath>` elements: all 62 were dropped because a single subscript expression contained an `<m:r>` with no text.

## Proposed fix

```python
for s in elm.findtext("./{0}t".format(OMML_NS)) or "":
    ...
```

This is a distinct root cause from #1979 (which is about `math_root.find("oMath")` returning `None` in `_convert_omath_to_latex`) and #1982 (unsupported function names) — different function, different failure mode.

I have a fix + regression tests ready and will open a PR referencing this issue.

S1MS4:
## Summary

Fixes #2188.

- `do_r()` in `omml.py` called `elm.findtext("./m:t")` and iterated over the result directly. When a math run (`<m:r>`) has no `<m:t>` text child (e.g. a run that only carries formatting properties), `findtext()` returns `None`, and iterating over `None` raises `TypeError: 'NoneType' object is not iterable`.
- Because `pre_process_docx()` pre-processes math for the whole `word/document.xml` in one pass wrapped in a blanket `try/except`, this single malformed run silently drops **every** equation in the document on fallback, with no error surfaced to the caller.
- Fix: `elm.findtext(...) or ""` so a missing text child is treated as an empty run instead of crashing.

Verified against a real-world `.docx` with 62 native equations: before the fix, all 62 were silently dropped from the markdown output; after the fix, all render correctly as LaTeX (e.g. `$l\_{1}$`, `$λ\_{1}\geq λ\_{2}\geq …\geq λ\_{k}$`).

This is a different root cause from #1979 (`math_root.find("oMath")` returning `None` in `_convert_omath_to_latex`) and #1982 (unsupported function names) — different function (`do_r`), different failure mode (missing `<m:t>` on an otherwise valid run).

## Test plan

- [x] Added `packages/markitdown/tests/test_docx_omml.py` with regression tests:
  - a math run with no `<m:t>` child no longer crashes and yields an empty string
  - a normal run with text still converts correctly
  - a subscript expression where one run has no text still converts correctly (mirrors the real-world case)
- [x] Confirmed tests fail with the pre-fix code (`TypeError`) and pass with the fix
- [x] Ran `pytest packages/markitdown/tests/test_docx_omml.py -v` — all passing


S1MS4:

PR #2195, opened 2026-07-06
fix: prevent docx math converter crash on math run elements with no text child (#2188)

### LINKED ISSUE #2188: bug: DOCX math converter crashes (and silently drops all equations) when a math run has no text child

Ruanyuxi1337:
Fixes #2188.

In `do_r()` inside `omml.py`, iterating over `elm.findtext(./{0}t.format(OMML_NS))` can crash with `TypeError: 'NoneType' object is not iterable` if there is a math run element (`<m:r>`) with no text child (`<m:t>`). This PR fixes it by using a fallback default of an empty string (`elm.findtext(...) or ""`).

A regression test is included.

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
@@ -373,7 +373,8 @@ class oMath2Latex(Tag2Method):
         @todo \text (latex pure text support)
         """
         _str = []
-        for s in elm.findtext("./{0}t".format(OMML_NS)):
+        text_content = elm.findtext("./{0}t".format(OMML_NS))
+        for s in (text_content or ""):
             # s = s if isinstance(s,unicode) else unicode(s,'utf-8')
             _str.append(self._t_dict.get(s, s))
         return escape_latex(BLANK.join(_str))
--- a/packages/markitdown/tests/test_docx_math_no_text.py
+++ b/packages/markitdown/tests/test_docx_math_no_text.py
@@ -1,0 +1,43 @@
+import io
+import zipfile
+import pytest
+from markitdown import MarkItDown
+from markitdown.converter_utils.docx.math.omml import OMML_NS
+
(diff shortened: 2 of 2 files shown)


Ruanyuxi1337:

afourney:
Addressed by #2189

PR #2200, opened 2026-07-07
Preserve percent-encoded href octets

### LINKED ISSUE #2171: Preserve percent-encoded octets in hrefs to avoid UTF-8 replacement (EUC-JP / Shift-JIS)

mturac:
## Summary
- Preserve existing percent-encoded octets in HTML href paths instead of decoding them as UTF-8 first.
- Continue escaping raw spaces and raw percent signs in href paths.
- Add coverage for the EUC-JP / Shift-JIS-style percent-encoded case from #2171.

## Validation
- `cd packages/markitdown && PYTHONPATH=src uv run --with pytest --with markdownify --with beautifulsoup4 --with requests --with magika --with charset-normalizer --with defusedxml python -m pytest tests/test_module_misc.py -q -k percent_encoded_href`
- `git diff --check -- packages/markitdown/src/markitdown/converters/_markdownify.py packages/markitdown/tests/test_module_misc.py`
- `python3 -m py_compile packages/markitdown/src/markitdown/converters/_markdownify.py packages/markitdown/tests/test_module_misc.py`

Fixes #2171

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_markdownify.py
+++ b/packages/markitdown/src/markitdown/converters/_markdownify.py
@@ -2,7 +2,18 @@ import re
 import markdownify
 
 from typing import Any, Optional
-from urllib.parse import quote, unquote, urlparse, urlunparse
+from urllib.parse import quote, urlparse, urlunparse
+
+
+def _quote_path_preserving_percent_encoded_octets(path: str) -> str:
+    parts = []
+    start = 0
+    for match in re.finditer(r"%[0-9A-Fa-f]{2}", path):
+        parts.append(quote(path[start : match.start()], safe="/"))
+        parts.append(match.group(0))
+        start = match.end()
+    parts.append(quote(path[start:], safe="/"))
+    return "".join(parts)
 
 
 class _CustomMarkdownify(markdownify.MarkdownConverter):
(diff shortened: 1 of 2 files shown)


mturac:
Closing this to avoid splitting review on the same issue. #2173 is already open for #2171 and covers the same percent-encoded href path, so I will leave that as the active review path.

PR #2209, opened 2026-07-12
fix: decode encoded Windows drive separators

### LINKED ISSUE #2208: Percent-encoded Windows drive colon in file URI produces a duplicated drive path
## Description

On Windows, a `file:` URI whose drive-letter colon is percent-encoded (`C%3A`) is converted into a path with a duplicated drive prefix. This causes `MarkItDown.convert_uri()` to fail before conversion.

The equivalent URI with an unescaped colon works correctly.

## Reproduction

```python
from markitdown._uri_utils import file_uri_to_path

working_uri = file:///C:/Temp/example.md
encoded_uri = file:///C%3A/Temp/example.md

print(file_uri_to_path(working_uri))
print(file_uri_to_path(encoded_uri))

MarkItDown().convert_uri(encoded_uri)
```

With an existing `C:\Temp\example.md`, the output is:

```text
(None, 'C:\\Temp\\example.md')
(None, 'C:\\C:\\Temp\\example.md')
OSError: [Errno 22] Invalid argument: 'C:\\C:\\Temp\\example.md'
```

## Expected behavior

The percent-encoded and unescaped drive-letter URIs should resolve to the same local path:

```text
C:\Temp\example.md
```

## Environment

- OS: Windows 11
- Python: 3.12.12
- MarkItDown: 0.1.6
- Repository commit: `e144e0a`

## Possible cause

`file_uri_to_path()` calls `url2pathname(parsed.path)` and then `os.path.abspath(...)`. On Windows, `%3A` is not decoded into the drive separator before `abspath` runs, so the current drive is prepended to a path that still contains the encoded drive component.

I searched existing issues for `C%3A`, percent-encoded Windows file URIs, and `file_uri_to_path` with no matching report found.

Bortlesboat:
## Summary
- recognize a percent-encoded Windows drive separator before url2pathname performs drive detection
- avoid abspath prepending the current drive to an already-decoded drive path
- cover parity between file:///C:/... and file:///C%3A/...

Closes #2208

## Verification
- PYTHONPATH=packages/markitdown/src python -m pytest packages/markitdown/tests/test_module_misc.py::test_file_uri_with_percent_encoded_drive_colon -q (1 passed on Windows)
- python -m ruff check packages/markitdown/src/markitdown/_uri_utils.py packages/markitdown/tests/test_module_misc.py
- python -m ruff format --check packages/markitdown/src/markitdown/_uri_utils.py packages/markitdown/tests/test_module_misc.py
- git diff --check

The existing test_file_uris case hard-codes POSIX paths and therefore does not run successfully on Windows; the new regression is isolated with a Windows marker.


afourney:
The encoded Windows drive-colon case has now been fixed by #2315. Thanks for diagnosing it. Please reopen if the original URI still produces an incorrect path on current main.

PR #2224, opened 2026-07-18
Preserve merged DOCX tables as HTML

### LINKED ISSUE #20: Word Document table conversion issue
This library is great. It would be even more useful if the table conversion is accurate with merged cells.

With the table inside this [docx file](https://1drv.ms/w/c/4ecd221091081033/EWwl0QRs_5JOro2habGbwlgBEiNGVNzTtXhWaiRNDPK3Kw?e=3nbFmw) 
Image

I got the parsing results as below:
```txt
| 1 | 2 | 3 | 4 | 5 | 6 | |
| --- | --- | --- | --- | --- | --- | --- |
| A | b | c | d | e | f | value |
| I | J | K | L | M |
| P | Q | R | S | T |
```

After rendering in markdown, it's like 
| 1 | 2 | 3 | 4 | 5 | 6 | |
| --- | --- | --- | --- | --- | --- | --- |
| A | b | c | d | e | f | value |
| I | J | K | L | M |
| P | Q | R | S | T |

StaryMoon:
## Summary

- preserve DOCX tables containing `rowspan` or `colspan` as semantic HTML
- keep simple DOCX tables in the existing pipe-table Markdown format
- scope the behavior to DOCX conversion so direct HTML conversion remains unchanged

## Why

Pipe-table Markdown cannot represent merged cells. Mammoth already emits the correct span attributes, but markdownify flattens the table and loses the relationship between cells. Keeping only complex DOCX tables as HTML preserves their structure for Markdown renderers and downstream LLM processing.

## Tests

- added regression coverage for row spans, column spans, simple tables, and the direct HTML default
- `7 passed` in the targeted DOCX/HTML test run
- verified end to end with a generated DOCX containing both horizontal and vertical merged cells

Fixes #20


StaryMoon:

afourney:
This is an interesting issue, and the fix looks reasonable. I'm not sure if we want to add more configuration options though (I worry there are already too many). Possibly we can detect merged cells and set the option automatically?   But that would yield inconsistent output formats from table to table....  I need to think a bit more about how to detect and handle this.

PR #2225, opened 2026-07-19
fix(html): reduce inline images to alt text under markdownify 1.x (convert_as_inline → parent_tags)

lntutor:
### What's broken

markdownify ≥1.0 changed the `convert_*` calling convention: it now invokes them as `convert_fn(node, text, parent_tags=<set>)` (where `parent_tags` contains the pseudo-tag `"_inline"` for headings/table cells) instead of the old positional `convert_as_inline` boolean. `_CustomMarkdownify.convert_img` still declares `convert_as_inline` and swallows `parent_tags` into `**kwargs`, so `convert_as_inline` is **always** `False` and the alt-text reduction never fires:

```python
HtmlConverter().convert_string('<h1>Hello <img src="a.png" alt="World"></h1>').markdown
# '# Hello ![World](a.png)'   -- expected '# Hello World'
```

An image in a table cell yields `| AA | bob |`, which with a long/`data:` URI can break the table row entirely. Base `markdownify.convert_img` handles this correctly via `"_inline" in parent_tags`; the override defeated it.

### Fix

Switch the override to accept `parent_tags` and check `"_inline" in parent_tags`, matching base markdownify. Block (paragraph) images and the `data:` URI truncation path are unaffected.

### Tests

Adds three cases to `test_module_misc.py`: inline image in a heading and in a table cell reduce to alt text; a block image is still rendered. All pass. (`convert_hn`/`convert_a`/`convert_input` carry the same obsolete signature but produce no visible wrong result, so this stays a minimal single-bug fix.)


lntutor:
@microsoft-github-policy-service agree [company="Vexere"]

lntutor:
@microsoft-github-policy-service agree company="Vexere"

PR #2227, opened 2026-07-21
fix: avoid O(n^2) value lookups in PPTX chart conversion

kz-000:
### Problem
Converting a PPTX that contains a large chart is extremely slow and can appear to hang. A real-world deck with a 1097-point, 3-series chart spent ~2.7s just reading the chart values, and larger charts effectively froze the conversion.

### Root cause
`PptxConverter._convert_chart_to_markdown` accessed `series.values[idx]` inside the nested (category × series) loop. In python-pptx, `series.values` rescans the cached data points via XPath on every access (O(n) per lookup), so the code was O(n²) per series and rebuilt the entire values tuple once per category — O(n³) overall.

### Fix
Materialize each series' values once before the loop, reducing the lookups to O(n). The generated Markdown is unchanged.

### Benchmark (3 series, varying category count; identical output verified)

| categories | before | after | speedup |
|---:|---:|---:|---:|
| 50  | 0.48s   | 0.01s | 47× |
| 100 | 3.12s   | 0.03s | 99× |
| 200 | 22.4s   | 0.11s | 205× |
| 400 | 164s    | 0.40s | 407× |

### Tests
Added `test_pptx_chart_multi_series_conversion`, which builds a multi-series chart with many categories and verifies the resulting Markdown table (series headers, first/last categories, and a representative data row).

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_pptx_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_pptx_converter.py
@@ -240,13 +240,19 @@ class PptxConverter(DocumentConverter):
             md += "\n\n"
             data = []
             category_names = [c.label for c in chart.plots[0].categories]
-            series_names = [s.name for s in chart.series]
+            series_list = list(chart.series)
+            series_names = [s.name for s in series_list]
             data.append(["Category"] + series_names)
+
+            # Materialize each series' values once. Accessing series.values[idx]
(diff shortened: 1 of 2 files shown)


kz-000:

afourney:
This looks quite sensible! Thanks for the fix.

PR #2228, opened 2026-07-21
Fix invalid LaTeX macros for mu, nu, tau, and down-arrow in equation conversion

ksprajapatii:
## What
Fixes typo'd LaTeX macros in `converter_utils/docx/math/latex_dict.py` 
that produced invalid LaTeX output when converting Word equations 
containing certain symbols.

## Bugs fixed
- `\m` → `\mu` (U+1D707, mathematical italic mu)
- `\n` → `\nu` (U+1D708, mathematical italic nu)
- `\ta` → `\tau` (U+1D70F, mathematical italic tau)
- `\downright` → `\downarrow` (U+2193, downwards arrow)

`\m`, `\n`, `\ta`, and `\downright` are not valid LaTeX commands. Any 
.docx file with an equation containing μ, ν, τ, or ↓ would produce 
broken/invalid LaTeX in the converted markdown output.

## How I found it
Ran `ruff` static analysis over the codebase, which flagged a 
duplicate dict key warning near these entries. Manual review of the 
surrounding Unicode-to-LaTeX mappings revealed the typos.

## Testing
- Verified the corrected dict values directly:
  `T['\U0001d707'] == '\\mu '`, etc.
- Existing `test_docx_equations` test in `test_module_misc.py` still 
  passes.
- Added a small regression test to catch these mappings from 
  regressing in future.

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/math/latex_dict.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/math/latex_dict.py
@@ -79,14 +79,11 @@ CHR_BO = {
 }
 
 T = {
-    "\u2192": "\\rightarrow ",
-    "\u2192": "\\rightarrow ",
-    "\u2192": "\\rightarrow ",
     # Greek letters
     "\U0001d6fc": "\\alpha ",
     "\U0001d6fd": "\\beta ",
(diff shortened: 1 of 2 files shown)


ksprajapatii:

afourney:
Merged! Thanks!

PR #2245, opened 2026-07-26
fix: resolve Outlook 365 sender address and convert HTML-only message bodies

### LINKED ISSUE #1188: Outlook converter reads incorrect fields from Outlook 365 (April 2025) files, misses HTML content - Fix included
A MSG file from Outlook 356 is not processed correctly. Fix attached (not production ready, please see if e.g. want to use Markdownify library?).

TO REPRODUCE
I'm passing a msg file saved from the browser version of Outlook 365. Unfortunately I can't share the file as I got it from another person, and I don't have direct access to O365 myself, but what failed was:
 - It took from field from a field which has organizational info, not email address (From field was something like "/O=EXCHANGELABS/OU=EXCHANGE ADMINISTRATIVE GROUP..."
 - It misses messages with HTML-only body; that is, ## Content is empty if the message is an html-only message

ELABORATION ON VERSIONS / TESTING
I'm not sure if the existing implementation works with some older/other MSG files? However, attached a version of the implementation which works with messages exported from Outlook O365 browser version as of today. Unfortunately I don't have access to Outlook myself, so I'm unable to verify if this new version works with other versions, or test this thoroughly.

THE FIX
Attached an Outlook converter source code file with the following changes:
 - take From field correctly
 - in case the text-only body is missing, try to find the html body
 -- it seems that at least with my python3.12, decode with UTF-16 succeeds without throwing even if the payload is actually iso-8859-1 (just output is then malformed), so we're attempting several different decodings, and see if the output looks like html
 -- finally, also run the html output through another library, markdownify, to turn HTML into markdown (I did not find html converter within markdown, so thus using a 3rd party library)

While creating the fix, I found this web page helpful in finding out the correct streams; I did check also Microsoft's own documentation that I found, but it did not match what I had in the MSG file. :)
https://www.devhut.net/retrieving-email-header-information-in-outlook-using-vba-part-2/ 

[_outlook_msg_converter_py.txt](https://github.com/user-attachments/files/19739392/_outlook_msg_converter_py.txt)

gaurav0107:
## Summary

`.msg` files saved from Outlook 365 / Exchange lose two things on conversion. Both are reproduced against `main` below.

**1. `From` is an Exchange DN, not an address.** The converter read the sender only from `PR_SENDER_EMAIL_ADDRESS` (`0x0C1F`). That property holds an email address only when `PR_SENDER_ADDRTYPE` (`0x0C1E`) is `SMTP`; anything routed through Exchange sets the address type to `EX` and stores a Distinguished Name there instead, so the header rendered as:

```
**From:** /O=EXCHANGELABS/OU=EXCHANGE ADMINISTRATIVE GROUP (FYDIBOHF23SPDLT)/CN=RECIPIENTS/CN=ab12cd34
```

The sender is now resolved in MAPI-correct order — `PR_SENDER_SMTP_ADDRESS` (`0x5D01`), then `PR_SENT_REPRESENTING_SMTP_ADDRESS` (`0x5D02`), then the legacy properties but only when their address type is not the DN form. If a file genuinely carries nothing else, the DN is still emitted, so no message loses its `From` header.

**2. HTML-only messages render an empty `## Content`.** The body was read only from `PR_BODY` (`0x1000`), which Outlook on the web leaves empty for HTML-only mail — the entire message body was silently dropped. It now falls back to `PR_HTML` (`0x1013`) and converts it with the `HtmlConverter` this repo already ships, the same way `PptxConverter` / `XlsxConverter` / `EpubConverter` delegate. Both the binary (`0102`) and string (`001F`) variants of the property are handled, and the byte stream's charset is detected with `charset_normalizer` — already a core dependency, used the same way in `_csv_converter.py` and `_plain_text_converter.py` — rather than assuming UTF-8.

Binary streams are read through a new `_read_stream` helper rather than `_get_stream_data`, because that function decodes UTF-16-LE first and arbitrary even-length bytes almost never fail that decode: `PR_HTML` would come back as mojibake instead of raising. `_get_stream_data` keeps its exact previous signature and decoding behavior.

No new dependency, and plain-text messages take the identical path they took before — the HTML fallback is reached only when `PR_BODY` is absent or empty.

Before / after on the message shape described in the issue:

```
**From:** /O=EXCHANGELABS/OU=.../CN=RECIPIENTS/CN=ab12cd34   ->  **From:** sender@example.com
## Content                                                    ->  ## Content
(empty)                                                              Hello **world**
```

## Testing

New `packages/markitdown/tests/test_outlook_msg.py`, 11 cases: SMTP-over-DN preference, the `sent representing` variants, the plain-`SMTP` case (unchanged behavior), the last-resort DN fallback, HTML-only body conversion, the string-variant property, non-UTF-8 (cp1252) HTML bytes, absent HTML, plain-text-wins-over-HTML, and an end-to-end `convert()` of the reported message shape.

A `.msg` file is an OLE2 compound file and nothing in the dependency set can write that container, so a synthetic binary fixture is not possible here the way `test_svg_no_fallback.pptx` was. The tests instead serve the MAPI streams through a stand-in for `olefile.OleFileIO`, matching the stub-the-collaborator style already used in `tests/test_docintel_html.py` and `tests/test_cu_converter.py`. Happy to swap in a real fixture if a maintainer can share a sanitised Exchange `.msg`.

```
pytest tests/test_outlook_msg.py            11 passed
pytest tests/ (full suite)                  347 passed, 3 skipped
black --check .                             73 files unchanged
mypy --ignore-missing-imports               no issues
```

The existing `test_outlook_msg.msg` vector is unaffected — that fixture has `PR_SENDER_ADDRTYPE == "SMTP"`, so `**From:** test.sender@example.com` still holds.

The only full-suite failure, `test_module_misc.py::test_speech_transcription`, fails identically on a clean checkout of `main`: the online recogniser returns `1 2 3 4` for a clip the test expects to end in `5`. Unrelated to this change.

Fixes #1188

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_outlook_msg_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_outlook_msg_converter.py
@@ -1,8 +1,13 @@
+import io
 import sys
 from typing import Any, Union, BinaryIO
+
+from charset_normalizer import from_bytes
+
 from .._stream_info import StreamInfo
 from .._base_converter import DocumentConverter, DocumentConverterResult
 from .._exceptions import MissingDependencyException, MISSING_DEPENDENCY_MESSAGE
+from ._html_converter import HtmlConverter
 
 # Try loading optional (but in this case, required) dependencies
 # Save reporting of any exceptions for later
@@ -28,6 +33,10 @@ class OutlookMsgConverter(DocumentConverter):
     - Email headers (From, To, Subject)
     - Email body content
     """
+
+    def __init__(self):
+        super().__init__()
+        self._html_converter = HtmlConverter()
 
     def accepts(
         self,
@@ -100,7 +109,7 @@ class OutlookMsgConverter(DocumentConverter):
 
         # Get headers
         headers = {
-            "From": self._get_stream_data(msg, "__substg1.0_0C1F001F"),
+            "From": self._get_sender(msg),
             "To": self._get_stream_data(msg, "__substg1.0_0E04001F"),
             "Subject": self._get_stream_data(msg, "__substg1.0_0037001F"),
         }
@@ -112,8 +121,11 @@ class OutlookMsgConverter(DocumentConverter):
 
         md_content += "\n## Content\n\n"
 
-        # Get email body
+        # Get email body. HTML-only messages leave PR_BODY empty and carry
+        # their content in PR_HTML instead.
         body = self._get_stream_data(msg, "__substg1.0_1000001F")
+        if not body:
+            body = self._get_html_body(msg)
         if body:
             md_content += body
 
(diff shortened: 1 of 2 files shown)


ericsondea-collab (commented):
> ## Summary
> 
> `.msg` files saved from Outlook 365 / Exchange lose two things on conversion. Both are reproduced against `main` below.
> 
> **1. `From` is an Exchange DN, not an address.** The converter read the sender only from `PR_SENDER_EMAIL_ADDRESS` (`0x0C1F`). That property holds an email address only when `PR_SENDER_ADDRTYPE` (`0x0C1E`) is `SMTP`; anything routed through Exchange sets the address type to `EX` and stores a Distinguished Name there instead, so the header rendered as:
> 
> ```
> **From:** /O=EXCHANGELABS/OU=EXCHANGE ADMINISTRATIVE GROUP (FYDIBOHF23SPDLT)/CN=RECIPIENTS/CN=ab12cd34
> ```
> 
> The sender is now resolved in MAPI-correct order — `PR_SENDER_SMTP_ADDRESS` (`0x5D01`), then `PR_SENT_REPRESENTING_SMTP_ADDRESS` (`0x5D02`), then the legacy properties but only when their address type is not the DN form. If a file genuinely carries nothing else, the DN is still emitted, so no message loses its `From` header.
> 
> **2. HTML-only messages render an empty `## Content`.** The body was read only from `PR_BODY` (`0x1000`), which Outlook on the web leaves empty for HTML-only mail — the entire message body was silently dropped. It now falls back to `PR_HTML` (`0x1013`) and converts it with the `HtmlConverter` this repo already ships, the same way `PptxConverter` / `XlsxConverter` / `EpubConverter` delegate. Both the binary (`0102`) and string (`001F`) variants of the property are handled, and the byte stream's charset is detected with `charset_normalizer` — already a core dependency, used the same way in `_csv_converter.py` and `_plain_text_converter.py` — rather than assuming UTF-8.
> 
> Binary streams are read through a new `_read_stream` helper rather than `_get_stream_data`, because that function decodes UTF-16-LE first and arbitrary even-length bytes almost never fail that decode: `PR_HTML` would come back as mojibake instead of raising. `_get_stream_data` keeps its exact previous signature and decoding behavior.
> 
> No new dependency, and plain-text messages take the identical path they took before — the HTML fallback is reached only when `PR_BODY` is absent or empty.
> 
> Before / after on the message shape described in the issue:
> 
> ```
> **From:** /O=EXCHANGELABS/OU=.../CN=RECIPIENTS/CN=ab12cd34   ->  **From:** sender@example.com
> ## Content                                                    ->  ## Content
> (empty)                                                              Hello **world**
> ```
> 
> ## Testing
> 
> New `packages/markitdown/tests/test_outlook_msg.py`, 11 cases: SMTP-over-DN preference, the `sent representing` variants, the plain-`SMTP` case (unchanged behavior), the last-resort DN fallback, HTML-only body conversion, the string-variant property, non-UTF-8 (cp1252) HTML bytes, absent HTML, plain-text-wins-over-HTML, and an end-to-end `convert()` of the reported message shape.
> 
> A `.msg` file is an OLE2 compound file and nothing in the dependency set can write that container, so a synthetic binary fixture is not possible here the way `test_svg_no_fallback.pptx` was. The tests instead serve the MAPI streams through a stand-in for `olefile.OleFileIO`, matching the stub-the-collaborator style already used in `tests/test_docintel_html.py` and `tests/test_cu_converter.py`. Happy to swap in a real fixture if a maintainer can share a sanitised Exchange `.msg`.
> 
> ```
> pytest tests/test_outlook_msg.py            11 passed
> pytest tests/ (full suite)                  347 passed, 3 skipped
> black --check .                             73 files unchanged
> mypy --ignore-missing-imports               no issues
> ```
> 
> The existing `test_outlook_msg.msg` vector is unaffected — that fixture has `PR_SENDER_ADDRTYPE == "SMTP"`, so `**From:** test.sender@example.com` still holds.
> 
> The only full-suite failure, `test_module_misc.py::test_speech_transcription`, fails identically on a clean checkout of `main`: the online recogniser returns `1 2 3 4` for a clip the test expects to end in `5`. Unrelated to this change.
> 
> Fixes #1188

PR #2248, opened 2026-07-27
fix(docx): guard None returned by findtext in OMML do_r method

amir-rezaei:
## Description
This PR resolves issue #2188 by guarding `elm.findtext(./{0}t.format(OMML_NS)) or ""` in `do_r()` (`packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py`).

## Details
- Prevents `TypeError: 'NoneType' object is not iterable` when an `<m:r>` math run has no `<m:t>` text child, preventing the entire document equation conversion pass from failing silently.


afourney:
Addressed by #2189

Issue #2166, opened 2026-06-26
- DocxConverter threw KeyError with message: 'w:type'


lbbniu:
```
markitdown 产品手册.docx
Traceback (most recent call last):
  File "/opt/anaconda3/bin/markitdown", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/anaconda3/lib/python3.12/site-packages/markitdown/__main__.py", line 254, in main
    result = markitdown.convert(
             ^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/lib/python3.12/site-packages/markitdown/_markitdown.py", line 306, in convert
    return self.convert_local(source, stream_info=stream_info, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/lib/python3.12/site-packages/markitdown/_markitdown.py", line 360, in convert_local
    return self._convert(file_stream=fh, stream_info_guesses=guesses, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/lib/python3.12/site-packages/markitdown/_markitdown.py", line 649, in _convert
    raise FileConversionException(attempts=failed_attempts)
markitdown._exceptions.FileConversionException: File conversion failed after 1 attempts:
 - DocxConverter threw KeyError with message: 'w:type'
 ```

PR #2250, opened 2026-07-27
fix(docx): handle non-OMML XML namespaces in process_children_list

amir-rezaei:
## Description
This PR resolves issue #2166 by stripping arbitrary XML namespaces in `process_children_list()` (`packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py`).

## Details
- Prevents `KeyError: 'w:type'` when elements carry WordprocessingML or custom XML namespaces during OMML conversion.

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/math/omml.py
@@ -88,9 +88,10 @@ class Tag2Method(object):
         process children of the elm,return iterable
         """
         for _e in list(elm):
-            if OMML_NS not in _e.tag:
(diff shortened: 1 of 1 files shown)


afourney:
Thanks for looking into this. The KeyError('w:type') reported in #2166 did not originate in the OMML parser; it came from Mammoth indexing a malformed style entry in word/styles.xml. That issue has now been fixed by #2190.

*This* change would also dispatch elements from arbitrary namespaces according to their local names, so an unrelated {custom}f or w:r could be interpreted as an OMML fraction or run.

Please try the most recent version. If the issue persists, we can re-visit this.

PR #2251, opened 2026-07-27
fix(html): preserve non-UTF-8 percent-encoded URL paths in HTML link conversion

amir-rezaei:
## Description
This PR resolves issue #2171 by catching `UnicodeDecodeError` during `unquote(parsed_url.path, errors="strict")` in `_markdownify.py`.

## Details
- Preserves original percent-encoded octets (e.g. EUC-JP / Shift-JIS URLs) instead of introducing UTF-8 replacement characters (`%EF%BF%BD`).


afourney:
Percent-encoded octets in HTML link paths are now preserved by #2173, with broader regression coverage. I’m closing this as superseded; please reopen if the original URL is still corrupted.

PR #2252, opened 2026-07-27
fix(html): promote first table row to header cells when th/thead elements are missing

amir-rezaei:
## Description
This PR addresses issue #2157 by promoting the first row's `<td>` cells to `<th>` in `convert_table()` (`_markdownify.py`) when a table lacks `<thead>` and `<th>` tags (such as DOCX tables converted via Mammoth).

## Details
- Prevents markdownify from generating an empty top header row (`| | |`) and divider for tables without explicit `<th>` tags.


e212156822000:
@amir-rezaei Help tag the author again.

Issue #23, opened 2024-12-14
Support for .doc extensions


mictab:
Are you planning to offer .doc support in addition to the current .docx support?

aviral-bhardwaj:
created PR https://github.com/microsoft/markitdown/pull/36

wfnian:
Patiently waiting for good news.

Shane32:
Just wanted to comment, I ran a check of file counts across our server, where I was considering using markitdown:
1. PDF - 1,150k files
2. JPG - 243k files
3. DOC - 112k files
4. XLS - 70k files
5. XLSX - 69k files
6. DOCX - 46k files

I think this demonstrates just how important DOC file handling is, even in 2025.

On a side note, what's not shown in the above list is the number of PDFs that are generated content versus scanned documents.  Most are probably scanned, and 95% of the scanned PDFs should be searchable images.

Other related PRs / discussions not already linked above:
- #281 
- #335 
- #1220

ashmod:
raised PR #1316

Chandu378:
Hi, I'm interested in working on this issue if it's still open. I'm new to open source and would appreciate any guidance.

tifilipebr:
I saw that some people have already made an effort and developed this functionality, which shows its importance to the community. I'd like to better understand what the assumptions are after the PR and why there's been a delay in providing support for .doc files.

ashmod:
I've recently tested this, #1316 , works well on both linux and windows with no need for third-party installations (provided MS Office is installed on the Windows client).

pazars:
The open-source LibreOffice has a CLI converter that also works for legacy .doc files. A bit bulky but doesn't require MS Office license and can be setup on linux machines.

dridk:
Hi,

With the help of Claude and the open format specification of binary Word doc 98, I created my own package to parse the *.doc format. 
It is written in Rust with a python binding. 

It works surprisingly well! With features like text box extraction and the preservation of titles translated into markdown.

You can try it using :  

cargo install unword. 
unword --input test.doc > test.md 

@yuchengpersonal did a greate pull request using antiword and textract . But from experience, the package I made works much better than antiword. 

Note: I'm a developer, I understood the generated code. But the doc format specification is so complex that Claude was a huge help

yousrae2004:
Hi is this issue still open for contribution? I'm new to open source and would like to help contribute!

dridk:
[...]

PR #2256, opened 2026-07-29
Add support for .doc (Microsoft Word 97-2003) files

awais786:
Implements issue #23 by adding DocConverter to handle legacy .doc format files. Uses unword library for parsing .doc files, which provides faster and more reliable conversion compared to older tools like antiword.

Changes:
- Add unword>=0.2.2 as optional dependency in pyproject.toml
- Create DocConverter class in _doc_converter.py to handle .doc files
- Register DocConverter in main MarkItDown class
- Add comprehensive tests for DocConverter (4/4 passing)
- Support both .doc extension and application/msword MIME type

The implementation follows the same pattern as DocxConverter, using unword to parse the binary .doc format and extract text content.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -166,3 +166,9 @@ cython_debug/
 src/.DS_Store
 .DS_Store
 .cursorrules
+# Added by code-review-graph
+.code-review-graph/
+# Claude Code configuration (local only)
+.claude/
+.mcp.json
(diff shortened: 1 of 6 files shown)


awais786:
@microsoft-github-policy-service agree company="Arbisoft"

Issue #1317, opened 2025-07-08
How to keep page number


harksin:
when converting pdf to markdown

kalle07:
use my pdfparser  ;)

https://github.com/kalle07/parsing

sanyam-dev:
I would like to contribute to this issue.

kalle07:
and what?
i mean his main question was not that specific ?
page number? from a pdf?
directly read out the number if a number is writen on the page is very complicate and you can only match it to your own PDF where you know where exactly is the numbers location ... because coordiates and postion differ always ... oc mostly on top or buttom left/midle or right but take a scientific paper and in this region u have more stuff and you cant finde out if its the page number...
so my parser only count each page and add "page X".

andreshere00:
Or use the mine, which integrates MarkItDown directly with the component `MarkItDownReader`. Note that this feature is only enabled when an LLM is used, if not, use the `VanillaReader` or `DoclingReader` components: https://github.com/andreshere00/Splitter_MR/

And it has a full ready-to-use documentation: https://andreshere00.github.io/Splitter_MR/

Issue #122, opened 2024-12-18
Pagewise Markdown output


bahtman:
Hi guys,

Love the work. In our current approach we convert documents to a list of markdown, where each element consists of the markdown for that specific page. This is really useful in RAG when providing citations and enriching the chunks during completion.

I'm unsure how feasible it is for PDF, docx and other "paginated" content.

jonasb:
The `.pptx` extractor includes `<!-- Slide number: X -->` which I use to split the markdown by page. In Excel there are `## Sheet name` titles for each sheet.

I agree that it would be nice with paged output on the other formats (pdf, docx for me) and also have an API to get the output paged directly so I don't need to split it myself based on markers.

tsitsimis:
This would be extremely useful. Currently using Azure Document Intelligence that can provide markdown of PDF files by page but we want to adopt markitdown to support more file formats (pdf, pptx, docx, xlsx)

emcf:
For those still looking for page-wise markdown extraction, [the library markitdown is based on](https://github.com/emcf/thepipe) has this feature

jedrazb:
Hey, for Word docs you can use `docx-to-markdown` package. Open-source, apache 2.0 and supports typescript ([npm](https://www.npmjs.com/package/@docx-editor.dev/docx-to-markdown)) and python ([pypi](https://pypi.org/project/docx-to-markdown/)) packages. 

You can try the demo: https://www.docx-editor.dev/solutions/word-to-markdown , or go to the [example in google colab notebook](https://colab.research.google.com/drive/1Gq30GRsgOP_2nakDHIE001wQMVnbOTvO?usp=sharing). Note that accurate page calculation depends on access to the fonts used in the document. Otherwise, it falls back to Google Fonts.

PR #2272, opened 2026-08-03
Add opt-in source anchors for PDF pages and spreadsheet rows

mlazarovits73:
Add opt-in source anchors for PDF pages and spreadsheet rows

Refs #122, #1317

Problem

Extracted Markdown loses the coordinates of the text it came from.

PdfConverter joins every page into a single string with no page marker — pages are concatenated with "\n\n", and in the prose path the whole document goes through pdfminer.high_level.extract_text at once, so even the page boundary is gone. XlsxConverter and XlsConverter drop the row index via to_html(index=False).

Consumers that need to cite an extracted claim back to a location in the original file therefore cannot do so. PptxConverter already solves this for slides by emitting  (_pptx_converter.py:88); this extends the same idea to PDF and Excel.

What this changes

A new opt-in option, source_anchors (CLI: --source-anchors), that emits stable coordinates alongside the text.

PDF —  before each page's content:

Invoice 2026-001006
Issued: 2026-04-02

| Item    | Amount | Date       |
| ------- | ------ | ---------- |
| Service | 18000  | 2026-01-05 |

The wording matches the existing slide marker so one parser handles both. HTML comments are invisible in rendered Markdown, survive round-trips, and are greppable.

XLSX / XLS —  before each sheet, plus a leading xlsx-row column carrying 1-based spreadsheet row numbers (header is row 1, so the first data row is 2):

## Payroll
| xlsx-row | name      | gross  |
| -------- | --------- | ------ |
| 2        | A. Kovacs | 450000 |
A side effect worth calling out

The existing PDF path picks its extraction strategy per document: if no page looks form-style, the whole file goes through pdfminer; if even one page does, every other page falls back to pdfplumber's page.extract_text(), which the code's own comment notes has worse spacing for prose.

The anchored path picks per page — form-style pages keep the pdfplumber table extraction, prose pages use pdfminer text. This costs one extra pdfminer pass, not one per page: pdfminer already terminates every page with a form feed, so a single whole-document extraction is split back per page.

Backwards compatibility

The option defaults to off and the existing code paths are untouched. Output without it is byte-identical to before — verified by diffing converted output against the unpatched tree for both PDF and XLSX fixtures, and covered by a test that asserts convert(path) equals convert(path, source_anchors=False).

The full existing suite reports the same results before and after this change (31 failed / 91 passed / 2 skipped in my environment; the failures are pre-existing and caused by optional dependencies I did not install, plus network-dependent tests).

Tests

packages/markitdown/tests/test_source_anchors.py, 12 tests covering:

unchanged default output, with and without the option explicitly set to False
anchors present, ordered, unique, starting at 1, never exceeding the page count
page-content correspondence: for each anchored section, the text must match its own page better than any other page in the document. This catches off-by-one and drift, and is robust against the tokenization differences between pdfminer and pdfplumber that a naive similarity threshold would trip over
sheet anchors present for every sheet, row numbers emitted
anchors absent by default, for both PDF and XLSX
the option reaching converters through convert_stream as well as convert
Validation on real documents

Beyond the repo fixtures, this was run over a set of real-world documents (government decisions, tax notices, an address card, an invoice, and a payroll spreadsheet — 8 files, mixed 1–2 page PDFs and one XLSX). Every file produced an anchor count equal to its actual page count, no text loss against the default output, and 99–100% own-page correspondence.

Open question for maintainers

#122 asks for paged output by default, consistent with how PPTX and XLSX already behave. I kept the default off so this change stays additive and nobody's existing output shifts. Flipping the default is a one-line change if you would rather have that — happy to do it in this PR.

Not covered

I did not test encrypted PDFs, rotated pages, or CJK documents. Scanned PDFs with no text layer produce no anchors, which is the correct behaviour (there is no text to anchor), but they are also not improved by this change.

### DIFF
--- a/packages/markitdown/src/markitdown/__main__.py
+++ b/packages/markitdown/src/markitdown/__main__.py
@@ -16,8 +16,7 @@ def main():
         description="Convert various file formats to markdown.",
         prog="markitdown",
         formatter_class=argparse.RawDescriptionHelpFormatter,
-        usage=dedent(
-            """
+        usage=dedent("""
             SYNTAX:
 
                 markitdown <OPTIONAL: FILENAME>
@@ -42,8 +41,7 @@ def main():
                 OR
 
                 markitdown example.pdf > example.md
-            """
-        ).strip(),
+            """).strip(),
     )
 
     parser.add_argument(
@@ -136,6 +134,12 @@ def main():
         "--keep-data-uris",
         action="store_true",
         help="Keep data URIs (like base64-encoded images) in the output. By default, data URIs are truncated.",
+    )
+
+    parser.add_argument(
+        "--source-anchors",
+        action="store_true",
+        help="Emit stable source coordinates in the output, so extracted text can be cited back to its location in the original file: '<!-- Page number: 3 -->' before each PDF page (matching the existing slide-number marker), and '<!-- Sheet name: Sheet1 -->' plus an xlsx-row column for spreadsheets. Off by default.",
     )
 
     parser.add_argument("filename", nargs="?")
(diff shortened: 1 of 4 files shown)


mlazarovits73:

PR #2276, opened 2026-08-05
Fix OCR fallback and image placement

MichaRotstein:
## AI assistance

This pull request was created with GitHub Copilot assistance and reviewed by the contributor before submission.

## Problem

OCR-enabled conversion did not reliably replace document images with text.

### Scanned PDFs

A scanned PDF could produce only page headings, such as:

`markdown
## Page 2
`

instead of a transcription. The converter decided whether to run full-page OCR by checking whether the **entire document** Markdown was empty. Because it adds ## Page N headings before extraction, that Markdown was never empty, even when a page contained only a scanned image. As a result, the page was never sent to the configured vision model.

### DOCX images

DOCX images were OCR'd before HTML generation, but the resulting text was later attached to  tags by list position. If one image returned no OCR text or failed, the remaining results moved up by one position. Text from a later image could appear where an earlier image belonged, while the final image could lose its transcription.

## Solution

- Track native text and successful inline-image OCR **per PDF page**.
- Render and OCR only pages with neither native text nor a successful embedded-image transcription.
- Keep native-text pages out of full-page OCR, preventing duplicate OCR of existing document text.
- Use Mammoth's convert_image callback to assign a unique placeholder to every DOCX image occurrence in document order.
- Replace only successful placeholders after HTML-to-Markdown conversion; remove empty or failed occurrences without shifting later image output.
- Preserve individual image and page OCR failure isolation.

## Result

- Scanned PDF pages receive full-page OCR instead of returning headings alone.
- Embedded PDF images are OCR'd inline without reprocessing native text.
- DOCX image transcription appears at the matching image location, including when a previous image OCR attempt fails.

## Tests

- python -m pytest packages\markitdown-ocr\tests\test_pdf_converter.py packages\markitdown-ocr\tests\test_docx_converter.py -q
- 23 focused tests passed.

### DIFF
--- a/packages/markitdown-ocr/src/markitdown_ocr/_docx_converter_with_ocr.py
+++ b/packages/markitdown-ocr/src/markitdown_ocr/_docx_converter_with_ocr.py
@@ -3,8 +3,6 @@ Enhanced DOCX Converter with OCR support for embedded images.
 Extracts images from Word documents and performs OCR while maintaining context.
 """
 
-import io
-import re
 import sys
 from typing import Any, BinaryIO, Optional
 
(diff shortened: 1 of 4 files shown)


MichaRotstein:

PR #2279, opened 2026-08-06
fix: remove extra closing brace from caron and ring-above accent templates

AndrewAvery7:
## Problem

A DOCX containing a caron- or ring-accented equation loses **every equation in the document**, silently, with a zero exit code.

## Root cause

In `converter_utils/docx/math/latex_dict.py`, two entries in `CHR` carry a fourth closing brace:

```python
"\u030a": "\ocirc{{{0}}}}",   # COMBINING RING ABOVE
"\u030c": "\check{{{0}}}}",   # COMBINING CARON
```

Every other accent in the table is `"\name{{{0}}}"`. `oMath2Latex.do_acc` applies these unguarded:

```python
return latex_s.format(c_dict["e"])   # omml.py:213
```

so both raise `ValueError: Single '}' encountered in format string`. Of the 46 `{0}` templates across `CHR`, `CHR_BO`, `CHR_DEFAULT`, `POS` and `POS_DEFAULT`, these two are the only ones affected.

The damage is not local to the accented equation. `pre_process_docx` runs `_pre_process_math` over the whole of `word/document.xml` inside a blanket `except Exception` and, on error, writes the **original unprocessed XML** back:

```python
try:
    updated_content = _pre_process_math(content)
    zip_output.writestr(name, updated_content)
except Exception:
    # If there is an error in processing the content, write the original content
    zip_output.writestr(name, content)
```

Mammoth does not render OMML, so every equation in that file disappears from the Markdown. Nothing is logged and the exit code is 0.

Both are standard entries in Word's **Equation → Accent** gallery, not obscure codepoints.

## Reproduction

Using this repository's own `tests/test_files/equations.docx`, repackaged, then repackaged again with a single caron `<m:acc>` paragraph appended:

| Input | Equations recovered |
|---|---|
| fixture, unmodified | 2 |
| fixture + **one** caron accent | **0** |
| same, with this fix | 3 (the original 2, plus `$$\check{x}$$`) |

## Fix

Remove the extra brace from both templates.

## Tests

Adds `tests/test_docx_math_accents.py`:

- `test_accent_templates_are_formattable` — asserts every `{0}` template in the accent tables survives `.format()`. This guards the whole table against the same class of typo rather than just these two entries.
- `test_caron_and_ring_accents` — covers U+030C and U+030A directly.
- `test_unmodified_accent_still_converts` — control on U+0302, which was never broken.

The first two fail on `main` and pass with the fix; the control passes either way.

Full suite locally: **322 passed before, 325 after**, with the same 14 pre-existing failures in `test_cli_vectors.py` / `test_module_misc.py` (Windows console encoding, unrelated to this change) in both runs.

## Notes

`\ocirc` is not standard LaTeX — `\mathring` is — but neighbouring entries follow the same upstream dwml/STIX naming (`\ovhook`, `\candra`, `\droang`), so I have kept this PR to the brace fix to stay obviously correct rather than mixing in a naming change.

Related: #2188 describes this same symptom (DOCX math converter silently dropping all equations) from a different root cause in `do_r`; the PRs against it do not touch `latex_dict.CHR`.


AndrewAvery7:

Issue #1514, opened 2025-12-27
[Security] Potential Denial of Service (DoS) via Zip Bomb in ZipConverter


sugarbinj:
A vulnerability was identified in the ZipConverter where it attempts to read the entire content of each file within a ZIP archive into memory simultaneously. This can be exploited using a "Zip Bomb" (a small ZIP file that decompresses to a very large size), leading to excessive memory consumption and potential system crashes (Out-of-Memory).‘
In packages/markitdown/src/markitdown/converters/_zip_converter.py, the code uses zipObj.read(name) to extract file contents:`z_file_stream = io.BytesIO(zipObj.read(name))`
Suggested Mitigation:
Check the file_size in ZipInfo before reading, and implement a configurable maximum size limit (e.g., 100MB). Alternatively, use a streaming approach if the underlying sub-converters support it.

abneeeees:
I would like to work on this issue.

mvanhorn:
I've submitted a PR to address this: #1628

The fix adds three pre-read checks in `ZipConverter.convert()`:

1. Per-file decompressed size limit (100 MB default)
2. Decompression ratio limit (100:1 default) to catch classic zip bombs
3. Cumulative total decompressed size limit (500 MB default)

Files exceeding any limit are skipped with a warning log rather than crashing the process. All limits are module-level constants for easy adjustment.

Issue #1167, opened 2025-04-05
while converting a URI/URL, requests needs timeout when server doesn't close the stream


ThePatelCode:
```
        elif uri.startswith("http:") or uri.startswith("https:"):
            response = self._requests_session.get(uri, stream=True)
            response.raise_for_status()
            return self.convert_response(
                response,
                stream_info=stream_info,
                file_extension=file_extension,
                url=mock_url,
                **kwargs,
            )
        else:
            raise ValueError(
                f"Unsupported URI scheme: {uri.split(':')[0]}. Supported schemes are: file:, data:, http:, https:"
            )

```

I have an encouter issues when server doesn't close the steam and it hangs forever trying to buffer the stream

I propose adding a timeout value. I have tested this locally, and if you give me direction, i.e how do we pass the timeout value to this function, I can make the change myself.

I propose modifying `stream_info` blob to add timeout.

afourney:
Yes, this is a good point and an easy fix. Let me pencil this in for the next update.

ThePatelCode:
@afourney , if you give me sense of direction which you are think, I am happy to contribute as well

sonnylaskar:
I was about to report the same issue. This seems to be an easy fix. Looking forward to the next build. Thanks to the team.

sammydeprez:
It is a very easy fix, just fixed in a PR, hope it can be merged soon.

PR #2296, opened 2026-08-12
Add resource-exhaustion guards for untrusted archives and URLs

SashaMIT:
## Problem

When converting untrusted input, two code paths allow unbounded resource consumption:

1. `ZipConverter` extracts every archive member into memory with no limits on member count, total uncompressed size, or compression ratio. A small crafted archive (a zip bomb, including nested archives processed recursively) can expand into gigabytes of memory and CPU work.
2. `convert_uri()` fetches remote content with no explicit timeout, and `convert_response()` buffers the entire response body with no size cap. A slow or endless endpoint can hang a conversion indefinitely or stream unbounded data into memory.

## Changes

**ZipConverter guards** (`_zip_converter.py`), each overridable per conversion via keyword arguments:

- `zip_max_members` (default 10000): maximum number of archive members
- `zip_max_total_uncompressed_size` (default 500 MB): enforced against declared sizes up front, and against actual bytes read during chunked extraction, since header sizes in crafted archives cannot be trusted
- `zip_max_compression_ratio` (default 100): per-member ratio check, applied only to members larger than 1 MB so small legitimately compressible files are not affected

Archives exceeding a limit fail with a `FileConversionException` describing the violated limit.

**URL fetch guards** (`_markitdown.py`):

- Explicit timeout passed to `requests` (default 30 seconds, overridable via the `timeout` keyword argument); `requests.Timeout` surfaces on slow endpoints
- `max_response_size` (default 100 MB, overridable via keyword argument): fast-fail on oversized `Content-Length` headers, and streaming reads abort past the limit with a `FileConversionException`

Defaults are conservative but chosen so normal documents are unaffected; the existing `test_files.zip` vector and the existing module tests pass unchanged.

## Tests

New `tests/test_resource_guards.py` (11 tests, no live network; HTTP is mocked):

- Zip: member count, compression ratio, and total size limits each reject crafted archives; small compressible members and normal archives still convert
- URL: default and custom timeout plumbed through to `session.get`; timeout errors surface; oversized `Content-Length` rejected before streaming; streaming past the limit aborts; a response exactly at the limit still converts

The guard tests were verified to fail without the source changes (revert-test), and `test_module_vectors.py` plus `test_module_misc.py` pass.

Refs #1514 and #1167.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -59,6 +59,12 @@ PRIORITY_GENERIC_FILE_FORMAT = (
     10.0  # Near catch-all converters for mimetypes like text/*, etc.
 )
 
+# Defaults for fetching HTTP/HTTPS URIs, guarding against hanging or
+# unbounded responses from untrusted endpoints. Both can be overridden per
+# call via the "timeout" and "max_response_size" keyword arguments.
+DEFAULT_REQUESTS_TIMEOUT = 30  # seconds
+DEFAULT_MAX_RESPONSE_SIZE = 100 * 1024 * 1024  # 100 MB
+
 
 _plugins: Union[None, List[Any]] = None  # If None, plugins have not been loaded yet.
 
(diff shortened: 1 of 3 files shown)


SashaMIT:

PR #2297, opened 2026-08-12
fix(rss): preserve Atom XHTML content

weivwang:
## What

Preserve structured Atom `summary` and `content` elements whose `type` is `xhtml`. Their child XHTML is now serialized and passed through the existing HTML-to-Markdown conversion path, while plain text and escaped HTML entries keep their current behavior.

## Why

Atom allows XHTML content as child elements rather than character data. The existing helper only returned the first text/CDATA child, so valid `type="xhtml"` summaries and content were silently dropped from converted feeds. This loses the most important part of many entries for downstream LLM and indexing workflows.

## Validation

- New regression tests reproduce the loss on `main` and cover XHTML `summary`, XHTML `content`, and unchanged plain-text content behavior (2 passed)
- Existing RSS/Atom vector coverage: 7 passed
- Black check: passed
- mypy for the changed source and tests: passed
- `git diff --check`: passed

I also ran the wider vector file in the minimal environment: 63 cases passed; 48 optional-format cases could not run because DOCX/XLSX/PDF/PPTX and related extras were not installed.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_rss_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_rss_converter.py
@@ -28,6 +28,8 @@ CANDIDATE_FILE_EXTENSIONS = [
     ".xml",
 ]
 
+XHTML_NAMESPACE = "http://www.w3.org/1999/xhtml"
+
 
 class RssConverter(DocumentConverter):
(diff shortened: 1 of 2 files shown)


weivwang:

PR #2318, opened 2026-08-19
fix(cli): allow Content Understanding conversion from stdin

uczltw6:
## Summary

- preserve the CLI's filename-optional stdin mode when `--use-cu` is enabled
- let piped input use the existing `convert_stream()` dispatch path
- add a regression test for Content Understanding stdin routing

## Problem

The CLI documents piping input without a filename, but the `--use-cu` branch exits before reaching the shared stdin path. Content Understanding analyzes binary input and does not require a local file path, so this guard unnecessarily prevents `cat report.pdf | markitdown --use-cu --cu-endpoint ...`.

The endpoint validation remains unchanged, and filename-based conversion is unaffected.

## Tests

- `hatch test tests/test_cu_converter.py tests/test_cli_misc.py -q` — 127 passed
- `hatch test -q` — 337 passed, 4 skipped
- Black 23.7.0 on both changed files — passed

OpenAI Codex assisted with identifying the inconsistent CLI branch and preparing the focused regression test; the behavior and test results were verified locally.

### DIFF
--- a/packages/markitdown/src/markitdown/__main__.py
+++ b/packages/markitdown/src/markitdown/__main__.py
@@ -216,8 +216,6 @@ def main():
             _exit_with_error(
                 "Content Understanding Endpoint (--cu-endpoint) is required when using --use-cu."
             )
-        elif args.filename is None:
-            _exit_with_error("Filename is required when using Content Understanding.")
 
         cu_kwargs: Dict[str, Any] = {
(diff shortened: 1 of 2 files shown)


sapentia-ola (approved):

sapentia-ola (approved):

afourney:
Merged. Thanks.

PR #2322, opened 2026-08-20
feat: add option to include DOCX comments

### LINKED ISSUE #2275: Feature request: Possibility to include comments of docx into md file
Use-case: As a user, who wants to convert *.docx file into *.md, I would like to have possibilty to include not only document text, but also comments of reviewer.
Input: *.docx (with reviewer comments)
Output: *.md (with review comment text and reference, where this comment belongs in the text)

garlic9912:
## What this changes

This adds an `include_comments` option for DOCX files:

```python
result = md.convert("review.docx", include_comments=True)
```

There is also a CLI flag:

```bash
markitdown review.docx --include-comments
```

Mammoth already knows how to convert Word comments, but it requires the following style-map rule:

```text
comment-reference =>
```

So this PR mainly makes that existing behavior easier to discover and use.

Comments are still excluded by default because changing the default would change the output for existing users. Comments can also contain internal review notes and add quite a lot of extra text.

If the user provides a custom `style_map`, it is kept and the comment rule is appended to it. User rules come first, so a custom `comment-reference` rule still takes priority.

I kept Mammoth's existing comment links and backlinks instead of converting them to Markdown footnotes. Footnotes would need extra decisions around Markdown dialects, existing footnotes, repeated references, tables, and backlinks. That felt outside the scope of this change.

The tests cover:

- comments disabled by default;
- enabling comments per conversion or on `MarkItDown`;
- overriding the constructor option for one conversion;
- multiple comments and their links/backlinks;
- comments inside normal text and tables;
- combining comments with a custom `style_map`;
- the new CLI flag.

Closes #2275.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,13 @@ You can also pipe content:
 cat path-to-file.pdf | markitdown
 ```
 
+Reviewer comments are excluded from DOCX output by default. Include them, along
+with references to their locations in the document, with `--include-comments`:
+
+```bash
+markitdown reviewed-document.docx --include-comments
+```
+
 ### Optional Dependencies
 MarkItDown has optional dependencies for activating various file formats. Earlier in this document, we installed all optional dependencies with the `[all]` option. However, you can also install them individually for more control. For example:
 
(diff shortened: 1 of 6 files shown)


garlic9912:

PR #2325, opened 2026-08-21
Add PyQt6 multi-file GUI for MarkItDown

rahulpwar654:
`markitdown-gui` desktop app implementation

### DIFF
--- a/MARKITDOWN_GUI_SETUP.md
+++ b/MARKITDOWN_GUI_SETUP.md
@@ -1,0 +1,115 @@
+# MarkItDown GUI Setup and Run Guide
+
+This guide explains how to install dependencies and run the `markitdown-gui` desktop app.
+
+## 1. Prerequisites
+
+- Python 3.10+
+- PowerShell (Windows)
+- This repository cloned locally
+
+## 2. Open the Project Folder
+
+```powershell
+cd D:\develpment\FileConverterToMD
+```
+
+## 3. Create and Activate Virtual Environment
+
+```powershell
+python -m venv .venv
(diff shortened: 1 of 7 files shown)


rahulpwar654:

R-zin (commented):
for source_file in self._selected_files:
    result = converter.convert(str(source_file))
If you use this then when a large PDF or any file is queued for conversion it uses the same thread for the Qt GUI as well (The GUI can freeze) so I think you should spawn a Worker Thread which will handle MarkItDown.convert() function which will be on separate thread and when it finishes it can update the Qt thread. You can even add a progress bar in GUI also.

R-zin (commented):
for source_file in self._selected_files:
    result = converter.convert(str(source_file))
If you use this then when a large PDF or any file is queued for conversion it uses the same thread for the Qt GUI as well (The GUI can freeze) so I think you should spawn a Worker Thread which will handle MarkItDown.convert() function which will be on separate thread and when it finishes it can update the Qt thread. You can even add a progress bar in GUI also.

rahulpwar654:
Changes in _qt_gui.py:

MarkItDown.convert() now runs inside a QObject worker moved to a QThread.
Added a determinate progress bar showing completed files.
Worker emits signals for:
Progress updates
Log messages
Final success/failure counts
All Qt widget updates remain on the main GUI thread.
File selection and conversion controls are disabled while work is running.
Worker and thread objects use deleteLater() cleanup.
Output collision handling was moved into the worker.

R-zin (approved):

afourney:

PR #2333, opened 2026-08-23
fix: fall back to plain text when RSS item content triggers RecursionError

shoemoney:
`RssConverter._parse_content` wraps markdownify conversion in a broad `except BaseException` that returns the original content. When an RSS item's `description` or `content:encoded` contains deeply nested HTML (roughly 400+ levels), markdownify's recursive DOM traversal raises `RecursionError`, and the converter silently embeds the raw unconverted HTML in the markdown output.

This PR mirrors the fix merged in #1644 for the HTML converter: catch `RecursionError` specifically, emit a warning, and fall back to BeautifulSoup's iterative `get_text()` so callers still get usable plain-text content. Other exceptions keep the existing behavior.

Includes a deterministic regression test using the same `sys.setrecursionlimit` pattern as the #1644 test: a feed item with 500 nested divs must produce text output with no raw HTML tags. Before the fix the output contained 500 raw `<div` tags; after, none. `black --check` clean.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_rss_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_rss_converter.py
@@ -1,3 +1,5 @@
+import warnings
+
 from defusedxml import minidom
 from xml.dom.minidom import Document, Element
 from typing import BinaryIO, Any, Union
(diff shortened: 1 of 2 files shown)


afourney:
Thanks for the contribution. Merged.

PR #2342, opened 2026-08-26
Preserve strikethrough from <strike> and CSS line-through

### LINKED ISSUE #2340: Strikethrough is silently lost for `<strike>`, CSS `line-through`, and DOCX double-strike
## Description

MarkItDown preserves some strikethrough representations but silently flattens semantically equivalent variants to plain text. This makes superseded or corrected text indistinguishable from current text in the Markdown output.

The behavior reproduces with the latest PyPI release, `markitdown==0.1.7`.

## Minimal HTML reproduction

```python
from pathlib import Path
from tempfile import TemporaryDirectory

html = """<!doctype html>
<html><body>
<p>Plain <s>s element</s> after.</p>
<p>Plain <del>del element</del> after.</p>
<p>Plain <strike>strike element</strike> after.</p>
<p>Plain <span style="text-decoration: line-through;">inline CSS line-through</span> after.</p>
<p>Plain <span style="text-decoration-line: line-through;">CSS text-decoration-line</span> after.</p>
</body></html>
"""

with TemporaryDirectory() as tmp:
    path = Path(tmp) / "fixture.html"
    path.write_text(html, encoding="utf-8")
    print(MarkItDown().convert(str(path)).markdown)
```

## Actual output

```markdown
Plain ~~s element~~ after.

Plain ~~del element~~ after.

Plain strike element after.

Plain inline CSS line-through after.

Plain CSS text-decoration-line after.
```

## Expected output

All five inputs express strikethrough and should retain that meaning, for example:

```markdown
Plain ~~s element~~ after.

Plain ~~del element~~ after.

Plain ~~strike element~~ after.

Plain ~~inline CSS line-through~~ after.

Plain ~~CSS text-decoration-line~~ after.
```

At minimum, the obsolete but still encountered `<strike>` element could be normalized consistently with `<s>` and `<del>`. If parsing inline CSS is intentionally out of scope, documenting that limitation would make the loss less surprising.

## DOCX observation

A minimal DOCX shows the same inconsistency between OOXML variants:

| OOXML run property | Actual Markdown |
| --- | --- |
| `<w:strike/>` | `~~single strike~~` |
| `<w:dstrike/>` | `double strike` |

Mapping double-strike to ordinary Markdown strikethrough would preserve the important semantic distinction from current text even if Markdown cannot represent the double line exactly.

## Impact

The text itself remains present, so conversion appears successful, but consumers can interpret superseded text as current. This is particularly misleading in notes, procedures, and other documents where strikethrough carries meaning.

## Environment

- MarkItDown: 0.1.7
- Python: 3.12.13
- OS: macOS 26.6.2, arm64

The same HTML and DOCX outputs were also observed through a MarkItDown MCP installation using `markitdown==0.1.5`.

## Duplicate check

I searched open and closed issues for `strikethrough`, `strike`, `line-through`, `double strike`, and `dstrike` and did not find a report matching this behavior.

gyanu2507:
`<s>` and `<del>` already convert to `~~text~~`. `<strike>` and `style="text-decoration: line-through"` were flattened, so struck text looked like current text.

I hooked those two cases in the markdownify subclass. DOCX double-strike still depends on how mammoth emits HTML; I left that alone here.

Fixes #2340

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_markdownify.py
+++ b/packages/markitdown/src/markitdown/converters/_markdownify.py
@@ -122,33 +122,9 @@ class _CustomMarkdownify(markdownify.MarkdownConverter):
             return "[x] " if el.has_attr("checked") else "[ ] "
         return ""
 
-    def convert_strike(
-        self,
-        el: Any,
-        text: str,
-        convert_as_inline: Optional[bool] = False,
-        **kwargs,
-    ) -> str:
+    def convert_strike(self, el: Any, text: str, *args, **kwargs) -> str:
         """Obsolete <strike> is still in the wild; treat it like <s>/<del>."""
-prefix, suffix, text = markdownify.chomp(text)  # type: ignore
-        return f"{prefix}~~{text}~~{suffix}" if text else ""
-
-    def convert_span(
-        self,
-        el: Any,
-        text: str,
-        convert_as_inline: Optional[bool] = False,
-        **kwargs,
-    ) -> str:
-        """Preserve CSS line-through as Markdown strikethrough."""
-        if text and self._has_line_through(el):
-            return f"~~{text}~~"
-        return text
-
-    @staticmethod
-    def _has_line_through(el: Any) -> bool:
-        style = (el.get("style") or "").lower()
-        return "line-through" in style
+        return self.convert_s(el, text, *args, **kwargs)  # type: ignore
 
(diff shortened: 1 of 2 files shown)


gyanu2507:

afourney:
I am not yet ready to parse CSS or style tags. We don't do that for other cases like font weight or text decoration. I've updated the PR to just handle the <strike> tags.

Issue #83, opened 2024-12-17
Markdown doesn't preserve text relationship


NikhilVerma:
Here's a sample PDF - https://www.bnm.gov.my/documents/20124/963937/Risk+Management+in+Technology+(RMiT).pdf/810b088e-6f4f-aa35-b603-1208ace33619?t=1592866162078

However there are several parsing errors which I will try to highlight below

## Line elements aren't preserved

<img width="1194" alt="Image" src="https://github.com/user-attachments/assets/b851cfe3-0e08-4b40-adc5-f0557b8d22bb" />

Output

```
G  8.5

To  promote  effective  technology  discussions  at  the  board  level,  the
composition  of  the  board  and  the  designated  board-level  committee  should
include at least a member with technology experience and competencies.
```

## Indentations are ignored and text ordering is altered

<img width="1138" alt="Image" src="https://github.com/user-attachments/assets/b0436d57-2332-4451-b4b5-d3d4e5578579" />

```
The TRMF must include the following:
(a)  clear definition of technology risk;
(b)  clear responsibilities assigned for the management of technology risk at
different  levels  and  across  functions,  with  appropriate  governance  and
reporting arrangements;
the  identification  of  technology  risks  to  which  the  financial  institution  is
exposed,  including  risks  from  the  adoption  of  new  or  emerging
technology;

(c)

(d)  risk classification of all information assets/systems based on its criticality;
(e)  risk measurement and assessment approaches and methodologies;
(f)
(g)  continuous monitoring to timely detect and address any material risks.
```

## List elements are ignored and presented in entirely new lines

<img width="1088" alt="Image" src="https://github.com/user-attachments/assets/bc206c7f-6f7e-4e53-9446-2ec225220a14" />

```
1.

2.

3.

4.

5.

6.

The assurance shall be conducted by an independent external service provider
(ESP) engaged by the financial institution.

The independent ESP must understand the proposed services, the data flows,
system architecture, connectivity as well as its dependencies.

The  independent  ESP  shall  review  the  comprehensiveness  of  the  risk
assessment performed by the financial institution and validate the adequacy of
the control measures implemented or to be implemented.

The Risk Assessment Report (as per Part D in Appendix 7) shall state among
others, the scope of review, risk assessment methodology, summary of findings
and remedial actions (if any).
```

PDF parsing will always be a PITA, but I think these issues can be addressed by tracking the locations of the elements, right now I feel it simply loops over the textual elements and uses simple algorithms to merge them together

[...]

Issue #41, opened 2024-12-15
Preserve tables, titles (structure) of PDF documents


shcheklein:
I was trying to convert a SEC 10-K (PDF) as an examples. Running it like this:

```shell
python -m markitdown ~/Downloads/4.General\ Electric\ Company.pdf > ge.md
```

And I see that the resulting Markdown doesn't include tables, titles, etc. Pretty much, no structure. It makes it suboptimal for analyzing documents downstream (e.g. to pass a particular table to LLM and ask it to calculate something, or at least extract a specific value).

If it is out of scope for this tool - feel free to close the ticket. I wonder if there simple to use tools that can do that. I was trying unstructured lib, but it requires a quite complicated setup also to extract tables and it seems their open source is becoming less maintained (?).

Documents:

 [4.General Electric Company.pdf](https://github.com/user-attachments/files/18141690/4.General.Electric.Company.pdf)

ckpearson:
PDFs are an absolute nightmare specifically for this reason, and we've had issues with reading order from direct file extraction because the order things are rendered in is not necessarily the order they're present in the file either.

We've just gone down the route of using Azure's Document Intelligence system which does a much better job of extracting structural information (even tables!), but of course that's not an offline option and does lead to some cost / latency considerations.

dipam7:
Just tested this as well with a v simple table and found the output no better than raw pdf > txt. 

Btw why is the output txt and not markdown? I don't see titles or section headings with `###`. The output seems plain text.

jvhgit:
Hi, you might want to take a look at https://github.com/Filimoa/open-parse .

gabrielstuff:
This would be a great feature. Pymupdf is the only one I found doing it almost perfectly : https://pymupdf.readthedocs.io/en/latest/pymupdf4llm/index.html

Docling manage to do it but it is not that good.

PeterStaar-IBM:
> This would be a great feature. Pymupdf is the only one I found doing it almost perfectly : https://pymupdf.readthedocs.io/en/latest/pymupdf4llm/index.html
> 
> Docling manage to do it but it is not that good.

@gabrielstuff What do you mean? Running the following command,

```
docling --from pdf --to html --image-export-mode embedded /Users/taa/Downloads/4.General.Electric.Company.pdf
```

I get the following (few screenshots)

![Image](https://github.com/user-attachments/assets/bb9699f4-e083-4bbd-9fcb-009e02b98c75)

[...]

PR #2345, opened 2026-08-26
feat(ocr): add opt-in semantic PDF extraction

cagdasyurekli:
## Summary

- add an opt-in `semantic_pdf_ocr` mode to the existing `markitdown-ocr` plugin
- render every PDF page for vision-assisted semantic Markdown extraction
- preserve headings, reading order, nested lists, and tables without changing the default converter path
- fall back to deterministic whole-document extraction when semantic OCR fails or returns empty

Addresses the remaining semantic-structure cases described in #41 and #83.

## Design

This is deliberately plugin-scoped and disabled by default. PDF structure recovery is model-dependent, slower, and more expensive, so the deterministic core converter remains unchanged.

The semantic prompt asks the model to preserve source text and hierarchy and to treat document instructions as untrusted data. Constructor and per-conversion overrides are supported. Both pdfplumber and PyMuPDF rendering paths use the same prompt, and PyMuPDF resources are closed on every exit path.

Default OCR remains backward compatible with custom services that implement the original one-argument `extract_text(image_stream)` method.

## Verification

- `pytest -q packages/markitdown-ocr/tests/test_pdf_converter.py -k 'semantic or single_argument'` — 8 passed, 14 deselected
- `pytest -q packages/markitdown-ocr/tests/test_pdf_converter.py -k 'not test_pdf_multipage'` — 21 passed, 1 deselected
- Black 23.7.0 — passed
- `git diff --check` — passed

The deselected multipage assertion is an unchanged baseline/dependency-version mismatch: current pdfplumber parses that fixture instead of taking the historical PyMuPDF fallback. New tests separately exercise the malformed-PDF PyMuPDF semantic path.

## Limitations

Semantic output quality depends on the selected vision model. The tests verify routing, fallback, resource cleanup, option precedence, prompt constraints, and Markdown preservation with deterministic mocks; they do not claim perfect structure recovery for every arbitrary PDF.

### DIFF
--- a/packages/markitdown-ocr/README.md
+++ b/packages/markitdown-ocr/README.md
@@ -65,6 +65,25 @@ md = MarkItDown(
 )
 ```
 
+### Semantic PDF OCR (Opt-in)
+
+For deeply complex or malformed PDFs (e.g., issues #41 and #83), you can opt into a full-page semantic structure mode. When enabled, this mode routes every PDF page through the vision model, asking it for faithful semantic Markdown (headings, lists, reading order, and tables), bypassing standard deterministic text extraction.
+
+```python
+md = MarkItDown(
+    enable_plugins=True,
+    llm_client=OpenAI(),
+    llm_model="gpt-4o",
+    semantic_pdf_ocr=True, # Explicit opt-in
+)
+```
+
+**Important caveats for semantic mode:**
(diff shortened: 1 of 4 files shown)


cagdasyurekli:

PR #2347, opened 2026-08-27
feat: preserve Excel cell formatting via XlsxConfig (fix #53)

### LINKED ISSUE #53: Unable to extract currency from excel formatted cells

Vanity-1:
Fixes #53

## Summary
Adds an opt-in XlsxConfig object to preserve Excel cell number formatting (currency, percentage, thousands separators, dates) when converting .xlsx files to Markdown.

## Motivation
pd.read_excel only reads raw cell values, so formatted values like $1,199.00 are emitted as 1199. This PR makes formatting preservation available without breaking existing consumers that parse numeric values from the output.

## Changes
- Add XlsxConfig dataclass with a preserve_formatting: bool = False flag (default keeps current behavior).
- XlsxConverter now accepts an optional config argument; when preserve_formatting is enabled it reads cells via openpyxl and renders values according to their number format.
- Currency symbols are extracted from the format string (e.g. $, €, ¥, []) rather than hardcoded.
- MarkItDown(xlsx_config=XlsxConfig(preserve_formatting=True)) wires the config through nable_builtins.
- XlsxConfig is exported from the markitdown package.
- Adds 	ests/test_xlsx_converter.py covering currency, percentage, thousands separators, negative values, other currency symbols, dates, and backward compatibility.

## Usage
python
from markitdown import MarkItDown, XlsxConfig

md = MarkItDown(xlsx_config=XlsxConfig(preserve_formatting=True))
result = md.convert("invoice.xlsx")

## Backward compatibility
Default behavior is unchanged (raw cell values), so existing code that does loat(cell_value) keeps working.

### DIFF
--- a/packages/markitdown/src/markitdown/__init__.py
+++ b/packages/markitdown/src/markitdown/__init__.py
@@ -17,6 +17,7 @@ from ._exceptions import (
     FileConversionException,
     UnsupportedFormatException,
 )
+from .converters._xlsx_converter import XlsxConfig
 
 __all__ = [
     "__version__",
@@ -29,6 +30,7 @@ __all__ = [
     "FileConversionException",
     "UnsupportedFormatException",
     "StreamInfo",
+    "XlsxConfig",
     "PRIORITY_SPECIFIC_FILE_FORMAT",
     "PRIORITY_GENERIC_FILE_FORMAT",
 ]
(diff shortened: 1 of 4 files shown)


Vanity-1:

PR #2351, opened 2026-08-28
fix: buffer CLI stdin before format detection on Windows

### LINKED ISSUE #1209: Basic HTML conversion via stdin not working on Windows
To reproduce on Windows 11, run this in Powershell or Command Prompt (not WSL):
```
echo "<html><body><h1>Test HTML</h1></body></html>" | markitdown -x html
```

Result: 
```
b''
```

Expected (this is what is returned if run in WSL):
```
# Test HTML
```

This also produces the expected results in Powershell:
```
echo "<html><body><h1>Test HTML</h1></body></html>" >test.html
markitdown test.html
```

Adding `-c cp1252` does not fix the issue. It appears to have something to do with the way stdin is read in binary mode that causes decoding issues on Windows.

mohui666:
Fixes #1209

## What changed
- buffer CLI stdin into `BytesIO` before conversion so format probing can rewind it on Windows
- make the existing stdin vector test actually exercise stdin
- add a regression for a pipe that reports seekable but cannot rewind
- run CLI subprocess tests with the active interpreter and read UTF-8 output as UTF-8

## Root cause
On Windows, pipe-backed `sys.stdin.buffer` can report `seekable() == True` even though `seek(0)` does not rewind it. Magika consumes the pipe during format detection, leaving the converter at EOF. This is why the CLI returned an empty result even for ASCII HTML.

## Verification
- real PowerShell pipe: HTML piped to `python -m markitdown -x html` produces `# Test HTML`
- direct raw-byte subprocess stdin: return code 0, expected stdout, empty stderr
- targeted stdin regressions: 13 passed
- CLI tests: 41 passed, 12 skipped
- package tests: 307 passed, 33 skipped, with the unchanged Windows-only URI path assertion deselected
- Black, compileall, and `git diff --check` pass

The only native-Windows full-suite failure is the pre-existing `test_file_uris` assertion, which expects the Unix path `/path/to/file.txt` instead of Windows `C:\path\to\file.txt`.

### DIFF
--- a/packages/markitdown/src/markitdown/__main__.py
+++ b/packages/markitdown/src/markitdown/__main__.py
@@ -2,8 +2,9 @@
 #
 # SPDX-License-Identifier: MIT
 import argparse
-import sys
 import codecs
+import io
+import sys
 from typing import Any, Dict
 from textwrap import dedent
 from importlib.metadata import entry_points
@@ -245,8 +246,9 @@ def main():
         markitdown = MarkItDown(enable_plugins=args.use_plugins)
 
     if args.filename is None:
+        # Windows pipe-backed stdin can report seekable() even though it cannot rewind.
         result = markitdown.convert_stream(
-            sys.stdin.buffer,
+            io.BytesIO(sys.stdin.buffer.read()),
             stream_info=stream_info,
             keep_data_uris=args.keep_data_uris,
         )
--- a/packages/markitdown/src/markitdown/__main__.py
+++ b/packages/markitdown/src/markitdown/__main__.py
@@ -6,7 +6,6 @@ import os
 import sys
 import codecs
 import io
-import sys
 from typing import Any, Dict
 from textwrap import dedent
 from importlib.metadata import entry_points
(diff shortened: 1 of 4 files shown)


mohui666:
> Mo Hui (@mohui666)请阅读以下贡献者许可协议（CLA）。如果您同意该 CLA，请回复以下信息。
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > 选项：
> > 
> > * （默认 - 未指定公司）我拥有我提交作品的全部知识产权，并且我不是在为我的雇主工作期间提交作品。
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * （如适用）本人提交作品系为雇主工作所致（或雇主根据合同或适用法律对本人提交的作品拥有知识产权）。本人已获得雇主许可，代表雇主提交作品并签署本协议。下方签名中的“您”一词包含本人及雇主。
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> 贡献者许可协议

@microsoft-github-policy-service agree [company="{your company}"]

mohui666:

PR #2356, opened 2026-08-30
fix: preserve strikethrough semantics for <strike>, line-through CSS, and w:dstrike

### LINKED ISSUE #2340: Strikethrough is silently lost for `<strike>`, CSS `line-through`, and DOCX double-strike

wasim-builds:
Fixes #2340.

This preserves strikethrough semantics that were previously lost during extraction:
1. Added `convert_strike` to `_CustomMarkdownify` for obsolete `<strike>` tags.
2. Normalized inline CSS `text-decoration: line-through` to `<del>` in HTML preprocessing.
3. Normalized DOCX `<w:dstrike>` (double strikethrough) to `<w:strike>` in DOCX preprocessing so Mammoth handles it correctly.

### DIFF
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -108,6 +108,9 @@ def _pre_process_math(content: bytes) -> bytes:
         bytes: The processed content with OMML elements replaced by their LaTeX equivalents, encoded as bytes.
     """
     soup = BeautifulSoup(content.decode(), features="xml")
+    # Normalize dstrike to strike to preserve strikethrough semantics
+    for tag in soup.find_all("dstrike"):
+        tag.name = "strike"
     for tag in soup.find_all("oMathPara"):
         _replace_equations(tag)
     for tag in soup.find_all("oMath"):
--- a/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
+++ b/packages/markitdown/src/markitdown/converter_utils/docx/pre_process.py
@@ -174,30 +174,26 @@ def pre_process_docx(input_docx: BinaryIO) -> BinaryIO:
         BinaryIO: A binary output stream representing the processed DOCX file.
     """
     output_docx = BytesIO()
-    # The files that need to be pre-processed from .docx
+    # The pre-processing steps to apply to each file in the .docx
     pre_process_enable_files = {
-        "word/document.xml": _pre_process_math,
-        "word/footnotes.xml": _pre_process_math,
-        "word/endnotes.xml": _pre_process_math,
-        "word/styles.xml": _pre_process_styles,
+        "word/document.xml": (_pre_process_strike, _pre_process_math),
+        "word/footnotes.xml": (_pre_process_strike, _pre_process_math),
+        "word/endnotes.xml": (_pre_process_strike, _pre_process_math),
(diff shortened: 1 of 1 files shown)


Elioooon:
I reproduced the DOCX path while investigating #2340. One edge is worth locking down before merge: `w:dstrike w:val="0"` is explicitly disabled and must remain plain text.

A focused end-to-end test can clone the existing `test.docx` fixture in memory, inject one run with `<w:dstrike/>` and one with `<w:dstrike w:val="0"/>`, then assert:

```python
assert "~~double strike marker~~" in result.text_content
assert "disabled double strike marker" in result.text_content
assert "~~disabled double strike marker~~" not in result.text_content
```

The HTML portions also overlap #2342; the distinct value in this PR is the `w:dstrike` normalization, so coverage for that path would make the scope much safer.

PR #2360, opened 2026-08-31
Mitigate UnicodeDecodeError due to wrong ASCII charset guess for long files

### LINKED ISSUE #2359: Bug: UnicodeDecodeError in JSON conversion when non-ASCII character appears after 4 KB
When converting long JSON files, a `UnicodeDecodeError` is raised by `IpynbConverter.accepts()` if a non-ASCII character appears after the first 4 KB of the file. I believe this occurs because only the first 4 KB are inspected by MarkItDown to guess the charset.

Here is a minimal reproducer:
```python
import json
import tempfile
from pathlib import Path

# JSON file with the first non-ASCII char after 4096 bytes.
payload = {
    "record": {
        "title": "Example record",
        "abstract": "This is sample test. " * 500,
    },
    "notes": "non-ASCII character: è",
}
text = json.dumps(payload, ensure_ascii=False, indent=2)

md = MarkItDown()

with tempfile.TemporaryDirectory() as temp_dir:
    path = Path(temp_dir) / "input.json"
    path.write_text(text, encoding="utf-8")

    result = md.convert(str(path))
```

Resulting traceback:
```
  File "/home/fcatalan/mid_issue/reproducer_4k.py", line 22, in <module>
    result = md.convert(str(path))
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/fcatalan/Desktop/Repos/markitdown/packages/markitdown/src/markitdown/_markitdown.py", line 306, in convert
    return self.convert_local(source, stream_info=stream_info, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fcatalan/Desktop/Repos/markitdown/packages/markitdown/src/markitdown/_markitdown.py", line 360, in convert_local
    return self._convert(file_stream=fh, stream_info_guesses=guesses, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fcatalan/Desktop/Repos/markitdown/packages/markitdown/src/markitdown/_markitdown.py", line 617, in _convert
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fcatalan/Desktop/Repos/markitdown/packages/markitdown/src/markitdown/converters/_ipynb_converter.py", line 36, in accepts
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 10604: ordinal not in range(128)
```

My setup is 
- MarkItDown: from 'main' (commit `9dc0d6579b8739c9d0671ff205e071e3053c7df1`)
- Python: 3.12.3
- OS: Ubuntu 24.04

This is similar to the issue reported for PDF conversion in #1894.

fcatalan92:
This PR mitigate `UnicodeDecodeError` issues encountered when a non-ASCII character appears late in the file, namely after the window used to guess the charset.

The changes are the following:
- prevent `IpynbConverter.accepts()` from failing when a file cannot be decoded with the guessed charset. I implemented the same solution used in [outlook_msg_converter](https://github.com/microsoft/markitdown/blob/main/packages/markitdown/src/markitdown/converters/_outlook_msg_converter.py#L58-L67)
- increase the charset-detection sample from 4 KB to 64 KB, making the issue unlikely to occur 
- add guidance to conversion errors caused by an incorrectly guessed charset

Another possibility is to force a retry if a `UnicodeDecodeError` is encountered in the converter using the same approach used when the detected charset is `None`. But I preferred to not change the current behavior of the converters. https://github.com/microsoft/markitdown/blob/9dc0d6579b8739c9d0671ff205e071e3053c7df1/packages/markitdown/src/markitdown/converters/_plain_text_converter.py#L66-L69

The PR fixes #2359. It could also address #1505, #1894 and #1949 (for which there are no working reproducers).

@afourney I hope you could have a look at this human-generated contribution  ;-)

### DIFF
--- a/packages/markitdown/src/markitdown/_exceptions.py
+++ b/packages/markitdown/src/markitdown/_exceptions.py
@@ -76,7 +76,7 @@ class FileConversionException(MarkItDownException):
                         if isinstance(attempt.exc_info[1], UnicodeDecodeError):
                             message += (
                                 "The charset may have been automatically guessed incorrectly. "
-                                "If you didn't specify the charset, please retry with an explicitly defined one.\n" 
+                                "If you didn't specify the charset, please retry with an explicitly defined one.\n"
                             )
 
         super().__init__(message)
--- a/packages/markitdown/src/markitdown/_exceptions.py
+++ b/packages/markitdown/src/markitdown/_exceptions.py
@@ -73,4 +73,10 @@ class FileConversionException(MarkItDownException):
                     else:
                         message += f" - {type(attempt.converter).__name__} threw {attempt.exc_info[0].__name__} with message: {attempt.exc_info[1]}\n"
 
+                        if isinstance(attempt.exc_info[1], UnicodeDecodeError):
+                            message += (
+                                "The charset may have been automatically guessed incorrectly. "
+                                "If you didn't specify the charset, please retry with an explicitly defined one.\n" 
+                            )
+
         super().__init__(message)
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -357,7 +357,6 @@ class MarkItDown:
             guesses = self._get_stream_info_guesses(
                 file_stream=fh, base_guess=base_guess
             )
-            print(f"Guesses for {path}: {guesses}")
             return self._convert(file_stream=fh, stream_info_guesses=guesses, **kwargs)
 
     def convert_stream(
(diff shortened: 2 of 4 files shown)


fcatalan92:

afourney:
Thanks! Merged!

PR #2387, opened 2026-09-04
fix(markitdown-ocr): sort placeholders by length before substitution to prevent prefix collision

### LINKED ISSUE #2383: markitdown-ocr: placeholder prefix collision drops OCR blocks in DOCX with 11 or more images
# markitdown-ocr: placeholder prefix collision drops OCR blocks in DOCX with 11 or more images

markitdown-ocr 0.1.0, `packages/markitdown-ocr/src/markitdown_ocr/_docx_converter_with_ocr.py`, checked against `main` @ `b752951`.

The internal placeholder is `MARKITDOWNOCRBLOCK{}` formatted with the plain index, and the substitution walks the indices in ascending order with `str.replace`. `MARKITDOWNOCRBLOCK1` is a prefix of `MARKITDOWNOCRBLOCK10` through `MARKITDOWNOCRBLOCK19`, so once a document has eleven images or more, replacing index 1 corrupts every two-digit placeholder. Some OCR blocks get duplicated in the wrong place, others disappear. Nothing raises.

## The code

Line 30:

```python
_PLACEHOLDER = "MARKITDOWNOCRBLOCK{}"
```

`convert`, line 139:

```python
for i, raw_text in enumerate(ocr_texts):
    placeholder = _PLACEHOLDER.format(i)
    ocr_block = f"*[Image OCR]\n{raw_text}\n[End OCR]*"
    md = md.replace(placeholder, ocr_block)
```

At `i == 1` the search string is `MARKITDOWNOCRBLOCK1`, which also matches the first 20 characters of `MARKITDOWNOCRBLOCK10`, `...11` and so on. Those become `<block for index 1>0`, `<block for index 1>1`, and when the loop later reaches indices 10 and 11 there is nothing left to replace.

No document needed to see it:

```python
from markitdown_ocr._docx_converter_with_ocr import _PLACEHOLDER

md = " ".join(_PLACEHOLDER.format(i) for i in range(13))
md = md.replace(_PLACEHOLDER.format(0), "<text-0>")
md = md.replace(_PLACEHOLDER.format(1), "<text-1>")
print(md)
# <text-0> <text-1> MARKITDOWNOCRBLOCK2 ... <text-1>0 <text-1>1 <text-1>2
#                                            placeholders 10, 11, 12 are gone
```

## Reproduction

[repro_markitdown_ocr.py](https://github.com/user-attachments/files/31825064/repro_markitdown_ocr.py) builds its own documents and calls no external service. The OCR service is a stub returning a marker derived from the image bytes.

```
pip install markitdown markitdown-ocr python-docx pillow
python repro_markitdown_ocr.py
```

Case A is a document with 12 distinct images in natural order, so relationship order equals body order and the matching logic is correct by construction. Only this bug can show:

```
=== A. 12 images, body order == relationship order ===
images in the document body : 12
OCR blocks produced         : 12
blocks in the right place   : 10/12
```

Looking at which markers came out:

```
expected : 1a14 7b8c 868d b349 c9bf b88e 118b 7c27 39b6 b7fa ffb5 4f5c
produced : 1a14 7b8c 868d b349 c9bf b88e 118b 7c27 39b6 b7fa 7b8c 7b8c

repeated : {'7b8c': 3}          the text for index 1, emitted three times
lost     : ['ffb5', '4f5c']     the texts for indices 10 and 11
```

Case C in the same script is a three-image control that passes.

## Why the current tests don't catch it

Every fixture in `packages/markitdown-ocr/tests/ocr_test_data/` has at most three images, and the threshold here is exactly eleven. Ten images or fewer are unaffected.

## Possible fix

Give the token a terminator so no index token is a prefix of another:

```python
_PLACEHOLDER = "MARKITDOWNOCRBLOCK{}END"
```

`MARKITDOWNOCRBLOCK1END` is not a prefix of `MARKITDOWNOCRBLOCK11END`, and the token stays alphanumeric so it still survives the HTML to markdown step unescaped. Both call sites already read the constant, so that's the whole change.

If changing the token isn't wanted, iterating the replacement in descending index order works too, or `re.sub` with a word boundary.

I can send a PR.

Found while looking into a different bug in the same converter, where OCR text is matched to images by position. Filed separately since the cause is unrelated, though a document with eleven or more images usually hits both.

bunnysayzz:
## Root cause

`_PLACEHOLDER = "MARKITDOWNOCRBLOCK{}"` in `packages/markitdown-ocr/src/markitdown_ocr/_docx_converter_with_ocr.py`.

The substitution loop in `DocxConverterWithOCR._convert_local()` iterates in ascending index order and calls `str.replace(placeholder, ocr_block)` for each image. Because `"MARKITDOWNOCRBLOCK1"` is a prefix of `"MARKITDOWNOCRBLOCK10"` through `"MARKITDOWNOCRBLOCK19"`, replacing index 1 first corrupts every two-digit placeholder: `MARKITDOWNOCRBLOCK10` becomes the OCR content of block 1 followed by a literal `"0"`, dropping the OCR content for image 10 and duplicating the content for image 1.

Reproduction (no OCR service needed):

```python
_PLACEHOLDER = "MARKITDOWNOCRBLOCK{}"
md = " ".join(f"<p>{_PLACEHOLDER.format(i)}</p>" for i in range(11))
for i in range(11):
    md = md.replace(_PLACEHOLDER.format(i), f"[OCR:{i}]")
# Result: ... <p>[OCR:1]</p> ... <p>[OCR:1]0</p>
# OCR block 10 is gone; block 1 appears twice.
```

## Fix

Build the full list of `(placeholder, ocr_block)` pairs first, then sort by **descending placeholder length** before iterating. Longer placeholders (`MARKITDOWNOCRBLOCK10`) are replaced before shorter ones (`MARKITDOWNOCRBLOCK1`), so no prefix collision can occur.

```python
replacements = [
    (_PLACEHOLDER.format(i), f"*[Image OCR]\n{raw_text}\n[End OCR]*")
    for i, raw_text in enumerate(ocr_texts)
]
replacements.sort(key=lambda pair: len(pair[0]), reverse=True)
for placeholder, ocr_block in replacements:
    md = md.replace(placeholder, ocr_block)
```

## Test evidence

Five regression tests in `packages/markitdown-ocr/tests/test_ocr_placeholder_collision.py`:

```
PASSED tests/test_ocr_placeholder_collision.py::test_buggy_code_leaves_corrupted_placeholder_for_11_images
PASSED tests/test_ocr_placeholder_collision.py::test_fixed_code_no_remaining_placeholders_for_11_images
PASSED tests/test_ocr_placeholder_collision.py::test_fixed_code_all_ocr_blocks_present_for_11_images
PASSED tests/test_ocr_placeholder_collision.py::test_fixed_code_no_remaining_placeholders_for_100_images
PASSED tests/test_ocr_placeholder_collision.py::test_no_regression_under_10_images
5 passed in 0.02s
```

The first test (`test_buggy_code_leaves_corrupted_placeholder_for_11_images`) documents the bug by asserting the old behavior. The remaining four verify the fix.

Fixes #2383

### DIFF
--- a/packages/markitdown-ocr/src/markitdown_ocr/_docx_converter_with_ocr.py
+++ b/packages/markitdown-ocr/src/markitdown_ocr/_docx_converter_with_ocr.py
@@ -144,9 +144,15 @@ class DocxConverterWithOCR(HtmlConverter):
 
             # 5. Swap placeholders for the actual OCR blocks (post-conversion
             #    so * and _ are never escaped by the markdown converter).
-            for i, raw_text in enumerate(ocr_texts):
-                placeholder = _PLACEHOLDER.format(i)
-                ocr_block = f"*[Image OCR]\n{raw_text}\n[End OCR]*"
+            # Sort by descending placeholder length so that e.g. "MARKITDOWNOCRBLOCK10"
+            # is replaced before "MARKITDOWNOCRBLOCK1", preventing prefix-collision corruption
+            # when a document has 10 or more images.
+            replacements = [
+                (_PLACEHOLDER.format(i), f"*[Image OCR]\n{raw_text}\n[End OCR]*")
+                for i, raw_text in enumerate(ocr_texts)
+            ]
+            replacements.sort(key=lambda pair: len(pair[0]), reverse=True)
+            for placeholder, ocr_block in replacements:
                 md = md.replace(placeholder, ocr_block)
 
             return DocumentConverterResult(markdown=md)
--- a/packages/markitdown-ocr/tests/test_ocr_placeholder_collision.py
+++ b/packages/markitdown-ocr/tests/test_ocr_placeholder_collision.py
@@ -1,0 +1,93 @@
+"""
+Regression test for https://github.com/microsoft/markitdown/issues/2383
+
+markitdown-ocr: placeholder prefix collision drops OCR blocks in DOCX with 11 or more images.
+
+_PLACEHOLDER = "MARKITDOWNOCRBLOCK{}" -- when substituted in ascending index order,
+"MARKITDOWNOCRBLOCK1" is a prefix of "MARKITDOWNOCRBLOCK10" through "MARKITDOWNOCRBLOCK19",
+so str.replace corrupts every two-digit placeholder.
+"""
+
+import re
+
+_PLACEHOLDER = "MARKITDOWNOCRBLOCK{}"
+
+
+def _simulate_substitution_buggy(n_images: int) -> str:
+    """Reproduce the old (buggy) ascending-order substitution."""
+    md = " ".join(f"<p>{_PLACEHOLDER.format(i)}</p>" for i in range(n_images))
+    ocr_texts = [f"OCR text for image {i}" for i in range(n_images)]
+    for i, raw_text in enumerate(ocr_texts):
+        placeholder = _PLACEHOLDER.format(i)
+        ocr_block = f"*[Image OCR]\n{raw_text}\n[End OCR]*"
+        md = md.replace(placeholder, ocr_block)
+    return md
+
+
+def _simulate_substitution_fixed(n_images: int) -> str:
+    """The fixed substitution: sort by descending placeholder length first."""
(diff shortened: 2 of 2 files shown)


bunnysayzz:

PR #2391, opened 2026-09-06
fix(csv): warn when a data row has more columns than the header

### LINKED ISSUE #2390: CSV converter silently truncates data columns when rows have more fields than the header row
Environment: markitdown 0.1.7 (also reproducible on main? confirmed on 0.1.7)

Describe the bug:
The CsvConverter treats rows[0] (the first CSV line) as the header without
any validation. When a data row contains more fields than the header row,
extra fields are silently dropped:
  row = row[: len(rows[0])]     # _csv_converter.py, "Truncate..." branch
No warning is emitted, so users lose columns without knowing.

Repro (no external files needed):
  1. Save this as test.csv (note the first line has ONE field, acting as a
     preamble/comment; the real header is on line 2 — this is exactly what
     WizTree exports):

     生成由 WizTree 4.28
     文件名称,大小,分配
     C:\a,100,100
     C:\b,200,200

  2. Run: markitdown test.csv
     Expected: a table with headers from line 2 (文件名称/大小/分配).
     Actual: a ONE-column table (only 文件名称); 大小/分配 are silently
     dropped, no error, no warning.

Expected behavior / suggestion:
  - Detect column-count mismatch (header vs any data row). If mismatch:
      * emit at least a warning, and/or
      * rebuild the header to the max width (e.g. column-N) instead of
        silently truncating.
  - Optionally: a heuristic for a preamble line (a lone single-field row
    followed by a wider row) is nice-to-have, but the non-silent behavior
    is the core ask.

Context:
  WizTree (disk-usage scanner) exports CSVs whose first line is a
  copyright/donation banner (1 field) and whose real header is line 2.
  This is a real-world file shape, not an exotic hand-edited input.
  (Related recent work on this converter: PR #2303 — BOM/blank-row fix.)

lazerg:
CsvConverter treats `rows[0]` as the header with no validation, so a data row with extra columns just gets truncated with `row[: len(rows[0])]`, dropping fields silently. WizTree exports hit this in practice: their CSV starts with a one-field banner line before the real header, so every real column after the first gets dropped without any indication something went wrong.

This adds a warning when that truncation happens, so the data loss is visible instead of silent. The truncation behavior itself is unchanged.

Fixes #2390

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -1,6 +1,7 @@
 import csv
 import io
 import re
+import warnings
 from typing import BinaryIO, Any
 from charset_normalizer import from_bytes
 from .._base_converter import DocumentConverter, DocumentConverterResult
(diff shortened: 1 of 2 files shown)


afourney:
I agree that we should expand the columns to match the widest row. However, I don't want to overfit to WizTree.  I will address the more general case, and recommend that people implement a custom plugin for WizTree specifically.

lazerg:
Agreed. #2430 is the better fix. Once no row gets truncated there is nothing left to warn about, so my change is redundant. I am closing this one. Thanks for taking it.

PR #2392, opened 2026-09-07
fix(ocr): warn when LLM extraction fails

### LINKED ISSUE #2313: Better exception handling/warning in LLMVisionOCRService.extract_text
LLMVisionOCRService.extract_text() catches all exceptions from client.chat.completions.create() and stores them in OCRResult.error. However, PdfConverterWithOCR never inspects OCRResult.error. It only checks whether ocr_result.text is non-empty.

Therefore configuration errors, incompatible clients, unavailable models, API failures, etc. appear to the user as if OCR simply produced no text, with no warning or exception.

This makes it particularly difficult to debug errors and bugs in using markitdown-ocr...

DEOWL-kan:
## Summary

- emit a Python warning when LLM Vision OCR catches a client failure, so converter callers no longer hide it
- include only the exception type in the warning while preserving the original message in `OCRResult.error`
- add a regression test for warning visibility, error preservation, sensitive-message exclusion, and stream rewind

Fixes #2313.

## Testing

- `python -m pytest -q packages/markitdown-ocr/tests` — 42 passed
- `git diff --check`

Not run locally: `pre-commit` (not installed in the test environment).

### DIFF
--- a/packages/markitdown-ocr/src/markitdown_ocr/_ocr_service.py
+++ b/packages/markitdown-ocr/src/markitdown_ocr/_ocr_service.py
@@ -4,8 +4,9 @@ Provides LLM Vision-based image text extraction.
 """
 
 import base64
-from typing import Any, BinaryIO
+import warnings
 from dataclasses import dataclass
+from typing import Any, BinaryIO
 
 from markitdown import StreamInfo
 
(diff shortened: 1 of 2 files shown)


DEOWL-kan:

PR #2396, opened 2026-09-07
fix(html): preserve whitespace around anchor text

Izharalijaffar:
HTML anchor conversion trims text with `chomp()`, but several return paths discard the removed whitespace. For example, `<p>before<a> middle </a>after</p>` currently becomes `beforemiddleafter`, losing word boundaries.

Preserve the original text when an anchor has no destination, contains only whitespace, or appears inside `<pre>`. Include the extracted prefix and suffix when emitting an autolink. Normal links and rejected JavaScript links retain their existing behavior.

Adds regression coverage through `MarkItDown.convert_stream()` for missing/empty hrefs, named anchors, whitespace-only anchors, preformatted text, and autolinks, plus controls for normal and rejected links.

Validation on Windows, Python 3.12.14, markdownify 1.2.3:
- Before the fix: 10 new regression cases failed; 8 cases passed.
- After the fix: all 18 HTML tests pass.
- Full core suite: 424 passed, 33 skipped, 3 failed. All three failures also reproduce with the original converter: `test_file_uris` and `test_convert_case_insensitive_uri_schemes` expect POSIX paths; `test_doc_rlink` attempts to write to `C:\tmp\test_rlink.txt`. Remote tests were skipped using the suite's `GITHUB_ACTIONS` setting; no live LLM calls were made.
- `pre-commit run --all-files`: passed.
- Wheel build: passed.
- Focused mypy check reports the same five existing `options` attribute errors on both original and changed converter; no new errors.

Prepared and tested with Codex assistance.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_markdownify.py
+++ b/packages/markitdown/src/markitdown/converters/_markdownify.py
@@ -61,12 +61,13 @@ class _CustomMarkdownify(markdownify.MarkdownConverter):
         **kwargs,
     ):
         """Same as usual converter, but removes JavaScript links and escapes URIs."""
+        original_text = text
         prefix, suffix, text = markdownify.chomp(text)  # type: ignore
         if not text:
-            return ""
+            return original_text
 
         if el.find_parent("pre") is not None:
-            return text
+            return original_text
 
         href = el.get("href")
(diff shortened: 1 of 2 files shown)


Izharalijaffar:

PR #2233, opened 2026-07-22
fix: handle PPTX SVG images without a rasterized fallback

guoyu-wang:
PowerPoint stores an SVG picture as an <a:blip> whose r:embed points to a raster PNG fallback plus an <asvg:svgBlip> extension for the SVG. When a picture has no raster fallback, the <a:blip> has no r:embed, so python-pptx's shape.image raises ValueError("no embedded image"). This caused PptxConverter to fail the entire conversion.

Resolve the SVG part directly from the svgBlip extension via a new _get_image_info helper, and guard _is_picture against the exception (hasattr only swallows AttributeError). Add unit tests plus a small synthetic PPTX fixture covering the SVG-without-fallback case.

PR #2404, opened 2026-09-07
fix(ocr): handle PPTX SVG images without a rasterized fallback

### LINKED ISSUE #2399: OCR PPTX converter crashes on SVG images without a raster fallback
## Summary

`PptxConverterWithOCR` still crashes on SVG pictures that do not have a raster fallback, even though the core `PptxConverter` was fixed for this in #2233.

Current OCR path:

```python
if self._is_picture(shape):
    image_stream = io.BytesIO(shape.image.blob)
```

For an SVG-only picture, python-pptx raises `ValueError("no embedded image")` when `shape.image` is accessed. The core converter now resolves the `<asvg:svgBlip>` relationship directly, but the OCR converter does not use that helper/path.

## Reproduction on current main

I used the repository's existing regression fixture:

`packages/markitdown/tests/test_files/test_svg_no_fallback.pptx`

In a clean local venv with the source packages installed:

```text
core 57
ocr error ValueError 'no embedded image'
```

So the same file that proves #2233 fixed the core converter still fails through `PptxConverterWithOCR`.

## Proposed scope

If this is available to take, I can send a focused OCR-only follow-up that reuses the core converter's SVG image-resolution behavior where practical and adds an OCR-side regression test using the existing fixture. No unrelated PPTX refactor.

I searched open and closed issues/PRs for an OCR-specific SVG/no-raster-fallback fix and found no matching implementation.

AI-assisted audit disclosure: found by comparing recent merged core-converter fixes against sibling OCR converters; reproduction was executed locally against current source.

lazerg:
`PptxConverterWithOCR` reads `shape.image.blob` directly, so a deck whose picture is an SVG with no rasterized fallback fails the whole conversion with `ValueError: no embedded image`. The core `PptxConverter` stopped doing that in #2233, where `_get_image_info` resolves the `<asvg:svgBlip>` relationship when `shape.image` raises.

The OCR converter now goes through the core converter for both the image blob and the picture check, so the SVG reaches the OCR service and everything else converts as before. Test uses the same fixture as the core regression test.

Fixes #2399.


rajpratham1 (approved):
The PPTX SVG OCR fix looks correct. The image handling now uses the core PptxConverter logic for SVGs without a rasterized fallback, and the added regression test covers this case. The dependency version is also updated consistently. I don't see any blocking issues in this PR.

PR #2405, opened 2026-09-07
Fix PPTX OCR converter emitting literal backslash-n instead of newlines

As9xm:
`_pptx_converter_with_ocr.py` used escaped `\\n` literals in slide headers, text frames, notes, tables, and charts, so output contained literal `\n` text instead of line breaks.
This change replaces all 10 occurrences with real `\n` newlines to restore correct markdown formatting.


As9xm:

PR #2406, opened 2026-09-07
Fix charset detection returning literal None string on empty input

As9xm:
`PlainTextConverter` and `CsvConverter` did `str(from_bytes(...).best())` without a None check, so when detection failed it returned the literal string `"None"` as document content.
This change checks for `None` and falls back to `""` to avoid corrupting output.


afourney:
I believe this was addressed in #2384. Please re-open, if I am mistaken.

PR #2412, opened 2026-09-08
Fix CSV parsing for CR-only line endings

### LINKED ISSUE #2411: CSV with CR-only line endings falls back to plain text instead of a Markdown table
### Summary

CSV input using CR-only (\r) record separators does not produce a Markdown table. With the explicit CSV/UTF-8 stream information below, `CsvConverter` raises `csv.Error`, while `MarkItDown.convert_stream()` falls back to plain text and returns the original CSV content. The same input with LF or CRLF separators produces the expected table.

### Environment

- Windows, Python 3.12.14
- MarkItDown 0.1.8b1, reproduced against current source at [b6e8bbd](https://github.com/microsoft/markitdown/commit/b6e8bbdce628d564c6af031b5f26cda6e818ea10)
- Base package dependencies; plugins disabled; no external services or input files

### Minimal reproduction

From a source checkout with the package installed (`python -m pip install ./packages/markitdown`):

```python
import io

from markitdown.converters import CsvConverter

info = StreamInfo(extension=".csv", charset="utf-8")
md = MarkItDown(enable_plugins=False)

for label, separator in [("LF", "\n"), ("CRLF", "\r\n"), ("CR", "\r")]:
    data = (separator.join(["name,age", "Alice,30"]) + separator).encode("utf-8")
    result = md.convert_stream(io.BytesIO(data), stream_info=info)
    print(label, repr(result.markdown))

# Isolate the CSV parser failure:
CsvConverter().convert(io.BytesIO(b"name,age\rAlice,30\r"), info)
```

### Expected and actual results

All three separators should produce:

```markdown
| name | age |
| --- | --- |
| Alice | 30 |
```

Observed public API output:

```text
LF '| name | age |\n| --- | --- |\n| Alice | 30 |'
CRLF '| name | age |\n| --- | --- |\n| Alice | 30 |'
CR 'name,age\rAlice,30'
```

The final direct converter call raises:

```text
_csv.Error: new-line character seen in unquoted field - do you need to open the file with newline=''?
```

The public API therefore appears to succeed, but table structure is lost.

### Suspected cause and proposed scope

[`_csv_converter.py:86`](https://github.com/microsoft/markitdown/blob/b6e8bbdce628d564c6af031b5f26cda6e818ea10/packages/markitdown/src/markitdown/converters/_csv_converter.py#L86) constructs the reader with `csv.reader(io.StringIO(content))`. The default `StringIO` newline setting does not split CR-only records as needed here.

Using `io.StringIO(content, newline="")` allows the standard CSV reader to parse this example correctly. The [Python CSV documentation](https://docs.python.org/3/library/csv.html#csv.reader) recommends `newline=""` for CSV file objects.

A focused fix would configure the stream's newline handling and add regression tests for LF/CRLF/CR record separators, optional final line endings, and quoted fields containing line breaks. The assertions should check the Markdown table through the public API so a plain-text fallback cannot count as success.

### Related work checked

Searched existing issues and PRs for CSV, CR-only, carriage returns, and newline handling. Related delimiter proposals #2021 and #2061 retain `io.StringIO(content)`; #2266 handles escaping line breaks **inside parsed cells**. These do not address CR-only record parsing.

AI assistance was used for the investigation and issue drafting. The reproduction above was executed against the referenced source checkout.

tkv00:
## Summary

Fixes #2411.

CSV input with CR-only (`\r`) record separators currently raises `_csv.Error` in `CsvConverter`. Through the public `MarkItDown.convert_stream()` API, conversion then falls back to plain text instead of producing a Markdown table.

Initialize the in-memory text stream with `io.StringIO(content, newline="")` so the CSV reader can handle LF, CRLF, and CR record separators without translating line endings inside quoted fields. The existing cell normalization and Markdown escaping are unchanged.

## Regression coverage

Add 15 parameterized cases through the public conversion API:

- LF, CRLF, and CR record separators, with and without a final newline (6 cases).
- Each record separator combined with LF, CRLF, or CR inside a quoted cell (9 cases).

Before the fix, the five CR-record-separator cases fail and the other ten pass. With the fix, all 15 pass.

## Validation

Environment: Windows, Python 3.12.14.

- New regression tests plus existing CSV tests: **29 passed, 52 deselected** (`python -m pytest packages/markitdown/tests/test_csv_line_endings.py packages/markitdown/tests/test_module_misc.py -k csv -q`, using the local source tree).
- `pre-commit run --all-files`: **passed** (Black).
- `git diff --check` and `git diff --cached --check`: **passed**.
- Full core suite (`hatch test -py 3.12` from `packages/markitdown`, with `GITHUB_ACTIONS=true` and no OpenAI API key): **425 passed, 34 skipped, 4 failed**.

All four full-suite failures also reproduce after restoring the CSV converter to the unchanged base revision (`b6e8bbdce628d564c6af031b5f26cda6e818ea10`):

- `test_output_to_stdout[test_vector6]` and `test_input_from_stdin_without_hints[test_vector6]`: Japanese CSV text cannot be represented faithfully by the local Windows default output encoding.
- `test_file_uris` and `test_convert_case_insensitive_uri_schemes`: tests expect POSIX paths, but Windows returns Windows-style paths.

These existing failures are outside this PR's scope. The local run also reported pytest cache permission warnings. Linux and other Python versions have not been verified locally.

## AI assistance

AI assistance was used to investigate the issue, prepare the fix and tests, and run validation.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -83,7 +83,7 @@ class CsvConverter(DocumentConverter):
         content = content.lstrip("\ufeff")
 
         # Parse CSV content
-        reader = csv.reader(io.StringIO(content))
+        reader = csv.reader(io.StringIO(content, newline=""))
         rows = list(reader)
         _trim_outer_blank_rows(rows)
 
--- a/packages/markitdown/tests/test_csv_line_endings.py
+++ b/packages/markitdown/tests/test_csv_line_endings.py
@@ -1,0 +1,53 @@
+"""Regression tests for CR-only CSV record separators (issue #2411)."""
+
+import io
+
+import pytest
+
+from markitdown import MarkItDown, StreamInfo
+
+
+@pytest.fixture(scope="module")
+def converter() -> MarkItDown:
+    return MarkItDown(enable_plugins=False)
+
+
+@pytest.mark.parametrize("newline", ["\n", "\r\n", "\r"], ids=["LF", "CRLF", "CR"])
+@pytest.mark.parametrize("final_newline", [False, True])
+def test_csv_record_separators_produce_a_table(
+    converter: MarkItDown, newline: str, final_newline: bool
+) -> None:
+    content = newline.join(["name,age", "Alice,30"])
+    if final_newline:
+        content += newline
+
+    result = converter.convert_stream(
+        io.BytesIO(content.encode("utf-8")),
+        stream_info=StreamInfo(extension=".csv", charset="utf-8"),
+    )
+
+    assert result.markdown == "| name | age |\n| --- | --- |\n| Alice | 30 |"
+
+
+@pytest.mark.parametrize("newline", ["\n", "\r\n", "\r"], ids=["LF", "CRLF", "CR"])
+@pytest.mark.parametrize(
+    "cell_newline", ["\n", "\r\n", "\r"], ids=["cell-LF", "cell-CRLF", "cell-CR"]
+)
+def test_csv_quoted_line_break_stays_inside_its_cell(
+    converter: MarkItDown, newline: str, cell_newline: str
+) -> None:
+    content = newline.join(
+        ["name,notes", f'Alice,"first{cell_newline}second"', "Bob,plain"]
+    )
+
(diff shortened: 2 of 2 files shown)


tkv00:

PR #2417, opened 2026-09-08
fix: prefer data-src over placeholder data URI in img src

macjayz:
convert_img falls back to data-src only when src is missing or empty. Lazy-loaded images commonly leave a placeholder data: URI in src and the real URL in data-src, so that case still lost the real image. This treats a data: URI in src the same as empty when data-src is present.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_markdownify.py
+++ b/packages/markitdown/src/markitdown/converters/_markdownify.py
@@ -115,7 +115,14 @@ class _CustomMarkdownify(markdownify.MarkdownConverter):
         """Same as usual converter, but removes data URIs"""
 
         alt = el.attrs.get("alt", None) or ""
-        src = el.attrs.get("src", None) or el.attrs.get("data-src", None) or ""
+        src = el.attrs.get("src", None) or ""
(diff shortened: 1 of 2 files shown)


Diogo-Damasceno (commented):
Verified locally (Arch Linux, Python 3.14.7, markitdown from source at `b6e8bbd`).

**Reproduced before the fix.** Applying only this PR's test file to `main` without the source change fails 1 test (`test_img_prefers_data_src_over_placeholder_data_uri`), so the placeholder data URI does currently win over the real URL in `data-src`.

**After the fix:** all 10 tests in `test_html_converter.py` pass. Wider selection (`-k "html or markdownify"`, ignoring `test_outlook_msg_ansi.py` which fails to collect here for an unrelated missing module): 24 passed / 2 failed with the change vs 20 passed / 2 failed on `main`. The 2 failures are the `test_docintel_*` tests, failing identically before and after, so they are pre-existing in my environment. No regressions.

Small note, not a blocker: `src[:5].lower() == "data:"` matches the scheme without requiring a colon-normalised parse. It is fine for real-world markup, and `DATA:`/`Data:` are covered by the `.lower()`.

PR #2418, opened 2026-09-08
Do not convert an undecodable file to the word "None"

kevin9327:
## What this changes

A binary file handed to MarkItDown with a text extension is converted to the four-character string
`None`.

`charset_normalizer.from_bytes(data).best()` returns `None` when nothing decodes the bytes, and
`str(None)` is `"None"`. Two converters write that expression inline:

```python
# _plain_text_converter.py
text_content = str(from_bytes(file_stream.read()).best())

# _csv_converter.py
content = str(from_bytes(file_stream.read()).best())
```

Measured, on 4096 bytes no charset claims:

```
.txt  ->  'None'
.md   ->  'None'
.csv  ->  '| None |\n| --- |'
```

That is content that was never in the file, produced by a tool whose output is read by a model. A
document that failed to decode reads downstream as a document whose contents are the word "None".

## The other two callers already guard it

This is the same call in four places, and the other two handle the `None`:

```python
# _outlook_msg_converter.py
detected = from_bytes(data).best()
if detected is not None:
    return str(detected).strip()
return data.decode("utf-8", errors="ignore").strip()

# _markitdown.py, _get_stream_info_guesses
charset_result = charset_normalizer.from_bytes(stream_page).best()
if charset_result is not None:
    charset = self._normalize_charset(charset_result.encoding)
```

The two that inlined `str(...)` are the two that skipped the check.

## Fix

Both now check, and fall back the way `_outlook_msg_converter` already does: decode as UTF-8 and
drop what does not fit. The CSV side gets a small `_decode_detected` helper so the reason is written
down once.

I chose the existing fallback over raising `FileConversionException` because it is the convention
already in the tree, and because the bytes are at least the file's own. If you would rather an
undecodable stream were refused outright, that is a one-line change and I am happy to make it.

Nothing else moves: a declared `stream_info.charset` still short-circuits detection, and any file
detection *does* claim decodes exactly as before.

## Tests

`packages/markitdown/tests/test_undetectable_charset.py`:

- the fixture really is undecodable, so the rest cannot pass for the wrong reason
- binary with `.txt` and `.md` is not the word "None"
- binary with `.csv` is not a one-cell table of "None"
- ordinary text and an ordinary CSV are byte-identical to before
- a declared `charset` still wins
- the same holds for `os.urandom` bytes, not only the fixed fixture

Against `main`:

```
E  AssertionError: assert 'None' != 'None'
E  AssertionError: assert '| None |\n| --- |' != '| None |\n| --- |'
```

## How I tested

Windows 11, Python 3.12, editable install of `packages/markitdown`.

```
pytest tests/test_undetectable_charset.py   ->  6 passed
pytest tests/test_module_misc.py            ->  15 failed, 48 passed, 3 skipped
```

Those 15 are unchanged by this PR — I ran that file with and without the two edited converters and
got `15 failed, 48 passed, 3 skipped` both times. They are the tests needing credentials and media
tooling this machine does not have (`test_speech_transcription`, `test_markitdown_llm_parameters`
and friends).

`black` clean on all three files.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -18,6 +18,24 @@ ACCEPTED_FILE_EXTENSIONS = [".csv"]
 _PIPE_ESCAPE_RE = re.compile(r"(\\*)\|")
 
 
+def _decode_detected(data: bytes) -> str:
+    """Decode ``data`` by detected charset, falling back rather than to ``None``.
+
+    ``from_bytes(...).best()`` returns ``None`` when nothing decodes the bytes,
+    and ``str(None)`` is the four-character string ``"None"``. A binary file
+    handed over with a text extension therefore became a document whose entire
+    content was the word "None" -- content that was never in the file, in a
+    tool whose output is read by a model.
+
+    The fallback is the one ``_outlook_msg_converter`` already uses for the same
+    call: decode as UTF-8 and drop what does not fit.
+    """
+    detected = from_bytes(data).best()
+    if detected is not None:
+        return str(detected)
+    return data.decode("utf-8", errors="ignore")
+
+
 def _escape_table_cell(value: str) -> str:
     r"""Escape a CSV value so it is safe inside a Markdown table cell.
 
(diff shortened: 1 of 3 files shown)


kevin9327:

PR #2419, opened 2026-09-08
fix(audio): honor extension case and common audio MIME aliases

MykolaBiron:
AudioConverter.accepts() writes the extension in lowercase and mimetype before the actual comparison happens. However, convert() has compared them verbatim. A file named "recording.WAV" was therefore accepted, emitted its metadata, and then silently skipped transcription.

The accepted mimetype list was also missing common aliases for formats already accepted by extension: audio/wav (far more common than audio/x-wav), audio/mp3, and audio/mp4, audio/m4a, audio/x-m4a. As a result, .m4a support depended entirely on the filename surviving. The alias set matches the one already used for these formats in _cu_converter.py.

Both the accept lists and the format lookup are now derived from a single extension/mimetype map.  They cannot drift apart after the fix, and convert() normalizes case the same way accepts() does.


MykolaBiron:

PR #2424, opened 2026-09-09
fix(rss): read an element's whole text, not only its first node

kevin9327:
## What this changes

An RSS item body written as a pretty-printed CDATA section is dropped. The item keeps its
heading; the content underneath it disappears.

```xml
<item>
  <title>Example item</title>
  <description>
    <![CDATA[<p>The <strong>body</strong> of the item.</p>]]>
  </description>
</item>
```

```
# Example feed
Example feed description

## Example item
```

That layout is ordinary in hand-written and CMS-generated feeds, and it is what WordPress
emits for `content:encoded`.

## Why

`_get_data_by_tag_name` reads one node:

```python
fc = nodes[0].firstChild
if fc:
    if hasattr(fc, "data"):
        return fc.data
return None
```

An element's text is not necessarily one node. The example above reaches the parser as three
children -- a text node holding `"\n    "`, the CDATA section, then another text node -- so
`firstChild` is the indentation and the value is never read. The same happens whenever a
CDATA section sits next to text:

```xml
<title>Quarterly <![CDATA[R&D]]> report</title>   ->  ## Quarterly
```

`_get_data_by_tag_name` backs `title`, `description`, `pubDate` and `content:encoded` on the
RSS side and `title`, `updated`, `summary` and `content` on the Atom side, so every one of
them truncates at the first CDATA boundary.

## Fix

Join the element's own text and CDATA children instead of reading only the first one. This is
what `_collect_node_text` in `_epub_converter` already does for the same minidom shape.

The join is deliberately not recursive: only the element's direct text and CDATA children are
read, which is exactly the set `firstChild` was sampling from. Element children keep behaving
as before -- `<summary type="xhtml">` is still handled by `_get_atom_content` before this
function is reached, and an element whose first child is markup still returns `None`.

## Tests

Added to `packages/markitdown/tests/test_rss_converter.py`:

- an RSS `<description>` whose CDATA sits on its own line
- an RSS `<title>` with a CDATA section in the middle of the text
- an Atom `<summary type="html">` whose CDATA sits on its own line
- pinned: a `<description>` that is nothing but a CDATA section, which already worked
- pinned: an item with no `<description>` at all, which must stay absent rather than become
  an empty string

Against unmodified `main`:

```
E  AssertionError: assert 'The **body** of the item.' in '# Example feed\nExample feed description\n\n## Example item\n'
E  AssertionError: assert '## Quarterly R&D report' in '# Example feed\nExample feed description\n\n## Quarterly \nBody.'
E  AssertionError: assert 'A *structured* summary.' in '# Example feed\n\n## Example entry\n'
3 failed, 14 passed
```

With the fix:

```
17 passed
```

## How I tested

```
pytest tests/test_rss_converter.py   ->  3 failed, 14 passed  (before)
pytest tests/test_rss_converter.py   ->  17 passed            (after)
pytest tests/                        ->  18 failed, 431 passed, 4 skipped
```

machine before any edit, and the 5 added tests account for the difference. They are the CLI
stdout-encoding, Windows file-URI and speech-transcription tests that need a UTF-8 console,
a case-sensitive path and network/ffmpeg.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_rss_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_rss_converter.py
@@ -310,12 +310,21 @@ class RssConverter(DocumentConverter):
     ) -> Union[str, None]:
         """Get data from first child element with the given tag name.
         Returns None when no such element is found.
+
+        An element's text is not necessarily a single node: a value written as
+        ``<description>\\n  <![CDATA[...]]>\\n</description>`` reaches the parser
+        as whitespace, then the CDATA section, then more whitespace. Reading
+        only the first of those returns the layout and drops the value, so all
+        of the element's own text and CDATA children are joined.
         """
         nodes = element.getElementsByTagName(tag_name)
         if not nodes:
             return None
-        fc = nodes[0].firstChild
-        if fc:
-            if hasattr(fc, "data"):
-                return fc.data
-        return None
+        parts = [
+            child.data
+            for child in nodes[0].childNodes
+            if child.nodeType in (Node.TEXT_NODE, Node.CDATA_SECTION_NODE)
+        ]
+        if not parts:
+            return None
+        return "".join(parts)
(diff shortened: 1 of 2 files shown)


afourney:
Addressed alternatively in #2432

PR #2429, opened 2026-09-09
fix(rss): support namespace-prefixed Atom feeds

Elysium-Seeker:
Atom feeds that use an explicit prefix, such as `<atom:feed>`, currently fail feed detection and can fall back to raw XML in `MarkItDown.convert_stream()`. Prefixes are arbitrary under [RFC 4287 section 1.3](https://www.rfc-editor.org/rfc/rfc4287.html#section-1.3), so these feeds should convert the same way as feeds using the default Atom namespace.

Match Atom elements by namespace URI and local name during detection and conversion. This covers feed metadata, entries, summaries, and content, while preserving support for unqualified feeds. The regression tests exercise fully prefixed and mixed-prefix feeds through both `RssConverter` and the public API, and check that elements from other namespaces are ignored.

Validation on Windows / Python 3.11.16:

- The initial regression test failed for all four prefixed variants before the fix; the default-namespace control passed.
- RSS/Atom tests: **20 passed** (`pytest tests/test_rss_converter.py tests/test_module_misc.py -k 'atom or rss'`).
- `hatch test --python 3.11`: **419 passed, 34 skipped, 2 failed**. Both failures (`test_file_uris` and `test_convert_case_insensitive_uri_schemes`) expect Unix paths on Windows and also fail in a clean worktree at the unchanged base commit `b6e8bbd`. Remote/LLM tests were disabled using the existing CI skip conditions.
- `pre-commit run --all-files`: **passed**.

Prepared by OpenAI Codex on behalf of Elysium-Seeker, including reproduction, implementation, and local checks.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_rss_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_rss_converter.py
@@ -29,6 +29,7 @@ CANDIDATE_FILE_EXTENSIONS = [
     ".xml",
 ]
 
+ATOM_NAMESPACE = "http://www.w3.org/2005/Atom"
 XHTML_NAMESPACE = "http://www.w3.org/1999/xhtml"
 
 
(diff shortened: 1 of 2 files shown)


Elysium-Seeker:

PR #2434, opened 2026-09-10
fix(zip): preserve content of entries with duplicate filenames

lyydsheep:
# fix(zip): preserve content of entries with duplicate filenames

ZIP archives can contain multiple entries with the same filename. ZipConverter iterated over names and passed each name to ZipFile.read(), so a duplicate name repeatedly selected the last matching entry. An archive containing notes.txt with content A followed by notes.txt with content B therefore produced B twice and silently lost A.

Iterate over ZipInfo entries and pass each entry to read() so every member retains its own content and archive order. Keep the existing filename-based headings and member conversion metadata.

Add an in-memory regression test through MarkItDown.convert_stream() that verifies the complete Markdown output for two same-named members. No new dependency or binary fixture is required.

## Validation

- Regression fails before the fix on upstream main f9042b4: the first member's content is replaced by the second member's content.
- Regression, existing ZIP header test, and test_zip_kwargs.py: 7 passed.
- Existing ZIP conversion vector across format detection, local files, streams with/without hints, file URIs, and data URIs: 6 passed; HTTP test skipped using the repository's GITHUB_ACTIONS setting.
- Black 23.7.0 check on both changed files and git diff --check passed.
- Full repository test suite was not run.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_zip_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_zip_converter.py
@@ -106,9 +106,10 @@ class ZipConverter(DocumentConverter):
         }
 
         with zipfile.ZipFile(file_stream, "r") as zipObj:
-            for name in zipObj.namelist():
+            for entry in zipObj.infolist():
+                name = entry.filename
                 try:
-                    z_file_stream = io.BytesIO(zipObj.read(name))
+                    z_file_stream = io.BytesIO(zipObj.read(entry))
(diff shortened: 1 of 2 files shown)


afourney:
This is an interesting one. I think most zip extractors also will, by default, keep the last entry. I feel like printing them both might be unexpected.

An alternative behavior here would be to avoid printing the file contents twice, and just keep the last one. 

Let me think about this, and also investigate the behavior of common extractors.

PR #2435, opened 2026-09-10
fix(html): keep links inside inline code as literal text

nickita-khylkouski:
hey, found a small issue with linked identifiers in API docs.
`<code><a href="Action.html">Action</a></code>` currently comes out as `[Action](Action.html)` instead of `Action`. The link syntax ends up inside the backticks, so it becomes part of the code. This markup also appears in OpenJDK’s JMenu docs. This adds an early return for anchors inside code, kbd, and samp, before trimming or link formatting. Regular links still work, including links wrapped around code spans. Kept this separate from #2396’s whitespace fix. These inline-code cases still fail with that PR applied, and pass with this change added. Added six regression cases and three controls: six fail before the fix, all nine pass after. Checked the real converter source locally on Python 3.13 with markdownify 1.2.2 using a source-subset import harness; I haven’t run the full Hatch suite yet. Prepared with ChatGPT assistance.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_markdownify.py
+++ b/packages/markitdown/src/markitdown/converters/_markdownify.py
@@ -61,6 +61,9 @@ class _CustomMarkdownify(markdownify.MarkdownConverter):
         **kwargs,
     ):
         """Same as usual converter, but removes JavaScript links and escapes URIs."""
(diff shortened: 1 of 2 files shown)


nickita-khylkouski:

PR #2439, opened 2026-09-10
fix(html): escape pipe characters in table cells

### LINKED ISSUE #2438: HTML/DOCX: pipe characters in table cells break Markdown tables
A literal pipe in an HTML table cell is data, not a column separator, but it is passed through untouched, so the rendered row gains phantom columns. DOCX tables are affected too since they convert via the HTML converter.

Repro:

```python
import io

html = (
    '<table><thead><tr><th>Name</th><th>Note</th></tr></thead>'
    '<tbody><tr><td>Alice</td><td>Has a | pipe</td></tr></tbody></table>'
)
print(MarkItDown().convert_stream(
    io.BytesIO(html.encode()), file_extension='.html').markdown)
# Before: | Alice | Has a | pipe |   (3 columns, broken)
# After:  | Alice | Has a \| pipe |  (2 columns, correct)
```

Same class as the CSV (#2266) and XLSX (#2436) pipe bugs, but in the shared HTML layer (`_CustomMarkdownify`), which neither of those covers. Searched open and closed issues/PRs: no report or fix for HTML/DOCX table pipes (open #1785 covers PPTX only, via pre-escaping).

soroush5:
Fixes #2438.

markdownify passes table cell text through untouched, so a literal pipe in a cell splits the Markdown row into phantom columns. This adds `convert_td`/`convert_th` overrides on `_CustomMarkdownify` that escape unescaped pipes, mirroring the base logic otherwise (colspan handling, newline collapsing).

The match is parity-correct: already-escaped pipes are left alone, so converters that pre-escape before reaching this layer (XLSX via `_escape_sheet`, PPTX in #1785) are unaffected — no double escaping. DOCX tables are fixed as well since they convert via the HTML converter (verified end-to-end with a .docx containing a pipe in a table cell).

Tests: 3 new cases in `test_html_converter.py` (cell pipe, header pipe, no double-escape). The first two fail before the fix, all pass after. Full package suite green: 649 passed, 4 skipped.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_markdownify.py
+++ b/packages/markitdown/src/markitdown/converters/_markdownify.py
@@ -7,6 +7,21 @@ from urllib.parse import quote, urlparse, urlunparse
 
 _PERCENT_ENCODED_OCTET = re.compile(r"%[0-9A-Fa-f]{2}")
 
+# Matches a pipe together with the (possibly empty) run of backslashes in
+# front of it, anchored so the run starts after a non-backslash (or at the
+# start of the string). The run therefore holds *every* backslash in front
+# of the pipe, which makes the substitution parity-correct: a pipe preceded
+# by an even run (including zero) is unescaped and gains one backslash,
+# while an already-escaped pipe (odd run) matches nothing and is left alone.
+# Idempotency matters because some converters (e.g. XLSX via _escape_sheet,
+# PPTX via html-escaped cells) pre-escape pipes before the HTML reaches us.
+_TABLE_PIPE_RE = re.compile(r"(?<!\\)((?:\\\\)*)\|")
+
+
+def _escape_table_cell_pipes(text: str) -> str:
+    """Escape unescaped pipes so cell text cannot split a Markdown table."""
+    return _TABLE_PIPE_RE.sub(r"\1\\|", text)
+
 
 def _quote_path_preserving_percent_encoded_octets(path: str) -> str:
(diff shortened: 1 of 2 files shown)


soroush5:

erekola:
I tested `d5d07b1` with four two-column HTML tables and rendered the Markdown with `markdown-it-py 4.2.0`, with table support enabled. The first cell contained a plain pipe, a pipe inside inline code, a pipe in link text or an HTML pipe entity. The second cell was always `end`. All four lose that control cell on current main. On this PR all four keep both cells, and the inline code and link still render as code and a link. The converter dependency was `markdownify 1.2.3` on Windows with Python `3.13.15`.

The inline-code case is a useful extra regression fixture because checking the rendered cells catches content loss that counting separators misses:

```html
<table><tr><th>Value</th><th>Control</th></tr><tr><td><code>a|b</code></td><td>end</td></tr></table>
```

Expected body cells after rendering are `<td><code>a|b</code></td>` and `<td>end</td>`. I ran four parameterized assertions against main and the PR checkout: four failed on main and four passed on the PR. I haven't tested DOCX or the full suite. AI assistance was used for the fixtures, checks and this comment.

soroush5:
Thanks for verifying this so thoroughly. Good call on the inline-code case, I added it as a regression test in test_html_table_pipe_in_inline_code_is_escaped. All 15 tests in the file pass.

soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

PR #2448, opened 2026-09-10
Make optional converters opt-in for MarkItDown MCP

### LINKED ISSUE #2440: markitdown-mcp: mandatory markitdown[all] breaks native Windows ARM64 installation
## Summary

`markitdown-mcp` currently has a mandatory dependency on:

```toml
"markitdown[all]>=0.1.1,<0.2.0"
```

This makes installing the MCP server on native Windows ARM64 depend on every optional MarkItDown converter, even when those converters are not needed.

On a native ARM64 Surface Pro with CPython 3.14, the base `markitdown` package installs and works correctly, including native `onnxruntime`. However, installing the current `markitdown-mcp` dependency set pulls in the PDF stack and eventually `cryptography==50.0.1`.

There is no `win_arm64` wheel for that current cryptography release, so the installer falls back to compiling it from source and requires a native Rust/OpenSSL build environment.

## Environment

```text
OS: Windows on ARM64, Surface Pro
Python: CPython 3.14 ARM64
Python package path: Python314-arm64
markitdown: 0.1.5
markitdown-mcp: 0.0.1a4
onnxruntime: 1.29.0 ARM64
```

The important part is that this is an entirely native ARM64 Python environment; no x64 emulation is involved.

## Reproduction

For example:

```powershell
uvx markitdown-mcp
```

eventually attempts to build:

```text
cryptography==50.0.1
```

from source:

```text
🍹 Building a mixed python/rust project
🐍 Found CPython 3.14
...
error: failed to run custom build command for `openssl-sys`
...
$HOST = aarch64-pc-windows-msvc
$TARGET = aarch64-pc-windows-msvc
...
Could not find directory of OpenSSL installation
```

The dependency chain reported by the installer is:

```text
markitdown-mcp
  -> markitdown[all]
    -> pdfminer-six
      -> cryptography
```

By contrast, the base package installs normally:

```powershell
pip install markitdown
```

and runs natively with ARM64 dependencies:

```text
markitdown 0.1.5
magika 0.6.2
onnxruntime 1.29.0
```

The `markitdown-mcp` wheel itself is also pure Python and can be installed without this failure when its dependencies are handled separately.

## Why this seems undesirable

The MCP server itself does not inherently require PDF, Office, Azure, audio transcription, YouTube transcription, etc. support just to start and expose `convert_to_markdown`.

Making `markitdown[all]` mandatory means an optional converter dependency can make the entire MCP server uninstallable on an otherwise supported Python/platform combination.

It also makes a lightweight local STDIO MCP installation substantially heavier than necessary.

## Proposed direction

I think the cleanest dependency model would be for the MCP package to depend on base MarkItDown:

```toml
dependencies = [
    "mcp>=...",
    "markitdown>=0.1.1,<0.2.0",
]
```

and expose full converter support as an explicit extra, for example:

```toml
[project.optional-dependencies]
all = [
    "markitdown[all]>=0.1.1,<0.2.0",
]
```

so users could choose:

```bash
pip install markitdown-mcp
```

for the lightweight/native server, or:

```bash
pip install "markitdown-mcp[all]"
```

when they explicitly want every optional converter.

Mirroring individual MarkItDown extras could also be considered later, but does not seem necessary to solve the underlying installation problem.

I would avoid solving this with an old `cryptography` pin: this is fundamentally an optional-dependency boundary issue, and pinning an older crypto package would create a maintenance/security liability.

## Related

#2363 migrates `markitdown-mcp` to MCP SDK 2.x. That fixes a separate compatibility problem but currently retains the mandatory `markitdown[all]` dependency, so the Windows ARM64 installation problem remains independent of that PR.

sylvesterkaczmarek:
Fixes #2440.

Depend on base MarkItDown instead of requiring every optional converter. `markitdown-mcp[all]` provides the previous full converter set, and the Dockerfile explicitly installs that extra to preserve its behavior. Document base, selective, and full installations.

Adds dependency-metadata and HTML-tool regressions, with a clean base-install CI job across Python 3.10 through 3.13. The MCP SDK constraint is unchanged; #2363 remains independent.

Validation:
- Both metadata regressions fail before the change.
- Built wheel; all three tests pass with Python 3.10 and 3.12, including base and full-converter environments on 3.12.
- Installed-wheel STDIO initialization, tool discovery, and HTML conversion pass.
- Python 3.14 / Windows ARM64 binary-only dependency resolution succeeds without the optional PDF/Office stack.
- Black and `git diff --check` pass.

Native Windows ARM64 execution and a Docker image build were not tested locally.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -84,8 +84,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      # Keep this environment free of optional converters to exercise the base install.
       - name: Install packages
-        run: pip install "./packages/markitdown[all]" ./packages/markitdown-mcp pytest
+        run: pip install ./packages/markitdown ./packages/markitdown-mcp pytest
 
       - name: Run tests
         working-directory: packages/markitdown-mcp
--- a/packages/markitdown-mcp/Dockerfile
+++ b/packages/markitdown-mcp/Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get update \
 
 COPY . /app
 
-RUN pip install --no-cache-dir /app
+RUN pip install --no-cache-dir "/app[all]"
 
 WORKDIR /workdir
 
--- a/packages/markitdown-mcp/README.md
+++ b/packages/markitdown-mcp/README.md
@@ -14,11 +14,37 @@ It exposes one tool: `convert_to_markdown(uri)`, where uri can be any `http:`, `
 
 ## Installation
 
-To install the package, use pip:
+The base installation includes the MCP server and MarkItDown's core converters,
+such as text and HTML. Optional document and media converters are installed
+separately. This avoids requiring platform-specific dependencies for formats
+you do not use.
+
+To install the base package, use pip:
 
 ```bash
 pip install markitdown-mcp
 ```
+
+To install every optional converter (the previous default), use:
+
+```bash
+pip install "markitdown-mcp[all]"
+```
+
+For selected formats, install the corresponding MarkItDown extras alongside the
+server. For example, to add PDF support:
+
+```bash
+pip install markitdown-mcp "markitdown[pdf]"
+```
+
+With `uvx`, select the extra explicitly when full converter support is needed:
+
+```bash
+uvx --from "markitdown-mcp[all]" markitdown-mcp
+```
+
+The provided Docker image continues to include all optional converters.
 
 ## Usage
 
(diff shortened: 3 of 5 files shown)


afourney:
I'm working to merge #2363 right now. Let me deal with that, then circle back here

sylvesterkaczmarek:
@afourney #2363 has now landed, so I rebased this onto current `main` and resolved the MCP 2.x and CI conflicts while preserving this PR’s opt-in converter scope. The branch is mergeable again.

PR #2450, opened 2026-09-10
Trim CSV blank runs without repeatedly shifting the row list

### LINKED ISSUE #2449: CSV conversion scales quadratically with leading blank lines
CSV conversion slows down quadratically when a file contains a long run of blank lines before the header or immediately after it. The output is correct, but trimming those rows takes most of the conversion time.

This script uses only in-memory CSV data:

```python
import io
import statistics
import time

converter = MarkItDown(enable_plugins=False)
info = StreamInfo(extension=".csv", charset="utf-8")

for count in (25_000, 50_000, 100_000):
    data = b"\n" * count + b"name,value\nexample,1\n"
    samples = []
    for _ in range(3):
        start = time.perf_counter()
        result = converter.convert_stream(io.BytesIO(data), stream_info=info)
        samples.append(time.perf_counter() - start)
    print(count, statistics.median(samples))
```

On Windows with Python 3.13.15 and MarkItDown 0.1.8b1 (source commit b59e0643dc865375cb9bd3f001ce0cc711d70459), the medians were 0.0296 s, 0.1170 s, and 0.6953 s. These are synthetic inputs, not timings from a production workload.

Since these inputs differ only in the number of blank lines, I'd expect conversion time to grow roughly linearly. `_trim_outer_blank_rows` currently calls `pop(0)` or `pop(1)` for every blank row, shifting the remaining list each time. Removing each blank run in one slice avoids that repeated work while preserving the output.

Prepared with OpenAI Codex assistance.

ROTl24:
Fixes #2449.

CSV conversion repeatedly shifts the row list when removing blank lines before the header or immediately after it. A long blank run therefore takes quadratic time even though the resulting table is tiny.

This change finds the boundaries first and deletes each run in one slice. The helper still mutates the original list and preserves the header, internal blank rows, and existing row objects.

Four conversion tests cover 100,000 blank lines before the header, after the header, at the end, and throughout the file. The table cases also check that an internal blank row and a wider data row survive. There is no timing assertion in the tests.

Validation on Windows / Python 3.13.15:

- 37 CSV tests passed; `pre-commit run --all-files` passed.
- An exhaustive comparison of 511 blank/nonblank row arrangements matched the original helper, including row identity.
- Using the reproduction in #2449, the median for 100,000 leading blank lines dropped from 0.6953 s to 0.0084 s across three runs. This is a synthetic case, not a general document-conversion benchmark.

The full optional-format and remote-service test suite was not run locally.

### DIFF
--- a/packages/markitdown/src/markitdown/converters/_csv_converter.py
+++ b/packages/markitdown/src/markitdown/converters/_csv_converter.py
@@ -30,17 +30,27 @@ def _escape_table_cell(value: str) -> str:
 
 def _trim_outer_blank_rows(rows: list[list[str]]) -> None:
     """Remove empty rows from the beginning and end, and immediately after the header. This operation is performed in-place."""
-    # Pop empty rows from the beginning
-    while len(rows) > 0 and not rows[0]:
-        rows.pop(0)
+    start = 0
+    while start < len(rows) and not rows[start]:
+        start += 1
 
-    # Pop empty rows after the header
-    while len(rows) > 1 and not rows[1]:
-        rows.pop(1)
+    if start == len(rows):
+        rows.clear()
+        return
 
-    # Pop empty rows from the end
-    while len(rows) > 0 and not rows[-1]:
-        rows.pop(-1)
+    header_index = start
+    start += 1
+    while start < len(rows) and not rows[start]:
+        start += 1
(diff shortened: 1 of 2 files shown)


ROTl24:

Issue #1073, opened 2025-02-28
Exclude Hidden Sheets in Excel Conversion


matteo-tafuro:
When processing Excel files (`.xlsx`) using the library, the current implementation reads all sheets—including hidden ones—using the following call:

```python
sheets = pd.read_excel(local_path, sheet_name=None, engine="openpyxl")
```

This behavior results in hidden sheets being processed and included in the final Markdown output, which is not the expected outcome in most cases.

**Steps to Reproduce:**  
1. Create or use an Excel workbook that contains both visible and hidden sheets.  
2. Run the conversion using the `XlsxConverter`.  
3. Observe that the output Markdown includes content from both visible and hidden sheets.

**Expected Behavior:**  
Only visible sheets should be processed by default. Ideally, there should be an option to toggle the inclusion of hidden sheets. Filtering out hidden sheets would prevent unintended content from appearing in the converted Markdown.

**Potential Solution:**  
One approach could be to use `openpyxl` to load the workbook and check each sheet’s `sheet_state` property. For example, filtering the sheets like so:

```python
wb = load_workbook(local_path, read_only=True, data_only=True)
visible_sheet_names = [sheet.title for sheet in wb.worksheets if sheet.sheet_state == 'visible']
```

Then pass `visible_sheet_names` to `pd.read_excel` instead of reading all sheets. Alternatively, adding an optional parameter to control this behavior would provide more flexibility.

Thank you for considering this enhancement!

SAGV:
This is indeed an issue because then LLM looks at the hidden sheets and doesn't know they were hidden…

PR #2456, opened 2026-09-11
Add opt-in filtering for hidden XLSX worksheets

aprylewu:
Related to #1073.

XLSX conversion currently includes worksheets marked `hidden` or `veryHidden`, so content intended to stay out of the visible workbook also appears in Markdown sent to an LLM. The built-in XLSX converter now accepts `include_hidden_sheets=False` through the public conversion API:

```python
result = md.convert("test.xlsx", include_hidden_sheets=False)
```

The default still includes every worksheet for compatibility with existing callers. This implements the issue's requested filtering option; changing the default can be considered separately.

Visible worksheets retain their workbook order. Filtering happens before cell parsing, so excluded sheets do not need to be read into DataFrames. The existing repair for legacy `showZeroes` sheet views preserves the filtering option when retrying. No new dependencies are added.

Validation:
- Five new in-memory workbook regressions: **3 failed / 2 passed before**, **5 passed after**. They cover both hidden states, visible-sheet order, explicit/default inclusion, the legacy repair path, and an invalid numeric cell in an excluded sheet.
- Existing XLSX regression tests: **2 passed**.
- XLSX/XLS API and CLI conversion vectors: **18 passed**.
- HTML converter tests: **11 passed**.
- `pre-commit run --all-files` and `git diff --check`: passed.

Developed and tested with OpenAI Codex assistance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,13 @@ result = md.convert("test.xlsx")
 print(result.markdown)
 ```
 
+The built-in XLSX converter includes all worksheets by default. To exclude
+worksheets marked `hidden` or `veryHidden`, pass `include_hidden_sheets=False`:
+
+```python
+result = md.convert("test.xlsx", include_hidden_sheets=False)
+```
+
 Document Intelligence conversion in Python:
 
 ```python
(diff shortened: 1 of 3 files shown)


aprylewu:

PR #2457, opened 2026-09-11
fix(docx): repair unindexed numbering definitions in docx preprocessor

EtherK3N:
When converting DOCX files generated by certain tools (such as Lark/Feishu or third-party web exporters), multi-level lists lose their hierarchy or have sub-levels dropped.

Looking into the OOXML structure, these documents define `<w:lvl>` elements inside `<w:abstractNum>` without explicit `w:ilvl` attributes. While Mammoth attempts a fallback to `"0"`, its internal dictionary only retains a single unindexed level per abstract definition, overwriting and discarding any subsequent sub-levels (1, 2, etc.).

This patch adds `_pre_process_numbering()` to `pre_process.py`, following the same approach already used in `_pre_process_styles`. It walks `word/numbering.xml` and assigns sequential indices (`str(idx)`) to any unindexed `<w:lvl>` tags before Mammoth parses the file. If all levels already carry `w:ilvl` (standard Word files) or if no numbering definitions exist, the content is returned untouched with no overhead.

Tested against real-world sample files and added a unit test in `test_module_vectors.py`. Formatting verified with `black`.


EtherK3N:

PR #2475, opened 2026-09-12
Avoid treating extensionless logs as CSV

### LINKED ISSUE #2474: Bug: Extensionless log streams can be rendered as CSV tables
### Bug

Extensionless plain-text log streams can be detected as CSV and rendered as a one-column Markdown table.

### Reproduction

Given a log-like stream with no filename or extension:

```text
2026-09-12 18:00:01 INFO boot ok
2026-09-12 18:00:02 ERROR disk full
```

Running through stdin/stream conversion can produce:

```markdown
| 2026-09-12 18:00:01 INFO boot ok |
| --- |
| 2026-09-12 18:00:02 ERROR disk full |
```

### Expected behavior

When there is no filename, extension, MIME hint, or CSV-style delimiter evidence, decodable plain text should stay plain text instead of being rendered as a CSV table.

Extensionless comma-delimited CSV should still convert to a Markdown table.

### Notes

This appears to be a content-detection precedence issue rather than a missing file-extension mapping. Magika can infer `text/csv` for extensionless text, and `CsvConverter` currently accepts that before `PlainTextConverter` can use the detected charset.

SHAI-shivansh-sharma:
## Summary

- Avoid treating extensionless plain-text logs from streams/stdin as CSV tables when the content has no CSV-style delimiter.
- Keep extensionless comma-delimited CSV converting to a Markdown table.
- Keep explicit `.csv` inputs on the existing CSV converter path.

Fixes #2474.

## Test

- `uv run --with pytest --with-editable . --python 3.12 pytest tests\test_csv_detection.py tests\test_csv_line_endings.py tests\test_csv_blank_runs.py`

### DIFF
--- a/packages/markitdown/src/markitdown/_markitdown.py
+++ b/packages/markitdown/src/markitdown/_markitdown.py
@@ -5,6 +5,7 @@ import sys
 import shutil
 import traceback
 import io
+import csv
 from dataclasses import dataclass
 from email.message import Message
 from email.utils import collapse_rfc2231_value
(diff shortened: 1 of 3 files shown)


SHAI-shivansh-sharma:

SHAI-shivansh-sharma:
Thanks for the detailed check. You were right that the original `.sql` / `.log` extension-list change did not materially improve the normal CLI or `MarkItDown().convert(...)` path.

I updated the PR to target the stronger issue you found instead: extensionless log/plain-text streams being inferred as CSV and rendered as a one-column Markdown table. The updated patch now downgrades content-only `text/csv` guesses to `text/plain` when there is no CSV-style delimiter evidence, while keeping extensionless comma-delimited CSV converting as a table.

I also added focused coverage for both cases and reran the existing CSV regression tests.

PR #2489, opened 2026-09-14
fix: decode percent-encoded data URI parameters

xblwh:
`MarkItDown(enable_plugins=False).convert_uri("data:text/plain;charset=iso%2D8859%2D1,Caf%E9")` currently raises `FileConversionException` because the escaped charset reaches the decoder unchanged. An escaped parameter name, such as `%63harset`, can instead cause the converter to guess the wrong charset. Both forms now preserve the declared charset and return `Café`.

Decode parameter names and values after splitting the metadata, following [RFC 2397, section 3](https://www.rfc-editor.org/rfc/rfc2397.html#section-3). This keeps escaped delimiters inside their parameters, preserves literal `+` characters, and decodes escapes once. The regression tests cover both base64 and non-base64 payloads, case normalization, escaped delimiters, empty values, and the public conversion API.

Related: #2486 covers percent-encoded base64 payloads.

Validation on macOS with Python 3.12.14:

- Core package: `GITHUB_ACTIONS=1 OPENAI_API_KEY='' hatch test --python 3.12 -q --tb=short` — **882 passed, 43 skipped**. The skips follow the repository's existing test conditions; live external services were not tested.
- `pre-commit run --all-files` — passed (Black 23.7.0).
- The 19 new regression cases produced 17 failures and 2 passes before the fix; all 19 pass after it.

Prepared with OpenAI Codex assistance.


xblwh:

Issue #1129, opened 2025-03-15
Add support for Google's Gemini and Anthropic's Claude models


Eigilak:
# Add support for Google's Gemini and Anthropic's Claude models

## Description
Currently, markitdown only supports OpenAI models for image captioning and content extraction. It would be valuable to add support for other leading multimodal LLMs, specifically:

1. Google's Gemini models (Pro and Ultra)
2. Anthropic's Claude models (Opus, Sonnet, and Haiku)

This would provide us with more flexibility and choice based on their preferences, API access, pricing, or specific model strengths.

## Motivation
- Different users have access to different AI provider APIs
- Some users may prefer the strengths of a particular model family
- Pricing and rate limits vary between providers
- Organizations may have existing enterprise agreements with Google or Anthropic

## Current implementation
Currently, `_image_converter.py` has hardcoded OpenAI-specific client API calls:

```python
# Prepare the OpenAI API request
messages = [
    {
        "role": "user",
        "content": [
            {"type": "text", "text": prompt},
            {
                "type": "image_url",
                "image_url": {
                    "url": data_uri,
                },
            },
        ],
    }
]

# Call the OpenAI API
response = client.chat.completions.create(model=model, messages=messages)
return response.choices[0].message.content
```

## Proposed solution
Create an abstraction layer for LLM providers that would:

1. Detect the client type (OpenAI, Google, or Anthropic)
2. Use the appropriate API format for each provider
3. Extract the response content consistently

This could be implemented either:
- As provider-specific adapter classes
- Through a simple detection mechanism based on client type
- Or potentially by leveraging [[Semantic Kernel](https://github.com/microsoft/markitdown/issues/232)](https://github.com/microsoft/markitdown/issues/232) as suggested in a related issue

## Implementation considerations
- Google's Gemini API requires a different message structure
- Anthropic's API has a slightly different format than OpenAI's
- Response parsing differs between providers
- Error handling should be consistent across providers

## Related issues
- [[Issue #232: Suggestion to use Semantic Kernel for different LLM providers](https://github.com/microsoft/markitdown/issues/232)](https://github.com/microsoft/markitdown/issues/232)
- [[Issue #12: LLM Integration for image understanding](https://github.com/microsoft/markitdown/issues/12)](https://github.com/microsoft/markitdown/issues/12)

afourney:
Thanks for the issue. 100% agree we need to abstract away the LLM client -- or perhaps even the idea of using an LLM. 

[...]

PR #2496, opened 2026-09-15
Add Claude (Anthropic) as an OCR backend for markitdown-ocr

youngh82:
Related: #1129

## Why

`markitdown-ocr` extracts text from images embedded in PDF/DOCX/PPTX/XLSX by sending
them to a vision model. Today that model has to be reachable through the OpenAI
chat-completions shape (`client.chat.completions.create`), because the OCR service
calls it directly. Claude is not, so users on Anthropic cannot use the plugin at all.

This adds a second backend that speaks the Anthropic Messages API, and picks between
the two from the client that was passed in.

## What changed

**`markitdown-ocr`**

- `AnthropicVisionOCRService` — same `extract_text()` interface as
  `LLMVisionOCRService`, built on `client.messages.create` with a base64 `image`
  content block. Handles the fixed set of media types the Messages API accepts, and
  treats a `stop_reason == "refusal"` response as an error rather than empty text.
- `create_ocr_service(client, model, default_prompt)` — selects the backend by
  duck-typing the client (`chat.completions.create` → OpenAI, `messages.create` →
  Anthropic), and raises a clear `ValueError` for anything else.
- New `ocr_llm_client` / `ocr_llm_model` / `ocr_llm_prompt` kwargs, falling back to
  `llm_client` / `llm_model` / `llm_prompt`.

  These exist because MarkItDown's own converters (`_image_converter.py`,
  `_pptx_converter.py`) call the OpenAI shape directly. Passing an Anthropic client
  as `llm_client` would satisfy the plugin but break image descriptions, so OCR gets
  its own client slot. The two can also be mixed — OpenAI for descriptions, Claude
  for OCR.
- `anthropic` optional dependency group; README and changelog updated.

The MCP server needs a small companion change before any of this is reachable over
MCP — `MARKITDOWN_ENABLE_PLUGINS` loads the plugin, but with no client reaching it OCR
silently no-ops. That's a separate concern from this PR (and `markitdown-mcp` has just
been migrated to SDK 2.x), so I'll raise it on its own if this direction is welcome.

## Backward compatibility

Additive only. `LLMVisionOCRService` keeps its constructor, interface, and
`backend_used="llm_vision"` value; existing `llm_client` / `llm_model` code paths are
unchanged. Two internals were factored out of the existing service and are now shared
by both backends: the default prompt (`_DEFAULT_PROMPT`) and content-type sniffing
(`_sniff_content_type`).

## Tests

19 new tests, no network and no API key — both backends run against fake clients:

- `tests/test_ocr_service.py` — backend selection; the Anthropic request shape
  (base64 image block, media-type fallback, model, `max_tokens`); multi-block and
  non-text-block responses; refusal; API failure surfaced as `OCRResult.error`; no
  client; stream rewind; plus regression coverage for the OpenAI backend's data URI.
- `tests/test_plugin_backend_selection.py` — `register_converters` wiring: no client
  leaves OCR disabled, `llm_client` selects OpenAI, `ocr_llm_client` selects
  Anthropic and takes precedence, `ocr_llm_prompt` overrides `llm_prompt`.

`pytest tests/` in `packages/markitdown-ocr`: 72 passed (53 existing + 19 new).
`black` clean at the pinned pre-commit version.

## Not covered

The backends are verified against fake clients only — no live API call is made in the
test suite, matching how the existing converter tests mock the OCR service.

## Open questions

#1129 asked for exactly this and has been open since March 2025, with a maintainer
comment agreeing the LLM client should be abstracted. I'd rather ask than assume:

1. **Plugin first, or core first?** This lands the backend in the plugin only;
   MarkItDown's core converters stay OpenAI-only. If you'd rather the abstraction go
   into `_image_converter.py` first and have the plugin inherit it, I'm happy to redo
   it in that order — I started here because it's the smaller, reversible change.
2. **Duck-typing, or a real provider interface?** Selecting the backend by client
   shape adds no dependency and handles the two shapes in ~40 lines, but it won't
   generalize to Gemini/Ollama the way an abstraction would (#1129 suggests
   `pydantic-ai` for that). If you'd prefer the interface, I'd rather write that than
   have this merged and replaced later.
3. **Is `ocr_llm_client` the right name and shape?** It's the only new public surface
   here, and it exists to work around the core being OpenAI-only. If it would clash
   with the eventual core API, I'd like to change it now.

Happy to defer to @Ritanshu-Kumar if they've already started on the core abstraction
in #1129.


youngh82:

PR #2497, opened 2026-09-15
Fix legacy showZeroes handling in OCR XLSX converter

### LINKED ISSUE #2400: OCR XLSX converter still fails on legacy showZeroes sheet views
## Summary

`XlsxConverterWithOCR` still fails on legacy workbooks whose worksheet view uses the historical `showZeroes` attribute, even though the core XLSX converter was fixed for this in #2064.

The OCR converter calls `pandas.read_excel(..., engine="openpyxl")` / `load_workbook()` directly and does not run the core converter's scoped `showZeroes` -> `showZeros` repair first.

## Reproduction on current main

I built the same malformed worksheet shape covered by `test_xlsx_legacy_show_zeroes_sheetview`: a normal workbook whose `xl/worksheets/sheet1.xml` contains:

```xml
<sheetView showZeroes="0" ...>
```

Local result against the current source packages:

```text
core: ## Data | | hello | world | | | --- | --- |
ocr error: TypeError "SheetView.__init__() got an unexpected keyword argument 'showZeroes'"
```

So enabling the OCR plugin reintroduces a workbook compatibility failure that the normal converter now handles.

## Proposed scope

If this is available to take, I can submit a focused OCR-side follow-up that reuses the same narrowly-scoped worksheet repair before `openpyxl`/pandas parses the workbook, plus a regression test derived from the existing core test. I would avoid broad XLSX refactoring.

I searched current PRs/issues for an OCR-specific `showZeroes` fix and found no matching implementation.

AI-assisted audit disclosure: found by comparing recent merged core fixes against sibling OCR converters; reproduction was executed locally against current source.

Ofcnitin:
## Summary

Fixes #2400.

- Reuse the existing XLSX `showZeroes` → `showZeros` repair in the OCR XLSX converter.
- Apply the repair when `openpyxl` fails to load a legacy workbook using `showZeroes`.
- Add a regression test covering the legacy worksheet XML format.

## Testing

- Added a regression test using a mock OCR service to exercise the OCR conversion path.


Ofcnitin:

PR #2499, opened 2026-09-15
Preserve literal NA values and blank cells in Excel conversions

### LINKED ISSUE #2498: Excel conversion replaces literal NA/NULL strings and blank cells with NaN
Excel cells containing literal strings such as `NA`, `NULL`, `None`, `n/a`, and `nan` are converted to `NaN`. Real blank cells also appear as `NaN`. This loses distinctions in spreadsheet data, for example a region code `NA` versus an empty region.

Reproduced on main at `eb31b5c9453628def5e6758a27a8e3a87b4ab101`:

```python
import io
from openpyxl import Workbook

workbook = Workbook()
workbook.active.append(["region", "status", "empty"])
workbook.active.append(["NA", "NULL", None])
stream = io.BytesIO()
workbook.save(stream)
stream.seek(0)
print(MarkItDown().convert_stream(stream, file_extension=".xlsx").markdown)
```

The data row is `| NaN | NaN | NaN |`; it should be `| NA | NULL |  |`.

Both XLSX and XLS converters use pandas' default missing-value recognition, which is useful for analysis but changes literal document text. `keep_default_na=False` preserves these strings and leaves blank cells empty. The XLSX fallback that repairs legacy `showZeroes` attributes also needs the same option.

I reproduced the failure for XLS, XLSX, and repaired XLSX and prepared a focused fix with regression tests. No model or remote service is needed. AI assistance: OpenAI Codex.

pentaoa:
Fixes #2498.

An Excel row containing `NA`, `NULL`, and a blank cell currently becomes `| NaN | NaN | NaN |`. Disable pandas' default NA-token recognition so literal cell text is preserved and blanks remain empty. Apply the same setting to XLS, XLSX, and the legacy `showZeroes` repair path.

The regression runs real workbooks through `MarkItDown.convert_stream`, including a small XLS fixture to avoid adding a workbook-writing test dependency. It checks several NA-like strings, blank cells, and integer values against the resulting Markdown table.

Validation:
- All three new format/path cases fail before the fix and pass afterward.
- Five focused Excel and legacy-repair tests pass.
- Eight existing XLS/XLSX file, hinted-stream, unhinted-stream, and file-URI conversion tests pass.
- Black and `git diff --check` pass. The full multi-format suite was not run; this change adds no dependencies.

AI assistance: OpenAI Codex was used for investigation, implementation, and local test execution.


pentaoa:

PR #2505, opened 2026-09-15
fix(html): use srcset when an img has no usable src

Jia-ben00:
`_CustomMarkdownify.convert_img` reads `src` and `data-src` but not `srcset`, so a responsive image that carries its URL only in `srcset` loses it:

    A photo

renders as `A photo` - an empty destination, with no indication that anything was dropped. The existing `data-src` fallback already covers the same situation for lazy-loaded images, where a placeholder data URI would otherwise win over the real URL; `srcset` and `data-srcset` are the responsive equivalents of that case, and neither was consulted.

This adds a `srcset` parser and extends the fallback chain to `src` (when usable) -> `data-src` -> `data-srcset` -> `srcset`. Candidates are split on a comma followed by whitespace, so commas inside a URL query string (`?crop=1,2`) are not mistaken for separators, and the descriptor (`2x`, `800w`) is stripped from the selected candidate. Only the first candidate is used, since markitdown emits a single destination and the first candidate is the one a browser picks at the default density. When `keep_data_uris` is set, an embedded data URI in `src` is still preferred, as before.

Adds `tests/test_img_srcset.py`, covering the fallback order, the placeholder case, descriptor and whitespace handling, and the comma-in-URL case.


Jia-ben00:

PR #2507, opened 2026-09-16
fix(html): keep the whitespace of a whitespace-only inline element

### LINKED ISSUE #1539: Whitespace lost when DOCX has formatting boundaries mid-text (upstream markdownify issue)
## Summary

When converting DOCX files where a space character has different formatting than adjacent text, the space is lost in the output. This happens because markdownify (used internally via mammoth) strips whitespace-only content from inline formatting tags.

**Example:**
- DOCX content: `further` [normal] + ` ` [bold] + `reference` [normal]
- HTML from mammoth: `further<strong> </strong>reference`
- Current output: `furtherreference`
- Expected output: `further reference`

## Root Cause

This is an upstream issue in markdownify, not markitdown itself. When `chomp()` encounters whitespace-only text inside inline tags like `<strong>`, `<b>`, `<em>`, or `<i>`, it strips the content entirely.

**Upstream issue:** https://github.com/matthewwithanm/python-markdownify/issues/155
**Fix PR:** https://github.com/matthewwithanm/python-markdownify/pull/253

## Workaround

Until the markdownify fix is released, users can apply a monkey-patch before importing markitdown:

```python
import markdownify

_original_convert_strong = markdownify.MarkdownConverter.convert_strong
_original_convert_b = markdownify.MarkdownConverter.convert_b
_original_convert_em = markdownify.MarkdownConverter.convert_em
_original_convert_i = markdownify.MarkdownConverter.convert_i

def _make_whitespace_preserving_converter(original_method):
    def wrapper(self, el, text, *args, **kwargs):
        if text and text.strip() == '' and len(text) > 0:
            return text
        return original_method(self, el, text, *args, **kwargs)
    return wrapper

markdownify.MarkdownConverter.convert_strong = _make_whitespace_preserving_converter(_original_convert_strong)
markdownify.MarkdownConverter.convert_b = _make_whitespace_preserving_converter(_original_convert_b)
markdownify.MarkdownConverter.convert_em = _make_whitespace_preserving_converter(_original_convert_em)
markdownify.MarkdownConverter.convert_i = _make_whitespace_preserving_converter(_original_convert_i)

from markitdown import MarkItDown  # Import after patch
```

## Suggested Action

Once markdownify releases the fix, consider bumping the dependency version to include it.

L4XB:
Closes #1539

### The bug

`markdownify.chomp()` lifts the whitespace surrounding an inline element's text
out of that text, and the conversion functions then return `''` once nothing is
left. A whitespace-only element therefore takes its whitespace with it and the
words on either side run together.

| input | on `main` | with this branch |
|---|---|---|
| `further<strong> </strong>reference` | `furtherreference` | `further reference` |
| `<b>First</b><b> </b><b>Last</b>` | `**First****Last**` | `**First** **Last**` |
| `Hello<code> </code>world` | `Helloworld` | `Hello world` |

This is #1539, which was closed pointing at an upstream fix. `markdownify 1.2.3`
is the current release and `abstract_inline_conversion` still reads:

```python
prefix, suffix, text = chomp(text)
if not text:
    return ''
```

so the behaviour is unchanged on `main` today, with that version installed.

It is not only an HTML-input problem. A word processor keeps a differently
formatted space as a run of its own, so a `.docx` where the space between two
plain words is bold reaches mammoth as `First<strong> </strong>Last` and comes
out of MarkItDown as `FirstLast`. `tests/test_docx_inline_whitespace.py` builds
exactly that document; three of its four formattings fail on `main`.

### The fix

#2477 solved this for `<u>` by overriding `convert_u`. There are twelve more
tags with the same conversion shape, so rather than twelve more overrides this
wraps the function markdownify resolves for a tag in `get_conv_fn`: a
whitespace-only element returns its text unchanged, everything else takes the
normal path.

`b, code, del, em, i, kbd, s, samp, strike, strong, sub, sup, u`

`a` is deliberately not in that set. #2396 is open for anchors and covers more
than the whitespace-only case, including `<a> middle </a>`, which this branch
does not touch.

`get_conv_fn` arrived in markdownify 1.0. On an older release the dispatcher
never calls it, so the override is simply never reached and nothing changes.

### Tests

- `tests/test_html_converter.py` gains the whitespace-only case for all thirteen
  tags, the exact-character case (`&#160;` stays ` `), the empty-element
  case, the joined-emphasis case, and controls that an element with real content
  still converts to `**word**`, `` `word` ``, `~~word~~`, `<u>word</u>`.
- `tests/test_docx_inline_whitespace.py` is new: it builds a minimal `.docx`
  whose middle run is a single space with bold, italic, underline or
  strikethrough, and asserts the result is `First Last`.

15 of the new cases fail on `main`. Seven mutants of the changed lines are
killed by them: the wrapper removed, the guard never firing, the guard inverted,
every element returning unconverted text, and `b`, `i`, `s` dropped from the tag
set one at a time.

```
packages/markitdown      919 passed, 14 skipped   (main: 892 passed, 14 skipped)
packages/markitdown-ocr   53 passed
packages/markitdown-mcp   21 passed
black 23.7.0 --check      97 files unchanged
```

Python 3.12, macOS, markdownify 1.2.3.


L4XB:

PR #2511, opened 2026-09-16
fix(ipynb): fence code cells with the notebook's own language

itzzdev09:
### Problem

`IpynbConverter` fences every code cell as `python`:

```python
md_output.append(f"```python\n{''.join(source_lines)}\n```")
```

A notebook saved by any other kernel is therefore converted into Markdown that says its code is Python:

```python
>>> from markitdown import MarkItDown, StreamInfo
>>> nb = {"nbformat": 4, "nbformat_minor": 5,
...       "metadata": {"language_info": {"name": "julia"}, "kernelspec": {"language": "julia"}},
...       "cells": [{"cell_type": "code", "source": ["x = 1\n"], "metadata": {}, "outputs": []}]}
>>> MarkItDown().convert_stream(io.BytesIO(json.dumps(nb).encode()),
...                             stream_info=StreamInfo(extension=".ipynb")).markdown
'```python\nx = 1\n\n```'
```

That is wrong syntax highlighting wherever the Markdown is rendered, and a wrong signal to an LLM reading the output, which is the case MarkItDown exists for.

### Change

Take the language from the notebook's own metadata: `metadata.language_info.name` (written by the kernel on save), falling back to `metadata.kernelspec.language`, and finally to `python` — so notebooks with no usable metadata keep exactly the output they have today.

The value comes from the file, so it is kept to the first whitespace-delimited token with backticks removed: an info string ends at the first space, and a backtick inside it would close the fence it is supposed to open.

### Tests

Four tests in `test_module_misc.py`: the `language_info` case, the `kernelspec` fallback, the default-to-Python case (missing, empty and whitespace-only metadata), and a language string crafted to break out of the fence.

The first three of those fail on `main` (`'```julia\nx <- 1' not in ...`, etc.) and pass with the change. Full suite: 901 passed, 4 failed before and after — the four are pre-existing failures in my environment that need network access or a speech model, unrelated to this change. `black` clean.


itzzdev09:

PR #2524, opened 2026-09-17
fix(xlsx): keep integer rendering when blanks promote a column to float

### LINKED ISSUE #2484: XLSX/XLS: whole numbers render as "1.0" when the column contains a blank cell
### Description

In XLSX/XLS conversion, a single blank cell forces pandas to read its column as `float64`, so every whole number in that column renders with a `.0` suffix, inconsistent with clean columns beside it:

```python
# sheet: A1='count', A2=1, A3 blank, A4=2
...
'| count |',
'| 1.0 |',
'| NaN |',
'| 2.0 |',
```

Expected the integer rendering of clean columns: `1`, `2`.

### Reproduction

```python
import io
from openpyxl import Workbook

wb = Workbook(); ws = wb.active
ws['A1'] = 'count'; ws['A2'] = 1; ws['A3'] = None; ws['A4'] = 2
buf = io.BytesIO(); wb.save(buf); buf.seek(0)
print(MarkItDown().convert_stream(buf, stream_info=StreamInfo(extension='.xlsx')).markdown)
```

No network or fixture needed.

### Notes

Distinct from the NaN-noise discussion in #2124 / #2286: simulating their patch still renders `1.0` for a numeric-header/data-only sheet, so the float64 promotion is a separate defect.

### Environment

markitdown main (`cc0ca9e`), Python 3.12

Frank-zhu0404:
## Summary
- After reading each sheet, convert float64 columns whose non-null values are all whole numbers back to ints (object dtype) before `to_html`, so values like `1` no longer render as `1.0` when a blank cell promoted the column.
- Applied to both the `.xlsx` and `.xls` converters.
- Adds regression tests covering the issue repro, mixed blank/int columns, fractional preservation, and the shared helper.

Fixes #2484.

## Test plan
- [x] Reproduced the issue on main (`1.0` / `2.0`); confirmed fixed output (`1` / `2`)
- [x] `pytest packages/markitdown/tests/test_xlsx_number_formatting.py` — 5 passed
- [x] Related xlsx tests (`test_xlsx_images.py`, showZeroes misc) — 25 passed


Frank-zhu0404:

PR #2525, opened 2026-09-18
fix(rss): emit item and entry links as heading anchors

ManoharPaturi:
The RSS and Atom converters read the feed-level link but never the per-item one, so every entry came out as a bare heading with no URL to reach. Titles now render as markdown links to the resolved item URL (Atom `<link href>` attribute; RSS `<link>` text; relative hrefs resolve against the feed URL when the document URI is known).

No output change for feeds whose items have no link element. Existing RSS suite green (107 passed).


sylvesterkaczmarek (commented):
When an item has a link this drops the heading marker. heading starts as ## title, then gets replaced with [title](url), so linked RSS/Atom entries stop being headings. Looks like that assignment just needs to keep the ## prefix.

ManoharPaturi:
good catch, fixed. the linked branch now keeps the heading marker, so it renders as a level-2 heading with an anchor instead of a bare paragraph link. all 164 rss tests still green.

sylvesterkaczmarek (approved):
Rechecked 40d6912b. The linked Atom entry now keeps the ## heading marker while wrapping the title in the anchor, so linked items remain level-2 headings. That resolves the formatting issue I raised.

ManoharPaturi:
thanks for the fast turnaround on all three of these, the heading catch saved the feature from shipping half broken.

PR #2529, opened 2026-09-18
fix: keep whitespace inside fenced code blocks when normalizing results

### LINKED ISSUE #2528: fenced code blocks lose blank-line runs and trailing spaces during conversion
### what happens

converting markdown through MarkItDown corrupts fenced code blocks:

- two or more consecutive blank lines inside a fence collapse to one
- trailing spaces on lines inside a fence get stripped

```python
def a():
    pass


def b():
    pass
```

comes back with the double blank line gone, and `line with trailing spaces   ` loses its spaces.

### why

the result normalization (rstrip each line, collapse 3+ newline runs) runs over the whole document with no awareness of code fences, so it mangles content where whitespace is significant.

### expected

whitespace inside fenced code blocks (``` or ~~~) survives exactly; normalization still applies to prose outside fences.

env: main `945314a`

ManoharPaturi:
Fixes #2528.

the normalizer now tracks fenced regions (backtick and tilde fences) and skips line-level rstrip and newline-run collapsing inside them, so code blocks come through byte-exact while prose outside fences keeps the existing normalization.

tests cover blank-line runs inside fences, trailing spaces, tilde fences, and that outside-fence normalization still applies. all fail on main, pass here.


sylvesterkaczmarek (commented):
There's still a boundary case here: a blank-line run immediately before a fence can escape the collapse. With `para\n\n\n```...`, `flush_segment()` leaves `para\n\n`, then the final `"\n".join(out)` adds another newline before the fence, so it stays at three newlines instead of two. Could you add an adjacent-fence case and avoid adding that extra separator?

ManoharPaturi:
good catch, reproduced it exactly as you described (three newlines surviving before the fence). fixed by stripping the trailing newlines when a segment flushes because a fence follows, and keeping exactly one blank line between preceding prose and the fence. the EOF flush keeps the old behavior so trailing newlines at end of document are untouched.

added the adjacent-fence case as a test, plus 2-and-4-newline variants. the whitespace suite is green and the only failure in the full run (test_speech_transcription) is the same on main, it needs a model download in this environment.

sylvesterkaczmarek (approved):
Rechecked 7e6ed5c5. The fence boundary case is fixed: a prose segment is stripped of trailing newlines before the separator is added, so runs immediately before a fence collapse to exactly one blank line. The adjacent-fence regression covers the case I raised. No blocker from me.

ManoharPaturi:
thanks for the quick recheck and the thorough review, the boundary catch made the fix properly correct. the workflows still need a maintainer approval to run, so whenever someone with write access can green-light those, this should be ready to go.

PR #2285, opened 2026-08-08
feat: extract outputs from Jupyter Notebook cells

manideep-malyala:
## Motivation and Context
Currently, the `IpynbConverter` only extracts source code from Jupyter Notebook cells and drops all cell execution outputs. Because `markitdown` is primarily designed as a utility tool to convert files into Markdown for Large Language Model (LLM) consumption, dropping these outputs results in a significant loss of context. 
For Data Science and RAG pipelines, the execution results of a notebook cell (such as printed metrics, `pandas` DataFrame plain-text representations, or error tracebacks) are often just as critical as the code itself for the LLM to understand the document's state.
This PR introduces robust output extraction for Jupyter Notebook code cells, drastically improving the utility of `.ipynb` conversion for AI and indexing workloads.
## Description of Changes
- **Targeted Output Parsing:** Updated `_ipynb_converter.py` to inspect the `outputs` array of any `cell_type == "code"`.
- **Supported Output Types:**
  - `stream`: Captures standard `stdout` and `stderr` outputs.
  - `execute_result` & `display_data`: Safely extracts the `text/plain` keys (falling back gracefully if they do not exist), which properly captures printed tables, integers, and standard objects.
  - `error`: Captures execution failures by extracting the `ename` (Error Name) and `evalue` (Error Value).
- **Safe Dictionary Access:** All JSON dictionary accesses use `.get()` with safe defaults to prevent `KeyError` exceptions on corrupted, malformed, or manually edited notebooks.
- **Clean Markdown Formatting:** Appends the parsed outputs cleanly below the source code inside a `**Output:**` heading and a ```text``` code block. This distinct formatting allows LLMs to easily differentiate between executable source code and console output. If a cell has no output (e.g. variable assignment), the block is gracefully omitted.
## Testing
- **Updated Test Vectors:** Modified the expected Markdown assertions for `test_notebook.ipynb` within `tests/_test_vectors.py` to ensure that both standard text streams (`"markitdown"`) and integer outputs (`"42"`) are successfully captured by the new logic.
- **Validation:** 
  - Ran `hatch test -v` locally; all 340 test cases passed.
  - Ran `pre-commit run --all-files`; code has been formatted using `black` to match Microsoft's repository standards.

PR #2534, opened 2026-09-18
feat(ipynb): render code cell outputs (streams, errors, text results)

ManoharPaturi:
Fixes #2285.

Code cell outputs were dropped entirely; only the sources converted. This renders the text-bearing ones after each cell:

- stream outputs (stdout/stderr) as fenced blocks
- error outputs with the recorded traceback, falling back to ename: evalue
- execute_result / display_data text/plain payloads

Binary outputs (images, HTML) are intentionally not handled. Cells without outputs are byte-for-byte unchanged (the empty-render join is skipped), so existing conversions do not change shape.

3 new tests (stream+error, text result, no-output cell unchanged); the first two fail on main. Notebook-related suites green (9 passed).


sylvesterkaczmarek (commented):
Output is put inside a fixed triple-backtick fence without checking its contents. A cell that prints a line containing three backticks will close the fence early and corrupt the rest of the Markdown. Could this pick a fence longer than any backtick run in the output?

ManoharPaturi:
good catch, that would corrupt the document badly. fixed with a small helper that scans the content for the longest backtick run and uses a fence one tick longer (min 3, so normal outputs are unchanged). applied to code sources, streams, errors, text results and raw cells. added a regression test that prints a triple-backtick line and asserts the output fence grows to four ticks while a backtick-free source fence stays at three.

sylvesterkaczmarek (approved):
Rechecked b24f82b7. The fence is now chosen one backtick longer than the longest run in the cell/output content, so embedded triple-backtick (or longer) runs cannot terminate the block early. The regression covers the exact case I raised, and the focused notebook-output suite passes 4/4 locally. My concern is resolved.

ManoharPaturi:
thanks again for the careful review, approving this and #2529 means a lot. the fence-length idea was exactly the right call, it's the same approach python's repr uses for quote selection so it felt natural here.

Issue #2336, opened 2026-08-24
Improve Arabic/RTL PDF text extraction: logical reading order and table column associations are not preserved


Dhaoui0Chaher:
<html><head></head><body><h1>Improve Arabic/RTL PDF text extraction: logical reading order and table column associations are not preserved</h1><h2>Description</h2><p>MarkItDown's PDF converter does not reliably preserve the logical reading order of Arabic and other right-to-left (RTL) text.</p><p>In some PDFs, Arabic characters, words, lines, and table columns are extracted according to their visual placement in the PDF rather than their Unicode logical reading order. This can produce reversed text and, more importantly, incorrect relationships between table headers and values.</p><p>I observed this behavior with MarkItDown <code inline="">0.1.3</code> and reproduced the reversed Arabic output with <code inline="">0.1.7</code>.</p><h2>Example</h2><p>Expected Arabic text:</p><pre><code class="language-text">تاريخ المزاد
مزاد ربى نجد
قطعة رقم 604 شمالاً
شركة أركان الكيان العقارية
</code></pre><p>Extracted output:</p><pre><code class="language-text">دازملا خيرات
داﺰـﻣ
ﺪﺠﻧ ﻰﺑر
604 مقر ةعطق لاً امش
ةيراقعلا نايكلا ناكرأ ةكرش
</code></pre><p>This appears to involve more than simple line ordering. Some text also contains Arabic Presentation Forms, and mixed Arabic/numeric sequences are reconstructed incorrectly.</p><h2>Table extraction issue</h2><p>The problem becomes more significant when extracting structured information from RTL tables.</p><p>For example, the source PDF may contain property boundaries equivalent to:</p>
Direction | Boundary
-- | --
North | Plot 604
South | Plot 602
East | Plot 597
West | 30-meter street

[...]

PR #2539, opened 2026-09-18
fix(pdf): normalize Arabic presentation forms (RTL phase 1)

aslamalkarywk7:
Related to #2336 (Phase 1 of the plan proposed there).

Some PDF producers emit Arabic Presentation Forms (U+FB50-U+FDFF, U+FE70-U+FEFF) instead of standard letters, which breaks downstream matching and LLM extraction. PdfConverter now normalizes exactly those two ranges to standard Unicode (NFKC per character, so ligatures expand too); everything else passes through byte-identical. Applied as a post-processing stage next to the existing numbering merge, so all extraction paths are covered.

Out of scope for this phase (stated openly): full BiDi reading-order reconstruction and RTL table column handling need the reporter redacted fixtures offered in the issue (or a new bidi dependency) — proposed as Phase 2 once fixtures land.

Verification: new tests/test_pdf_rtl.py (5 tests, fixtures use explicit escapes so they are encoding-proof) pass; existing pdf suites (masterformat, tables: 27 tests) pass unchanged.


aslamalkarywk7:

PR #2540, opened 2026-09-19
fix(bing-serp): do not rewrite result URLs that use their own u parameter

RainieLLM:
## Summary

`BingSerpConverter` decoded the `u` query parameter on every result link. As a result, a normal URL such as `https://example.com/profile?u=abcdef` could become an invalid link.

This change limits decoding to Bing's exact `/ck/a` path on `bing.com` or its subdomains. Normal URLs and links with similar hosts or paths stay unchanged.

## Validation

- The regression tests report 5 passed.
- The existing Bing SERP vector reports 6 local cases passed. The remote HTTP case could not run because `raw.githubusercontent.com` was unavailable.
- The full non-CLI suite previously reported 831 passed. Its five external-service failures also occur on `main`.
- Black reports no formatting changes.


RainieLLM:

PR #2543, opened 2026-09-21
Fix charset parsing of quoted HTTP Content-Type parameters

jesko2004:
`convert_response()` splits `Content-Type` on every semicolon, including semicolons inside quoted parameter values. For example, `text/plain; charset=iso-8859-1; profile="urn:example;charset=utf-8"` incorrectly overrides the real charset with UTF-8 and fails to convert `b"Caf\xe9"`.

Use the already imported standard-library `email.message.Message` to parse the charset parameter. This preserves quoted values and converts the response to `Café`, without adding a dependency or changing explicit `StreamInfo` overrides.

Nine regression and control cases use in-memory `requests.Response` objects and the public conversion API. Against unchanged main, two cases fail and seven pass; all nine pass with this fix.

Related: #1844 handles case-insensitive parameter names. This PR addresses quoted parameter boundaries; the HTTP parser hunk overlaps, while Data URI parsing is unchanged.

Validation (Windows, Python 3.12.14): full core suite **984 passed, 34 skipped** with CI skip conditions; related tests **89 passed, 5 skipped**; `pre-commit run --all-files` and `git diff --check` passed.

Prepared and tested with OpenAI Codex assistance.


VANDRANKI (approved):
Community review, does not clear the merge gate.

I read the diff and confirmed `from email.message import Message` is already imported at the top of `_markitdown.py` (line 9, alongside `collapse_rfc2231_value` from `email.utils`), so this isn't introducing a new dependency, just reusing an idiom already used elsewhere in the file for header parsing.

I traced the actual bug by hand against the old code:

```python
parts = response.headers["content-type"].split(";")
...
for part in parts:
    if part.strip().startswith("charset="):
        _charset = part.split("=")[1].strip()
```

For `'text/plain; charset=iso-8859-1; profile="urn:example;charset=utf-8"'`, the naive `.split(";")` doesn't know the semicolon inside the quoted `profile` value isn't a real delimiter, so it produces parts `["text/plain", " charset=iso-8859-1", ' profile="urn:example', 'charset=utf-8"']`. The loop first sets `charset = "iso-8859-1"` from the real parameter, then hits the stray `'charset=utf-8"'` fragment (the tail end of the quoted profile string, which happens to start with `charset=` after stripping) and overwrites it with `'utf-8"'`, a malformed value with a trailing quote still attached. That's a real, silent charset-detection bug whenever any other Content-Type parameter's quoted value happens to contain the substring `charset=` after a semicolon.

The fix delegates to `email.message.Message.get_param("charset")`, which is quote-aware RFC 2045 parameter parsing and doesn't split inside quoted strings. I checked this against all 5 parametrized cases in the new test: charset-before-profile, charset-before-profile-with-extra-params, profile-before-charset, a quoted charset value, and the plain unquoted case. All of them use `Caf\xe9` decoded against `iso-8859-1` (giving `Café`), so the test is actually checking the real decoded value, not just that some charset string was extracted, which is a meaningful assertion since decoding with the wrong/malformed charset would produce different or broken output.

I also checked the empty-charset case (`charset=""`) against the `isinstance(_charset, str) and _charset.strip()` guard: `get_param` would return `""` for an explicitly empty quoted value, `.strip()` on that is falsy, so `charset` stays `None` and falls through to whatever the existing default-charset handling is, same intent as the old `if len(_charset) > 0` check.

This is a correct, well-targeted fix using the standard library's own header-parsing logic instead of a hand-rolled split that can't handle quoting.

